PR #373, opened 2026-07-21
chore: update ios submodule to latest main

panghy:
Bumps `packages/ios` to `9be4bd4`, the squash-merge of [intent-hq/ios#34](https://github.com/intent-hq/ios/pull/34).

## What

- **Submodule bump**: `packages/ios` `cb97b27` → `9be4bd4` (`fix: preserve tool block name/input when sparse progress ticks arrive over chat.subscribe`).
- **STAB tracker**: files **STAB-162** (iOS chat streaming, P2, fixed) in `docs/01_stabilizing/KNOWN_ISSUES.md` for the live-path clobber regression, and cross-references it from STAB-158.

## Background

After the STAB-158 tool-title fix (ios #32), the bug persisted in a different form: sparse `tool_call_update` progress ticks over `chat.subscribe` carry default/empty fields (`name: ''`, `input: {}`, `toolKind: 'other'`) and clobbered previously titled `tool_use` blocks in the iOS `ConversationStore` — so every live tool call rendered as a bare spanner until conversation re-entry. Desktop was unaffected because its bridge merges instead of replacing.

ios #34 makes `ConversationStore` merge empty-name ticks onto the prior block (desktop bridge parity). Regression test fails without the fix; full suite green (304 tests).

### DIFF
--- a/docs/01_stabilizing/KNOWN_ISSUES.md
+++ b/docs/01_stabilizing/KNOWN_ISSUES.md
@@ -2,7 +2,7 @@
 
 Live issue tracker for the **01_stabilizing** self-hosting phase.
 
-**Next available ID:** STAB-162 (as of 2026-07-21)
+**Next available ID:** STAB-163 (as of 2026-07-21)
 
 ## Intake Convention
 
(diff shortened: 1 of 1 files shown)


panghy:
## Verifier review — ✅ APPROVED (confidence: High)

Verified all three review criteria with concrete evidence:

**1. Gitlink points at real ios origin/main HEAD; nothing else touched — ✅**
- `intent-hq/ios` `main` HEAD is `9be4bd4`; its sole parent is `cb97b27` (the previous gitlink), so the bump is exactly one commit: the GitHub-signed squash merge of intent-hq/ios#34 (`fix: preserve tool block name/input when sparse progress ticks arrive over chat.subscribe (#34)`).
- PR diff touches exactly 2 files: `packages/ios` (gitlink `cb97b27` → `9be4bd4`) and `docs/01_stabilizing/KNOWN_ISSUES.md`. No intentd or cloudlands-fe gitlink changes.

**2. STAB-162 entry follows intake convention and is factually consistent with ios#34 — ✅**
- Heading carries ID/date/area/severity (`STAB-162 (2026-07-21, area: iOS chat streaming, severity: P2)`); body has Repro, Root cause, Expected, and `Status: fixed (PR link, 2026-07-21)` — matching the file's convention and the structure of neighboring entries.
- Placement is newest-first under Fixed Issues (above STAB-161); **Next available ID** correctly bumped to STAB-163.
- Facts cross-check against ios#34's diff/body: sparse `tool_call_update` defaults (`name:""`, `input:{}`, `toolKind:"other"`), `ConversationStore` verbatim-upsert clobber, new `mergedToolUseBlock(prior:incoming:)` merge (desktop bridge parity), regression test `sparseProgressTickPreservesPriorToolBlockFields` verified to fail without the fix, 304 tests green.
- The repro's build reference `c3c27a4` is the real STAB-158 fix commit (ios#32 squash) on ios main.
- STAB-158 cross-reference is a clean append noting the partial fix, without rewriting its original status.

**3. Conventional commits — ✅**
- Single commit `927d26f` and PR title are both `chore: update ios submodule to latest main` — valid conventional commit; `pr-title` check passed. Squash-merge will land a conventional title on main.

**CI**: `pr-title` ✅, `CodeQL` ✅, `Analyze (actions)` ✅ at time of review.

Minor non-blocking note: ios#34's problem statement describes only the *currently-executing* tool call as bare while STAB-162 says every live tool call "stayed that way for the rest of the turn" — the latter matches ios#34's root-cause section ("every subsequent sparse tick kept it clobbered until turn end") and the second commit (`treat any empty-name tool tick as progress-only`), so the tracker wording reflects the fuller diagnosis. No change needed.

(Posted as a comment rather than a formal GitHub approval because the reviewing credential is the PR author.)

PR #434, opened 2026-07-22
fix: make ensure-submodules init all submodules recursively

panghy:
## Summary

`make ensure-submodules` now initializes **all** submodules (`packages/intentd`, `packages/cloudlands-fe`, `packages/ios`) with `git submodule update --init --recursive`, idempotently, instead of only intentd.

## Changes (Makefile only)

- Widened `SUBMODULES` to `$(INTENTD_DIR) $(FE_DIR) $(IOS_DIR)`; the existing init-if-missing loop now covers all three.
- Added `ensure-intentd-submodule` and pointed the Rust-only targets (`build-intentd`, `fmt`, `clippy`, `test-intentd`, `dev-daemon`, `release-daemon`) at it, so backend-only workflows (`make check`, `make test`) still don't pull the heavy FE/iOS submodules. `build-sidecar` and `dev` additionally depend on `ensure-fe-submodule`; nothing forces the iOS init.
- Updated `.PHONY`, the comment block above `SUBMODULES`, and the `make help` description.

## Verification

- `make ensure-submodules` twice: first run initializes any missing submodules at their pinned gitlinks; second run is a full no-op ("already initialized — leaving as-is" for all three).
- `make -n build-intentd` and `make -n fmt` show only the intentd init check.
- `make help` lists the updated description.
- `git submodule status` shows no ref drift.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,12 +19,13 @@ INTENTD_DIR = packages/intentd
 FE_DIR = packages/cloudlands-fe
 IOS_DIR = packages/ios
 
-# `ensure-submodules` covers the intentd submodule that every Rust target
-# (build/test/fmt/clippy/check) needs. The FE and iOS submodules are heavy and
-# not needed for backend-only workflows, so they are initialized on demand by
-# their own `ensure-fe-submodule` / `ensure-ios-submodule` targets (same
-# idempotent init-if-missing contract, just narrower).
-SUBMODULES = $(INTENTD_DIR)
+# `ensure-submodules` covers ALL submodules (intentd + FE + iOS), initializing
+# any that are missing. The FE and iOS submodules are heavy and not needed for
+# backend-only workflows, so those workflows use the narrower per-submodule
+# targets instead: `ensure-intentd-submodule` for every Rust target
+# (build/test/fmt/clippy/check), and `ensure-fe-submodule` /
+# `ensure-ios-submodule` on demand (same idempotent init-if-missing contract).
+SUBMODULES = $(INTENTD_DIR) $(FE_DIR) $(IOS_DIR)
 
 # Dev-seat data + ports (overridable, e.g. `make dev-daemon DEV_TCP_PORT=6181`).
 #
@@ -56,7 +57,7 @@ export DEV_PORT
 # TCP+UDS or TCP-only postures, e.g. `make release-daemon LISTEN=both`.
 LISTEN ?= uds
 
-.PHONY: all help ensure-submodules ensure-fe-submodule ensure-ios-submodule \
+.PHONY: all help ensure-submodules ensure-intentd-submodule ensure-fe-submodule ensure-ios-submodule \
 	build build-intentd build-sidecar test test-intentd fmt clippy check clean clean-dev \
 	dev-daemon release-daemon run-intentd run-fe dev ios-open ios-info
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -65,7 +66,7 @@ all: build
 help: ## List documented targets
 	@awk 'BEGIN {FS = ":.*## "} /^[a-zA-Z0-9_-]+:.*## / {printf "  %-16s %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
-ensure-submodules: ## Initialize any missing Rust-workflow submodules (idempotent)
+ensure-submodules: ## Initialize any missing submodules — intentd, FE, iOS (idempotent)
 	@for sm in $(SUBMODULES); do \
```
panghy, line 70: **Correctness:** now that the loop iterates over three submodules, a failed `git submodule update` on an earlier submodule is silently masked — the recipe's exit status is that of the *last* iteration, and make invokes the shell without `-e`. Reproduced the shape of the bug:

```sh
$ sh -c 'for x in 1 2; do if [ "$x" = 1 ]; then false; else echo "iter $x ok"; fi; done'; echo $?
iter 2 ok
0
```

So if e.g. the intentd init fails (network, bad gitlink) but FE/iOS are already initialized, `make ensure-submodules` exits 0, which misleads CI/scripts and the `git submodule update --init --recursive; make check` setup flow in AGENTS.md. With the old single-entry `SUBMODULES` this couldn't happen (the failing iteration was always the last one), so this is newly reachable in this PR.

Minimal fix:
```make
			git submodule update --init --recursive "$$sm" || exit 1; \
```
(or prefix the recipe with `set -e;`).

Makefile
```diff
@@ -75,6 +76,17 @@ ensure-submodules: ## Initialize any missing Rust-workflow submodules (idempoten
...
+ensure-intentd-submodule:
+	@if [ ! -e "$(INTENTD_DIR)/.git" ]; then \
+		echo "[ensure-intentd-submodule] initializing $(INTENTD_DIR)"; \
+		git submodule update --init --recursive "$(INTENTD_DIR)"; \
+	else \
+		echo "[ensure-intentd-submodule] $(INTENTD_DIR) already initialized — leaving as-is"; \
+	fi
+
 # On-demand init for the FE submodule — pulled in only by targets that need it
 # (`run-fe`), so backend-only workflows stay fast.
```
panghy: **Nit (pre-existing, non-blocking):** this comment says the FE submodule is "pulled in only by targets that need it (`run-fe`)", but `build-sidecar` and `dev` also depend on `ensure-fe-submodule`. Since this PR is already updating the neighboring comment blocks for accuracy, consider listing all three (or dropping the parenthetical) while you're here.

panghy (commented):
Reviewed the Makefile diff against the stated goals. Most of it checks out:

- **Dependency graph** ✅ — all Rust targets (`build-intentd`, `fmt`, `clippy`, `test-intentd`, `dev-daemon`, `release-daemon`) depend only on `ensure-intentd-submodule`; `build-sidecar`/`dev` get intentd+FE; `ios-open` gets iOS; nothing forces FE/iOS init for backend-only workflows. Verified with `make -n build-intentd` / `make -n ios-open`.
- **.PHONY** ✅ — `ensure-intentd-submodule` added; all 23 targets in the file are listed.
- **Idempotency** ✅ — ran `make ensure-submodules` on an initialized tree; full no-op for all three.
- **Conventional commit** ✅ — PR title and the single commit `ad04afc` (`fix: ...`) both validate.

One correctness finding on the widened loop and one comment nit, inline below.

Makefile
panghy: Fixed in 9bc43b0 — appended `|| exit 1` to the `git submodule update` inside the loop, so a failed init aborts immediately and the recipe exits non-zero. Verified: `make ensure-submodules` is still a no-op (exit 0) on the initialized tree, and `make ensure-submodules SUBMODULES="packages/bogus packages/intentd"` now fails with `make: *** [ensure-submodules] Error 1` even though the later submodule is already initialized (previously this masked the failure and exited 0).

Makefile
panghy: Done in 9bc43b0 — the comment now reads "pulled in only by targets that need it (`run-fe`, `build-sidecar`, `dev`)".

PR #629, opened 2026-07-24
chore: update intentd submodule and document protocol 2.2

### LINKED ISSUE #592: intentd legacy import skips workspaces with valid string setupScript and marks partial import complete
## Symptom

On first boot, the legacy workspace importer reported:

- 60 imported
- 120 skipped
- 308 notes, 444 agent sessions, and 2220 agent messages imported

Every sampled skip failed with:

`workspace.json parse failed: invalid type: string "<shell script>", expected struct SetupScript`

The importer then logged `first-boot legacy workspace import complete` and wrote its completion marker.

## Root cause

The legacy FE workspace schema defines `setupScript?: string` (`cloudlands-fe/src/shared/types.ts`). This is a valid legacy manifest shape.

In `crates/intentd/src/legacy_import.rs`, `workspace_from_legacy_json` removes only `changesets`, `conversationInfo`, and `timeline`, supplies defaults, then deserializes the remaining object directly into the current `intent_core::Workspace`. The current domain model expects `setup_script: Option<SetupScript>` where `SetupScript` is an object (`{ script, updatedAt, ... }`). No import-only conversion handles the legacy string representation.

As a result, the workspace row fails before `import_workspace_extras`, so its notes, comments, transcripts, and assets are all skipped too.

The run still returns `Ok(Report)` because per-workspace parse failures are soft skips. `maybe_import_on_first_boot` therefore writes `import.legacyCompletedAt` even when most candidates failed, preventing startup from retrying them.

## Expected fix

- In the legacy importer only, normalize a string `setupScript` into the current structured `SetupScript` representation (preserve already-structured values).
- Add a regression test using the actual legacy shape: `"setupScript": "#!/bin/bash\n..."`, asserting the workspace and extras import.
- Revisit completion semantics: either do not mark a run complete when compatibility/parse failures occur, or otherwise make partial completion/recovery explicit.
- Verify a manual rerun without `--force` safely skips the 60 existing rows and imports the previously skipped rows and their extras.

## Recovery for affected installs

After shipping the compatibility fix, stop the daemon and run `intentd import-legacy` against the same data directory. The command runs independently of the first-boot marker; without `--force`, existing workspace IDs are skipped while missing ones are imported. Legacy source directories are read-only.

## Provenance

Observed while dogfooding the importer on 2026-07-23. No duplicate found in the monorepo tracker. Related implementation: intent-hq/intentd#372.

panghy:
Updates the `packages/intentd` submodule to latest main and documents protocol 2.2 in `docs/PROTOCOL.md`.

## Submodule bump

`packages/intentd` now points at `c746c25` — the squash merge of intent-hq/intentd#423 (recoverable legacy import RPC).

## Protocol docs (2.1 → 2.2)

- Header, version section, and `client.hello` / `system.status` examples now advertise `2.2`; the compatibility section notes 2.2 adds `system.importLegacy`.
- Method counts updated: **296 dispatchable** (263 router + **31 fast-path** + 2 aliases); `system.importLegacy` added to the fast-path list.
- New `system.importLegacy` subsection in the §5 fast-path catalog:
  - UDS-only — remote WSS/TCP callers rejected with `-32001`
  - Params `{ force?: boolean }` (non-boolean `force` → `-32602`); `force` is overwrite-only
  - Result `{ imported, updated, skipped, notes, comments, agents, assets, skipSummary (first 20 {id, reason}), compatibilityFailures, markerWritten }`; run failures → `-32603`
  - Completion-marker semantics: marker withheld on compatibility failures; explicit CLI/RPC runs are never gated by the marker
  - No bus events are published — clients must refresh to see imported workspaces
- `-32001` error-table row now cites `system.importLegacy` alongside `pairing.getInfo`.

Refs intent-hq/intentd#423. Closes #592.

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -261,7 +261,7 @@ Runs the daemon's legacy workspace import over RPC — the same engine behind th
 }
 ```
 
-- `imported` / `updated` / `skipped` count workspaces; `notes`, `comments`, `agents`, `assets` are total rows/files imported across all workspaces.
+- `imported` / `updated` / `skipped` count workspaces; `notes`, `comments`, `assets` are total rows/files imported across all workspaces, and `agents` counts imported agent **sessions**.
 - `skipSummary` lists the first **20** skipped workspaces as `{ id, reason }`.
 - `compatibilityFailures` is `true` when any workspace was skipped for a non-operational reason (unreadable/unparseable `workspace.json`, missing id, etc.). Operational skips — `already in DB` and transient `update failed:` / `insert failed:` / `lookup failed:` — do not count.
 - `markerWritten` reports whether the first-boot completion marker was (re)written after this run. The marker is **not** written when compatibility failures occur, so the first-boot hook stays armed to retry; explicit CLI/RPC runs are not gated by the marker either way.
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -1,6 +1,6 @@
-# Intent Backend — JSON-RPC Protocol v2.1
+# Intent Backend — JSON-RPC Protocol v2.2
 
-**Protocol Version:** `2.1`
+**Protocol Version:** `2.2`
 
 This document is the canonical wire contract between Intent clients (desktop, iOS, CLI, and agent developers building clients) and the Intent backend daemon (`intentd`): transport, JSON-RPC envelope, the full method catalog, events, agent streaming, the permission flow, error codes, and thin-client guidance. It is a **living specification**: changes land through the compatibility policy below, and the method surface is enforced by golden tests in the `intent-transport` crate.
 
@@ -21,14 +21,14 @@ This document is the canonical wire contract between Intent clients (desktop, iO
 
 ## Protocol Version & Compatibility
 
-**Version:** `2.1`
+**Version:** `2.2`
 
-Version 2.1 is an **additive** minor bump over 2.0: it adds the `pr.capabilities` router method and the provider capability gating described in §5.7. No existing method changed shape.
+Version 2.1 was an **additive** minor bump over 2.0: it added the `pr.capabilities` router method and the provider capability gating described in §5.7. Version 2.2 is an **additive** minor bump over 2.1: it adds the `system.importLegacy` fast-path method (UDS-only — see the §5 fast-path catalog). No existing method changed shape in either bump.
 
 The protocol version is advertised in two places:
 
-- `client.hello` response: `{ protocolVersion: "2.1", server: { protocolVersion: "2.1", ... }, ... }` — the top-level `protocolVersion` is an explicit copy of `server.protocolVersion` so clients can version-check without digging into the `server` block (§5.17).
-- `system.status` response: `{ protocolVersion: "2.1", ... }`
+- `client.hello` response: `{ protocolVersion: "2.2", server: { protocolVersion: "2.2", ... }, ... }` — the top-level `protocolVersion` is an explicit copy of `server.protocolVersion` so clients can version-check without digging into the `server` block (§5.17).
+- `system.status` response: `{ protocolVersion: "2.2", ... }`
 
 ### Compatibility Policy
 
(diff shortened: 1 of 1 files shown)


panghy:
## Docs verification — `docs/PROTOCOL.md` vs intentd @ `c746c25`

Reviewed the protocol-2.2 doc changes against the submodule sources of truth. **All documented claims check out.**

### Verified ✅

| Claim | Source of truth | Evidence |
|---|---|---|
| Protocol version `2.2` | `crates/intent-transport/src/protocol.rs` | `PROTOCOL_VERSION: &str = "2.2"` (line 10); doc comment states 296 = 263 router + 31 fast-path + 2 aliases |
| 296 dispatchable = 263 + 31 + 2 | `crates/intent-transport/src/catalog.rs` | Counted programmatically: `ROUTER_METHODS` = 263, `FASTPATH_METHODS` = 31, `METHOD_ALIASES` = 2 (`git.diff`→`git.diffs`, `git.log`→`git.commits`); `NOTIFICATIONS` = 1, `REVERSE_METHODS` = 4 |
| Fast-path list (31 names) | `catalog.rs:301-333` | Doc's list matches `FASTPATH_METHODS` name-for-name in order, including `system.importLegacy` |
| UDS-only guard `-32001 "system.importLegacy is available over UDS only"` | `crates/intent-transport/src/control.rs:181-184` | Exact code + message; gated on `is_uds` (derived from `!is_tcp_connection()` in `conn.rs:145`) |
| Non-boolean `force` → `-32602 "force must be a boolean"` | `control.rs:185-187` | Exact code + message; `force` defaults to `false` when absent/null (`control.rs:113-121`) |
| Run failure → `-32603` with underlying message | `control.rs:188-191` | `import_legacy(force).await.map_err(\|message\| (-32603, message))` |
| Result shape | `crates/intentd/src/main.rs:1385-1396` | Exactly `imported, updated, skipped, notes, comments, agents, assets, skipSummary, compatibilityFailures, markerWritten` |
| `skipSummary` = first 20 skipped as `{id, reason}` | `main.rs:1374-1384` | `.filter_map(Skipped → json!({id, reason})).take(20)` |
| `compatibilityFailures` semantics (operational skips excluded) | `crates/intentd/src/legacy_import.rs:292-297` | `is_operational_skip`: exactly `"already in DB"`, `"update failed:"`, `"insert failed:"`, `"lookup failed:"` — matches the doc's list verbatim |
| Marker withheld on compatibility failures | `main.rs:1362-1373` (RPC), `legacy_import.rs:1786-1791` (first-boot) | `marker_written = false` when `has_compatibility_failures()`; first-boot hook likewise skips the marker so it stays armed |
| Explicit CLI/RPC runs not gated by marker | `main.rs:1342-1360` (RPC never reads the marker), `cmd_import_legacy` doc comment (`main.rs:331`) | Only the first-boot hook checks the marker before running |
| `force` overwrite-only | `legacy_import.rs:592-613` | Existing workspace → `Skipped("already in DB")` unless `force`, then update; `force` plays no role in marker gating |
| Concurrent calls serialized | `main.rs:1348` | `self.legacy_import_lock.lock().await` held for the whole run |
| No bus events | `legacy_import.rs` | No `EventBus`/`publish` usage anywhere in the module; writes go through `Store` directly |
| Default legacy roots | `main.rs:1352` | RPC uses `legacy_import::default_roots()` + `default_app_dir()` |

### Stale-reference sweep ✅

No leftover `295`, `30 fast-path`, or current-version `2.1` references. Remaining `2.1` strings in the doc are legitimate (historical "Version 2.1 was…" note, `pr.capabilities (v2.1)` annotations, §2.1/§5.2.1 section numbers, an IP address).

### Non-blocking observations (pre-existing, not introduced by this PR)

1. **`system.shutdown` UDS-only claim vs code.** The doc (carried over from 2.1 and rephrased here) says `system.shutdown` is only available over UDS. `system.importLegacy` enforces this with the `-32001` guard, but `SystemMethod::Shutdown` has **no** `is_uds` guard in `control::handle()` (`control.rs:177-180`), and the WSS listener is wired with `Some(control)` (`main.rs:941/1093`, passed at `ws.rs:573`) — so at the code level `system.shutdown` appears reachable by an authenticated remote WSS client. Relatedly, the comment at `ws.rs:566-567` ("does not expose the `system.*` control surface … pass `None`") contradicts the code on the next line. Worth an intentd issue to either add the guard or correct the docs/comment; not a defect of this PR.
2. **`agents` field precision.** The result's `agents` counts imported agent *sessions* (`report.agent_sessions_imported()`); imported agent *messages* are tracked in the report but not surfaced over RPC. The doc's "total rows" phrasing is acceptable, but "agent sessions" would be more precise.

**Verdict:** docs accurately reflect intentd @ `c746c25`. LGTM.

panghy:
Thanks for the verification. Both non-blocking observations are addressed:

1. **`system.shutdown` UDS-only claim vs code** — filed as #630 (pre-existing intentd discrepancy, not introduced here).
2. **`agents` field precision** — fixed in `168b3cb5`: the doc now reads "`agents` counts imported agent **sessions**".

Issue #720, opened 2026-07-25
intentd: event-table dead space never reclaimed (auto_vacuum=0) + line-attribution:updated retention gap


panghy:
Found while auditing a dev-seat database (`make dev`, `.dev/intentd/intentd.db`). Two related findings; sibling of #620 (which fixed the `script:output` sweep gap but not these).

## Finding 1: retention sweeps free pages, but the space is never reclaimed (`auto_vacuum=0`)

The audited dev DB was **1.1 GB on disk with only ~309 MB live data** — `PRAGMA freelist_count` reported 222,062 free pages (~867 MB, 74% of the file) and `PRAGMA auto_vacuum` = 0.

The retention loop's space-reclaim step is a bounded `PRAGMA incremental_vacuum`, but that is a no-op unless the database was created with (or converted to) incremental auto-vacuum — `docs/PROTOCOL.md` §event retention even scopes it to "incremental-auto-vacuum databases". Since the daemon's SQLite databases are created with the default `auto_vacuum=0`, every install falls outside that condition: the 24h/72h sweeps delete rows correctly (verified: `agent:tool:call` and `file:*` windows hold), but the freed pages accumulate forever. #620's remediation also required a manual `VACUUM` for the same reason.

A one-off `VACUUM` on the audited DB took ~12 s and shrank the file 1.1 GB → 287 MB.

**Suggested direction:** set `PRAGMA auto_vacuum=INCREMENTAL` at pool init for new databases, and migrate existing ones (changing `auto_vacuum` requires a `VACUUM` to take effect — could be a one-time maintenance step, e.g. gated on `freelist_count` exceeding a threshold at startup).

## Finding 2: `line-attribution:updated` events are persisted forever and are the heaviest per-row event type

Same dev DB, `SELECT event_type, count(*), sum(length(data_json)) FROM event GROUP BY event_type`:

| event_type | rows | ~MB | notes |
|---|---|---|---|
| `agent:tool:call` | 20,499 | 46.6 | within 24h TTL — retention working |
| `line-attribution:updated` | 4,320 | 45.6 | **~11 KB/row avg; oldest row = day 1 of the install; not in any retention family** |
| `file:changed` | 25,957 | 8.5 | within 72h window — retention working |

`line-attribution:updated` (PROTOCOL §5.2.1 / §6.5) carries a self-sufficient payload: the **entire recomputed attribution map for every line of the note** on each debounced recompute. Persisting each emission has no reader:

[...]

PR #739, opened 2026-07-25
docs: record line-attribution:updated as broadcast-only and bump intentd (#720)

panghy:
Closes finding 2 of #720.

Documents the behavior change from intent-hq/intentd#488 and bumps the `packages/intentd` gitlink to the merged main commit (`5bf0d1e`).

## Changes

- **docs/PROTOCOL.md**
  - §5.2.1 recompute-lifecycle: `line-attribution:updated` is now **transient / broadcast-only** — published through the same transient path as `agent:stream:chunk`, never written to the event table, invisible to `event.query` / §5.10 historical reads. Notes migration `0052_delete_line_attribution_events.sql` (legacy-row cleanup) and that the durable state remains the `note_line_attribution` row.
  - §6.5 taxonomy row for `line-attribution`: same transient/broadcast-only note, pointing at `note.lineAttribution.load` for the durable snapshot.
  - The §5.10 retention note does not name the event, so it is untouched.
- **packages/intentd** gitlink → `5bf0d1e` (intent-hq/intentd#488, `fix: route line-attribution:updated through the transient publish path (monorepo#720)`).

Rebased on main after the #487 bump landed there via #734 (the redundant bump commit was dropped).

## Verification

- `make check` (cargo fmt --check, clippy -D warnings, build) — pass
- `make test` — pass

---


docs/PROTOCOL.md
```diff
@@ -747,8 +747,12 @@ see §5.2 version-history extensions).
 **Recompute lifecycle.** Every content-changing `note.*` mutation schedules a debounced
 recompute (5 s, mirroring `LineAttributionService.DEBOUNCE_MS`). A fresh mutation cancels
 any pending timer so a burst of writes coalesces into one persist + one
-`line-attribution:updated` emit (§6.5). Persistence is one row per note in
-`note_line_attribution` (SQLite migration `0028_note_line_attribution.sql`), upserted on
+`line-attribution:updated` emit (§6.5). The emit is **transient / broadcast-only**
+(published through the same transient path as `agent:stream:chunk`, §7): it is never
+written to the event table, so it does not appear in `event.query` or the other §6.4
```
panghy: **Wrong section reference:** §6.4 is *Filter semantics* (how `eventTypes` compiles to subscription filters) — it contains no historical reads. `event.query` and the other historical/aggregate read helpers (`event.recentFiles`, `event.agentActivity`, …) are documented in **§5.10** ("These are **historical/aggregate read** helpers — distinct from live streaming (§6)"), which also hosts the retention note. Suggest: "…it does not appear in `event.query` or the other §5.10 historical reads".

panghy (commented):
Reviewed as verifier against the intentd implementation at `5bf0d1e` (intent-hq/intentd#488).

**Verified accurate:**
- Submodule gitlink `5bf0d1e` == current tip of intent-hq/intentd `main` == the #488 squash merge. ✅
- `compute_and_persist_line_attribution` publishes `line-attribution:updated` via `publish_event_transient` (`intent-services/src/lib.rs:2984`) — the same path `agent:stream:chunk` uses (`agent_session.rs:1061`). ✅
- Migration `0052_delete_line_attribution_events.sql` exists and deletes legacy `line-attribution:updated` rows. ✅
- Durable state is the `note_line_attribution` row (migration 0028): one row per note, upserted per recompute, `ON DELETE CASCADE` on note delete. Served by `note.lineAttribution.load`. ✅
- Payload `{ workspaceId, noteId, attributions }` matches `line_attribution_updated_event`, and the never-persisted claim is backed by the `uds_line_attribution.rs` regression test (`event.query` returns `[]`). ✅

**Issues (inline comments):**
1. Both new passages cite **"§6.4 historical reads"**, but §6.4 is *Filter semantics* (subscription filter compilation). The historical/aggregate reads including `event.query` are documented in **§5.10** ("These are **historical/aggregate read** helpers"). The PR body repeats the same mislabel ("the §6.4 retention note" — the retention note is also in §5.10).

**Non-blocking observation:** the §5.2.1 sentence points at §7 for the `agent:stream:chunk` transient path, but §7 itself never states chunk is broadcast-only/never persisted (its *Dedup* bullet says "the same agent output is also persisted", meaning the transcript), and the §5.10 retention note lists `agent:stream:*` among *persisted* pruned families. The pointer is acceptable as the section where `agent:stream:chunk` is defined, but the transient nature of chunk is otherwise undocumented in PROTOCOL.md — a follow-up could make §7 explicit.

**Also:** the PR currently shows `mergeable_state: dirty` (conflict with `main`, presumably the `packages/intentd` gitlink) and is still a draft — needs a rebase/merge before it can land.

docs/PROTOCOL.md
```diff
@@ -3448,7 +3452,7 @@ All filters on a subscription are combined with **AND**. Delivery is gated *only
...
-| line-attribution (new in intentd) | line-attribution:updated | Emitted after the daemon recomputes per-line attributions for a note (§5.2.1). data = { workspaceId, noteId, attributions } where `attributions` is the FE-parity `Record<lineNumber, { timestamp, author? }>`. Self-sufficient payload (§6.7) so the FE gutter re-renders without a follow-up `note.lineAttribution.load`. |
+| line-attribution (new in intentd) | line-attribution:updated | Emitted after the daemon recomputes per-line attributions for a note (§5.2.1). data = { workspaceId, noteId, attributions } where `attributions` is the FE-parity `Record<lineNumber, { timestamp, author? }>`. Self-sufficient payload (§6.7) so the FE gutter re-renders without a follow-up `note.lineAttribution.load`. **Transient / broadcast-only** (same publish path as `agent:stream:chunk`): never persisted to the event table, so it is invisible to `event.query` / §6.4 historical reads — the durable snapshot lives in `note_line_attribution` and is served by `note.lineAttribution.load` (§5.2.1). |
```
panghy: Same mislabel as §5.2.1: "invisible to `event.query` / §6.4 historical reads" should reference **§5.10** (`event.*` query/aggregation — the section that defines the historical reads), not §6.4 (*Filter semantics*).

docs/PROTOCOL.md
panghy: Fixed in acb7e3f4 — now reads "…does not appear in `event.query` or the other §5.10 historical reads".

docs/PROTOCOL.md
panghy: Fixed in acb7e3f4 — the §6.5 taxonomy row now references §5.10.

PR #760, opened 2026-07-25
chore: add cargo-sweep make targets (sweep, sweep-all)

panghy:
Adds build-artifact GC targets to the monorepo-root Makefile so stale `target/` artifacts can be pruned regularly.

## Changes
- `make sweep` — runs `cargo sweep --time $(SWEEP_DAYS)` (default 3 days) in `$(INTENTD_DIR)`; fails with a clear message suggesting `cargo install cargo-sweep` when the tool is missing.
- `make sweep-all` — iterates `$(WORKSPACES_DIR)/*/monorepo/packages/intentd` (default `WORKSPACES_DIR ?= $(HOME)/intent/workspaces`, overridable) and sweeps each worktree that has a `target/` dir, skipping missing/uninitialized ones gracefully and continuing past per-worktree failures.
- Both targets added to `.PHONY` with `## description` comments so `make help` lists them; `SWEEP_DAYS` and `WORKSPACES_DIR` are overridable vars in the existing Makefile style.

## Verification
- `make help | grep sweep` lists both targets.
- `make sweep` runs in the current worktree (exits 0; this worktree's `target/` was already removed by a separate cleanup, so cargo-sweep prints its own warning).
- `make sweep-all` iterated 13 sibling worktrees: swept 8 with `target/` dirs, skipped 5 without, exit 0.
- Missing-tool path verified by inspection (`command -v cargo-sweep` guard).

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,9 @@ SUBMODULES = $(INTENTD_DIR) $(FE_DIR) $(IOS_DIR)
 DEV_DATA_DIR ?= $(CURDIR)/.dev/intentd
 DEV_TCP_PORT ?= 5181
 DEV_PORT ?= 5190
+# Export DEV_PORT so `make run-fe` (and any recipe that shells out to the FE)
+# actually sees the default/override in the child environment.
+export DEV_PORT
 
 # Build-artifact GC (cargo-sweep). Rust target/ dirs grow without bound as
 # deps and toolchains churn; `sweep` prunes artifacts older than SWEEP_DAYS
@@ -61,9 +64,6 @@ DEV_PORT ?= 5190
 # or `make sweep-all WORKSPACES_DIR=/elsewhere/workspaces`.
 WORKSPACES_DIR ?= $(HOME)/intent/workspaces
 SWEEP_DAYS ?= 3
-# Export DEV_PORT so `make run-fe` (and any recipe that shells out to the FE)
-# actually sees the default/override in the child environment.
-export DEV_PORT
 
 .PHONY: all help ensure-submodules ensure-intentd-submodule ensure-fe-submodule ensure-ios-submodule \
 	build build-intentd build-sidecar test test-intentd fmt clippy check clean clean-dev \
@@ -139,12 +139,19 @@ clean: ## Remove cargo build artifacts (packages/intentd/target)
 clean-dev: ## Wipe the dev-seat state dir (.dev/)
 	rm -rf "$(CURDIR)/.dev"
 
+# `sweep` deliberately has no ensure-intentd-submodule prerequisite:
+# initializing the submodule just to sweep a target/ dir that cannot exist yet
+# would be overkill, so it short-circuits with a friendly no-op instead.
 sweep: ## Prune intentd build artifacts older than $(SWEEP_DAYS) days (needs cargo-sweep)
 	@command -v cargo-sweep >/dev/null 2>&1 || { \
 		echo "[sweep] ERROR: cargo-sweep is not installed — run 'cargo install cargo-sweep'"; \
 		exit 1; \
 	}
-	cd $(INTENTD_DIR) && cargo sweep --time $(SWEEP_DAYS)
+	@if [ -d "$(INTENTD_DIR)/target" ]; then \
+		cd $(INTENTD_DIR) && cargo sweep --time $(SWEEP_DAYS); \
+	else \
+		echo "[sweep] nothing to sweep — $(INTENTD_DIR)/target does not exist"; \
+	fi
 
 sweep-all: ## Sweep intentd build artifacts in every worktree under $(WORKSPACES_DIR)
 	@command -v cargo-sweep >/dev/null 2>&1 || { \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,13 +51,23 @@ SUBMODULES = $(INTENTD_DIR) $(FE_DIR) $(IOS_DIR)
...
+
+# Build-artifact GC (cargo-sweep). Rust target/ dirs grow without bound as
+# deps and toolchains churn; `sweep` prunes artifacts older than SWEEP_DAYS
+# days in this worktree's intentd, and `sweep-all` does the same across every
+# sibling worktree under WORKSPACES_DIR (the per-worktree monorepo checkouts
+# live at $(WORKSPACES_DIR)/<name>/monorepo). Both need cargo-sweep
+# (`cargo install cargo-sweep`). Overridable, e.g. `make sweep SWEEP_DAYS=7`
+# or `make sweep-all WORKSPACES_DIR=/elsewhere/workspaces`.
+WORKSPACES_DIR ?= $(HOME)/intent/workspaces
+SWEEP_DAYS ?= 3
```
panghy, line 66: Style nit: this block is inserted between `DEV_PORT ?= 5190` and the `# Export DEV_PORT ...` comment + `export DEV_PORT` pair that belongs to it, splitting a logically contiguous block. Consider moving the sweep vars below `export DEV_PORT` (or above `DEV_DATA_DIR`) so the DEV_PORT declaration stays adjacent to its export.

Makefile
```diff
@@ -129,6 +139,28 @@ clean: ## Remove cargo build artifacts (packages/intentd/target)
 clean-dev: ## Wipe the dev-seat state dir (.dev/)
 	rm -rf "$(CURDIR)/.dev"
 
+sweep: ## Prune intentd build artifacts older than $(SWEEP_DAYS) days (needs cargo-sweep)
+	@command -v cargo-sweep >/dev/null 2>&1 || { \
+		echo "[sweep] ERROR: cargo-sweep is not installed — run 'cargo install cargo-sweep'"; \
+		exit 1; \
+	}
+	cd $(INTENTD_DIR) && cargo sweep --time $(SWEEP_DAYS)
```
panghy: Every other target that runs `cd $(INTENTD_DIR) && cargo ...` (build-intentd, fmt, clippy, test-intentd) declares `ensure-intentd-submodule` as a prerequisite. Without it, running `sweep` in a checkout with an uninitialized submodule fails with cargo's `could not find Cargo.toml` error instead of a clear message. Alternatively, a cheap `[ -d $(INTENTD_DIR)/target ]` guard (mirroring sweep-all's per-dir check) would short-circuit with a friendlier no-op — initializing the submodule just to sweep nothing is arguably overkill.

Makefile
```diff
@@ -129,6 +139,28 @@ clean: ## Remove cargo build artifacts (packages/intentd/target)
...
+
+sweep-all: ## Sweep intentd build artifacts in every worktree under $(WORKSPACES_DIR)
+	@command -v cargo-sweep >/dev/null 2>&1 || { \
+		echo "[sweep-all] ERROR: cargo-sweep is not installed — run 'cargo install cargo-sweep'"; \
+		exit 1; \
+	}
+	@for dir in $(WORKSPACES_DIR)/*/monorepo/$(INTENTD_DIR); do \
```
panghy, line 161: Edge case (verified locally): when the glob matches nothing (e.g. `WORKSPACES_DIR` points at an empty/nonexistent dir), POSIX sh passes the pattern through literally, so the loop runs once and prints:

```
[sweep-all] skipping /tmp/emptydir/*/monorepo/packages/intentd (no target/ dir)
```

Exit status is still 0, so behavior is graceful — just slightly confusing output. If you want to tidy it, add `[ -d "$$dir" ] || continue;` as the first statement in the loop body (an unexpanded pattern is never an existing directory). Non-blocking.

panghy (commented):
Reviewed the diff for correctness, scope, and Makefile style. Overall this looks solid: shell quoting is correct (`"$$dir"` quoted everywhere it's used, `$(WORKSPACES_DIR)/*/...` intentionally unquoted for glob expansion), `.PHONY` is updated, the `command -v cargo-sweep` guard matches how `cargo sweep` resolves the subcommand, and the `##` help comments follow the existing convention (verified `make help` lists both targets, and `make -n sweep` / `make -n sweep-all` expand as intended). Scope is Makefile-only as advertised.

Three minor, non-blocking observations left as inline comments:
1. The new variable block splits `DEV_PORT ?= 5190` from its `# Export DEV_PORT` comment + `export DEV_PORT` line.
2. `sweep` has no `ensure-intentd-submodule` prerequisite (every other target that `cd $(INTENTD_DIR) && cargo ...` has one), so an uninitialized submodule yields a confusing cargo error.
3. When the glob matches nothing, the loop prints a skip line containing the literal `*` pattern (verified: `make sweep-all WORKSPACES_DIR=/tmp/emptydir` → `[sweep-all] skipping /tmp/emptydir/*/monorepo/packages/intentd (no target/ dir)`, exit 0). Graceful, just slightly confusing output.

None of these block the change.

Makefile
panghy: Fixed in b7f15393 — moved the sweep var block below `export DEV_PORT` so the DEV_PORT declaration stays adjacent to its export comment.

Makefile
panghy: Fixed in b7f15393 — went with the cheap `[ -d $(INTENTD_DIR)/target ]` guard: `sweep` now prints `[sweep] nothing to sweep — packages/intentd/target does not exist` and exits 0, with a comment explaining why there's deliberately no `ensure-intentd-submodule` prerequisite (initializing a submodule just to sweep a target/ that can't exist yet is overkill).

Makefile
panghy: Fixed in b7f15393 — added `[ -d "$$dir" ] || continue;` as the first loop statement. Verified: `make sweep-all WORKSPACES_DIR=/tmp/nonexistent-sweep-test` now prints nothing and exits 0.

Issue #761, opened 2026-07-25
Chief workspace-create proposals can carry a non-existent base branch; failure only surfaces at apply as 'cannot resolve base ref'


panghy:
## Symptom

Chief-of-staff workspace-create proposal renders fine (ProposalCard), but applying it fails with:

    Workspace creation failed: invalid params: cannot resolve base ref 'token-usage-baseline'

The Retry button cannot succeed — the ref does not exist, so every retry fails identically.

## Root cause

Two compounding problems:

1. **The chief agent invents base branch names.** In this case it proposed base branch "token-usage-baseline" for intent-hq/monorepo — a branch that does not exist locally or on the remote (only main + assorted feature branches do). The model appears to have treated the base-branch field as a "name for the new working branch" or simply hallucinated it. Nothing in the ws.app.workspaces.propose*/proposal.show binding docs tells the model the value must be an EXISTING ref, and nothing validates it.
2. **No validation until apply time.** The base ref is first resolved deep in worktree creation (intentd intent-git/src/worktree.rs: refs/remotes/{remote}/{r} -> refs/heads/{r} -> rev-parse, else InvalidParams "cannot resolve base ref"). The proposal flow (daemon bindings -> ProposalCard -> apply) never checks that the ref resolves, so the user only finds out after clicking Apply, with no guidance.

## Proposed fix (layered)

- **intentd (binding docs + validation):** document on the propose_create/workspaces binding surface that baseRef/branch must be an existing ref and defaults to the repo's default branch when omitted. Where the daemon knows the local repo path at propose time, best-effort validate the ref (same 3-spec resolution as worktree.rs) and reject/annotate the proposal with a warning instead of deferring failure to apply.
- **FE (graceful degradation):** the Create workspace modal already has a Base branch selector — when the proposal's branch fails to resolve against the repo's known branches, surface the mismatch up-front (warning state + preselect the repo default branch) rather than letting Apply fail. Keep the raw proposed value visible so the user sees what the agent asked for.
- **Error affordance:** on "cannot resolve base ref" failures, the dialog should point the user at the Base branch field instead of offering a Retry that cannot work.

## Environment

- monorepo 27ad67ed (intentd 7a83f37b, cloudlands-fe 552bf9b5) — observed while dogfooding the freshly merged turn-attachment/card-registry stack (rendering side works; this is the apply path).

## Dedup

[...]

PR #784, opened 2026-07-25
docs: document base-ref-unresolvable error.data in PROTOCOL.md (#761)

panghy:
Documents the structured error payload for unresolvable base refs in the canonical wire contract, per the follow-up flagged on cloudlands-fe#388 and tracked from #761.

## Contract

When `workspace.create` fails because the base ref cannot be resolved, the daemon keeps the existing `-32602` code and human message but adds:

```json
"error": {
  "code": -32602,
  "message": "invalid params: cannot resolve base ref 'token-usage-baseline'",
  "data": { "code": "base-ref-unresolvable", "baseRef": "token-usage-baseline" }
}
```

so clients detect the condition from `data.code` instead of matching error prose.

## Changes (docs-only, `docs/PROTOCOL.md`)

- **§9 error-code table**: the `-32602` row now documents the `workspace.create` base-ref failure `error.data` shape, following the `-32005` conflict-`data` precedent in the same table.
- **§3.3 error response**: the `error.data` note now mentions structured machine-readable payloads (conflict object, base-ref failure data) alongside the `-32603` message passthrough.
- **§5.1 worktree provisioning**: the existing "unresolvable `baseRef` … fails with `-32602`" sentence now cross-references the `base-ref-unresolvable` payload (§9).

The daemon (intentd) and FE (cloudlands-fe) halves of this contract land in parallel PRs in their own repos; the wire shape documented here is the pinned contract all three follow.

Refs #761.


panghy (commented):
Reviewed the docs-only diff against the pinned wire contract (`-32602` retained, human message unchanged, additive `error.data = { "code": "base-ref-unresolvable", "baseRef": "<ref>" }`) and against PROTOCOL.md conventions.

**Accuracy** — the §9 row matches the pinned contract exactly and correctly follows the `-32005` conflict-`data` precedent in the same table. The §5.1 cross-reference and the §3.3 `error.data` note are consistent with it.

**Coverage** — I searched the PR-branch doc for every other mention of this failure mode (`base ref`, `unresolvable`, `cannot resolve`, `-32602`): the three touched spots (§3.3 line 128, §5.1 lines 432–433, §9 line 3931) are the only places that reference the `workspace.create` base-ref failure. The `workspace.create` table row in §5.1 defers to the worktree-provisioning prose, so nothing is missed.

Three non-blocking comments inline: a typo introduced in §3.3, a `data`-shape notation inconsistency with the `-32005` row, and a precision suggestion on which form of the ref `data.baseRef` carries given the §5.1 baseRef-canonicalisation note.

docs/PROTOCOL.md
```diff
@@ -125,7 +125,7 @@ All application messages are **JSON-RPC 2.0** text frames. The handler is transp
 { "jsonrpc": "2.0", "id": 1, "error": { "code": -32602, "message": "Missing required parameter: noteId" } }
 ```
 
-`error.data` is optional and carries extra context (e.g. the original internal error message for`-32603`). See §9 for the code table.
+`error.data` is optional and carries extra context — the original internal error message for`-32603`, or a structured machine-readable payload (e.g. the `-32005` conflict object, or the`workspace.create` base-ref failure data). See §9 for the code table.
```
panghy: Nit (typo, introduced here): "or the`workspace.create` base-ref failure data" is missing a space between "the" and the backtick. The same line carries the pre-existing "for`-32603`" missing space — since this line is being touched anyway, both could be fixed in one pass.

docs/PROTOCOL.md
```diff
@@ -3927,7 +3928,7 @@ Errors use the standard JSON-RPC 2.0 `error` object `{ code, message, data? }`.
 | -32700 | Parse error | Body is not valid JSON. Always answered (id null), even for would-be notifications. |
 | -32600 | Invalid Request | Not an object, jsonrpc !== "2.0", missing/empty method, or bad id type. |
 | -32601 | Method not found | Unknown method (only for requests; unknown notifications are dropped). |
-| -32602 | Invalid params | Missing required param ("Missing required parameter: <name>"), bad workspaceId ("workspaceId is required"), non-array where an array is required, "not found" lookups, unauthorized repoPath, etc. |
+| -32602 | Invalid params | Missing required param ("Missing required parameter: <name>"), bad workspaceId ("workspaceId is required"), non-array where an array is required, "not found" lookups, unauthorized repoPath, etc. A `workspace.create` failure from an unresolvable base ref keeps this code and its human message but adds `error.data = { "code": "base-ref-unresolvable", "baseRef": "<the ref>" }` so clients detect the condition from `data.code` instead of parsing the message (§5.1 worktree provisioning). |
```
panghy: Nit (notation consistency): the `-32005` row two lines below writes its shape as `error.data = { code: "conflict", current }` (unquoted keys), while this row uses JSON-quoted keys (`{ "code": ..., "baseRef": ... }`). Aligning on one notation within the table would read better; not blocking either way.

Separately, a precision suggestion: `"<the ref>"` leaves it ambiguous whether `data.baseRef` is the caller's literal input or the canonical form. Per the §5.1 baseRef-canonicalisation note, `workspace.create` strips known remote prefixes (`origin/` etc.) before provisioning, and the daemon's error site (intent-git `worktree.rs`, `cannot resolve base ref '{r}'`) interpolates that post-canonicalisation ref — so a caller passing `origin/nonexistent` would presumably get `baseRef: "nonexistent"`, not their literal input. Since the whole point of the structured payload is that clients stop parsing prose, stating explicitly that `baseRef` is the canonical (prefix-stripped) ref — the same value interpolated into the human message — would remove the last ambiguity. Worth pinning in the intentd PR's tests too.

docs/PROTOCOL.md
```diff
@@ -429,7 +429,8 @@ slugified `repositoryName` (basename fallback) — checked out on the workspace
 any rev-parsable spec, else `HEAD`. No network fetch is performed — the base resolves from
 local state. The returned `Workspace` carries `worktreePath`, `baseCommitSha` (the
 checked-out tip), and `checkoutMode` (`"worktree"` here; see the CoW note below). An
-unresolvable `baseRef` on a valid repo fails with `-32602`.
+unresolvable `baseRef` on a valid repo fails with `-32602` carrying the
+`base-ref-unresolvable` `error.data` payload (§9).
```
panghy, line 433: The cross-reference to §9 is the right structure (matches how the `-32005` conflict data is documented once in the table and referenced elsewhere). Reads fine.

docs/PROTOCOL.md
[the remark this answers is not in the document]
panghy: Fixed in fcb63245 — §3.3 now says `error.data` for `-32603` *may* carry the original internal message (not guaranteed; many shims pass it through as `message` directly), matching the §9 `-32603` row.

docs/PROTOCOL.md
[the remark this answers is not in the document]
panghy: Fixed in fcb63245 — the row now says `error.data.code`.

docs/PROTOCOL.md
panghy: Fixed in fcb63245 — the whole sentence was rewritten (for the other review thread on this line) and now has proper spacing throughout, including the pre-existing missing space before the -32603 code span.

docs/PROTOCOL.md
panghy: Both taken in 53369a53 — the row now uses the -32005 notation (`error.data = { code: "base-ref-unresolvable", baseRef }`) and explicitly pins `baseRef` as the canonical (remote-prefix-stripped) ref, the same value interpolated into the human message, cross-referencing the §5.1 canonicalisation note. I'll flag to the coordinator that the intentd PR's tests should pin the canonical-form behavior too.

PR #809, opened 2026-07-26
docs: PROTOCOL §5.3 documents Round 15 comment semantics (intentd#541)

panghy:
Docs-only follow-up closing the doc drift flagged by the Round 15 verifier: PROTOCOL.md §5.3 did not yet describe the comment-anchoring semantics that landed in intent-hq/intentd#541 (merged as intentd `7f46466`, pulled into this repo via #808).

## What changed

Adds one paragraph to §5.3 (“Overlapping ranges + phantom-marker scrub (intentd#541)”) documenting the current daemon behavior:

- **Overlap acceptance** — a `comment.add` target span may contain other comments' `<!--anchor:…-->` markers; interleaved pairs (`a:start … b:start … a:end … b:end`) are valid content and stay healthy (the overlap rejection in `find_and_anchor_text` was removed).
- **Phantom scrub in the recovery pass** — after per-comment classification, `reanchor_note_comments` scrubs UUID-format markers whose id has no live (non-orphaned) comment row (phantom debris and stale-orphan debris alike, per `live_comment_ids`), so polluted notes self-heal on the next content-changing `note.*` mutation.
- **`comment.add` pre-scrub** — the same scrub runs on the fetched note content before matching, so phantom debris can no longer block a new comment; the cleaned content persists only as part of the add's atomic rewrite.
- **Stored-field stripping** — `anchorText` / `anchorBefore` / `anchorAfter` are stripped of all marker substrings via `strip_anchor_marker_text`, with markers stripped from the full prefix/suffix before the 50-char context window is taken (the PR #541 review fix).
- **Doc-literal exemption** — non-UUID marker-lookalikes (e.g. the literal `<!--anchor:{id}:start-->`) are ordinary user content: commentable, never scrubbed; this also motivates the canonical-hyphenated-UUID requirement on client-supplied `commentId`.

## References

- Implementation: intent-hq/intentd#541
- Submodule bump that brought it into this repo: #808
- Wording checked against `crates/intent-services/src/note_ops.rs` (`scrub_phantom_anchor_markers`, `strip_anchor_marker_text`, `find_and_anchor_text`) and `crates/intent-services/src/lib.rs` (`reanchor_note_comments`, `live_comment_ids`, the `comment.add` pre-scrub) at `7f46466`.

No code changes; `packages/intentd` gitlink untouched.

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -849,6 +849,33 @@ markers scrubbed from the persisted content and the comment is flipped to
 `isOrphaned: bool` field (omitted when unset, `true` for orphaned comments,
 `false` explicitly when a previously-orphaned comment heals).
 
+**Overlapping ranges + phantom-marker scrub (intentd#541).** Overlapping
+comment ranges are allowed: a `comment.add` target span may contain other
+comments' `<!--anchor:…-->` markers, producing interleaved pairs
+(`a:start … b:start … a:end … b:end`) that are valid note content — each
+comment's own id still pins its markers, and interleaved anchors stay
+healthy. The add embeds the raw span (contained markers intact, in place)
+back between the new pair, while the STORED `anchorText` / `anchorBefore` /
+`anchorAfter` fields are stripped of all `<!--anchor:…-->` substrings —
+markers are stripped from the full prefix/suffix before the 50-character
+context window is taken, so a marker adjacent to the span cannot leak a
+clipped fragment — and raw marker text never appears in comment rows. The
+recovery pass additionally scrubs **phantom markers**: after the per-comment
+classification above, any UUID-format marker whose id has no live
+(non-orphaned) comment row — an id with no comment row at all, or markers
+left behind by a row already flagged `isOrphaned` — is removed from the
+persisted content, so a polluted note self-heals on its next content-changing
+`note.*` mutation. `comment.add` runs the same scrub on the fetched note
+content before matching, so phantom debris can never block a new comment; the
+cleaned content persists only as part of the add's atomic note rewrite (a
+failed add changes nothing — no separate rev bump). Non-UUID
+marker-lookalikes (documentation literals such as
+`<!--anchor:{id}:start-->`) are ordinary user content: commentable, and never
+scrubbed. This is also why a client-supplied `commentId` must be a
+**canonical hyphenated** UUID — the scrub only recognizes canonical ids
+inside markers, so a looser spelling would mint markers the recovery pass
+could never classify or clean up.
+
 **Note rewrite visibility (monorepo#638).** Because `comment.add` rewrites the
 note markdown (anchor-marker insertion is an `update_note` that bumps the
 note's `rev`), the result echoes the authoritative post-rewrite revision as
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -828,7 +828,7 @@ each recompute so the read path is O(1) and survives restart. `note.delete` casc
 
 | Method | Params | Result |
 | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


panghy (commented):
Verified every claim in the new §5.3 paragraph against the merged intentd#541 implementation at `7f46466` (the exact gitlink currently tracked by this repo). All claims check out; no contradictions with the surrounding §5.3 paragraphs. Two non-blocking nits below.

## Claim-by-claim verification (intentd @ `7f46466`)

- **Overlap acceptance / interleaved pairs stay healthy** — ✅ `find_and_anchor_text` has no marker-overlap rejection; the plaintext projection strips `<!--…-->` and the overlap comment in `plaintext_fallback_anchor` (note_ops.rs:852–857) states exactly this contract. `classify_anchor_state` keys on each comment's own id, so `a:start … b:start … a:end … b:end` classifies Healthy for both. Test: `comment_add_overlapping_ranges_interleave_and_survive_roundtrip` asserts the exact interleaved layout and that both comments survive a `note.edit` round-trip un-orphaned.
- **Raw span embedded intact between the new pair** — ✅ lib.rs:10967, 10983–10989 (`anchored_span` = raw `content[from..to]`, re-embedded verbatim).
- **Stored-field stripping, full prefix/suffix before the 50-char window** — ✅ lib.rs:10968 (`strip_anchor_marker_text` on `anchorText`) and lib.rs:10979–10982 (`strip_anchor_marker_text` on the FULL prefix/suffix, then `context_before`/`context_after`), matching the "PR #541 review" fix the paragraph cites. Test asserts no clipped `:start-->`/`:end-->` fragments in either adjacency direction.
- **Phantom scrub in the recovery pass, after per-comment classification** — ✅ `reanchor_note_comments` (lib.rs:2830–2909): builds `live_comment_ids` (rows with `is_orphaned != Some(true)`), removes ids flipped orphaned during the pass, then runs `scrub_phantom_anchor_markers` last. Covers both debris classes the paragraph names (no row at all; stale-orphan leftovers). Called from all five content-changing `note.*` paths (lib.rs:9705/9776/9846/9915/10008), matching "self-heals on its next content-changing `note.*` mutation". Test: `note_edit_scrubs_phantom_and_orphan_markers_keeps_live`.
- **`comment.add` pre-scrub, cleaned content persists only via the atomic rewrite** — ✅ lib.rs:10944–10956 scrubs the fetched content in memory before `find_and_anchor_text`; the only persist is `update_note_with_comment` (single transaction, lib.rs:11040), so a failed add changes nothing — consistent with the "Note rewrite visibility" paragraph's atomicity claim. Test: `comment_add_scrubs_phantom_markers_before_anchoring`.
- **Doc-literal exemption** — ✅ `parse_uuid_anchor_marker` requires a canonical UUID id (`is_canonical_uuid`, 8-4-4-4-12), so `scrub_phantom_anchor_markers` preserves lookalikes byte-for-byte (note_ops.rs:1069–1072). Tests: `scrub_never_touches_non_uuid_marker_lookalikes`, `comment_add_doc_literal_marker_is_commentable_and_survives` (commentable via the exact-match path, survives the scrub pass).
- **Canonical-hyphenated `commentId` rationale** — ✅ mirrors the gate and its code comment at lib.rs:10928–10941 (`is_canonical_uuid`, not `Uuid::parse_str`).

Tests run locally against the workspace submodule at `7f46466`: `cargo test -p intent-services -- comment_add note_edit_scrubs note_update_scrubs scrub_ anchor` → **58 passed, 0 failed** (includes all the Round-15 tests above; the WSS e2e `comment_add_scrubs_phantoms_and_overlaps_over_wss` was not re-run here — it passed in intentd#541 CI).

## Non-blocking nits (no change required to merge)

1. **§5.3 `comment.add` table row (line 831) is now slightly loose**: it says a "non-UUID value" for `commentId` is rejected, but since #541 the gate is stricter — any non-*canonical-hyphenated* spelling (e.g. the 32-hex simple form, which `Uuid::parse_str` would accept) is also rejected. The new paragraph states this correctly two paragraphs below, so §5.3 as a whole is accurate; tightening the table row to "non-canonical-UUID value" would remove the ambiguity. Pre-existing wording, not introduced by this PR.
2. **"could never classify or clean up"**: strictly, `classify_anchor_state` is id-string-agnostic and could still classify a non-canonical id while its row exists — it is the phantom *scrub* that can never recognize such markers once the row is gone. The doc wording mirrors the implementation's own comment (lib.rs:10929–10932) verbatim, so this is a shared imprecision, not drift.

No stale claims elsewhere in §5.3 are invalidated by the new semantics (checked "Anchor resilience", "Note rewrite visibility", "Tolerant anchoring", "Reply anchoring", and the method table). Approving. Not merging per instructions.

panghy:
Addressed both review nits: tightened the comment.add table row to say non-canonical-UUID values (e.g. the 32-hex simple form) are rejected, and reworded the rationale sentence to attribute the limitation to the phantom scrub (which cannot recognize non-canonical marker ids once the row is gone) rather than the recovery pass classification.

PR #817, opened 2026-07-26
docs: record local wall-clock timezone semantics for stats.getUsage (D12)

panghy:
Documents the D12 local wall-clock bucketing semantics in the canonical wire contract and bumps the intentd submodule to the merged implementation.

## Changes

- **docs/PROTOCOL.md §5.36 "Timezone semantics"** — rewritten to the D12 contract:
  - Buckets stay UTC hour floors, now carrying a local wall-clock stamp (`local_date` / `local_hour`) captured from the daemon's system timezone when the bucket row is first inserted (later writes folding into the same bucket keep the first-writer's stamp).
  - `month`/`year` period filtering, hour-of-day / month grouping, and `availablePeriods` follow the recorded stamp — "1pm" means 1pm on the daemon's machine at the moment the activity was recorded, immune to later DST transitions or timezone moves.
  - Rows with NULL stamps (local offset indeterminate at record time; pre-migration rows are backfilled by migration 0057) or malformed stamps fall back to shifting `bucket_utc` by `tzOffsetMinutes`.
  - The `24h` period is unchanged: an absolute rolling window over the trailing 24 hourly UTC buckets, with `tzOffsetMinutes` affecting only per-bucket hour labels.
- **packages/intentd** — bumped to intentd main `5794b9f` (squash commit of intent-hq/intentd#544, which implements the above).

## References

- Implementation PR: intent-hq/intentd#544

## Verification

- `make check` — clean (fmt, clippy `-D warnings`, build)
- `make test` — all tests pass

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -3551,8 +3551,9 @@ the first-writer's stamp). For `month`/`year` periods, period filtering, hour-of
 grouping, and `availablePeriods` follow that recorded stamp, so "1pm" means 1pm on the daemon's
 machine **at the moment the activity was recorded** — immune to later DST transitions or
 timezone moves. Rows whose stamp columns are NULL (written while the daemon's local offset was
-indeterminate; pre-migration rows are backfilled) or malformed fall back to shifting
-`bucket_utc` by `tzOffsetMinutes`. The `24h` period is an **absolute rolling window** — the
+indeterminate; pre-migration rows are backfilled from `bucket_utc` using the timezone in effect
+at migration time) or malformed fall back to shifting `bucket_utc` by `tzOffsetMinutes`. The
+`24h` period is an **absolute rolling window** — the
 trailing 24 hourly UTC buckets ending at the current hour — unaffected by `tzOffsetMinutes`
 except that per-bucket hour labels are rendered in local time.
 
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -3544,11 +3544,17 @@ wire — recording has no RPC (§6.8). `stats.getUsage` is global: it takes **no
 | --- | --- | --- |
 | stats.getUsage | period (req): "24h" \| "month" \| "year"; key: "YYYY-MM" (req for month) \| "YYYY" (req for year), ignored for 24h; tzOffsetMinutes: integer, minutes east of UTC, default 0, must be within ±840 | UsageStats — -32602 on a bad period/key/tzOffsetMinutes |
 
-**Timezone semantics** — buckets are stored as UTC hour floors. For `month`/`year` periods each
-bucket is shifted by `tzOffsetMinutes` before period filtering and hour-of-day / month grouping,
-so results reflect the client's local calendar. The `24h` period is an **absolute rolling
-window** — the trailing 24 hourly UTC buckets ending at the current hour — unaffected by
-`tzOffsetMinutes` except that per-bucket hour labels are rendered in local time.
+**Timezone semantics** — buckets are stored as UTC hour floors alongside a **local wall-clock
+stamp** (`local_date` `"YYYY-MM-DD"` / `local_hour` 0–23) captured from the daemon's system
+timezone when the bucket row is first inserted (later writes folding into the same bucket keep
+the first-writer's stamp). For `month`/`year` periods, period filtering, hour-of-day / month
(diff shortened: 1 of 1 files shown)


docs/PROTOCOL.md
```diff
@@ -3544,11 +3544,17 @@ wire — recording has no RPC (§6.8). `stats.getUsage` is global: it takes **no
...
-window** — the trailing 24 hourly UTC buckets ending at the current hour — unaffected by
-`tzOffsetMinutes` except that per-bucket hour labels are rendered in local time.
+**Timezone semantics** — buckets are stored as UTC hour floors alongside a **local wall-clock
+stamp** (`local_date` `"YYYY-MM-DD"` / `local_hour` 0–23) captured from the daemon's system
+timezone when the bucket row is first inserted (later writes folding into the same bucket keep
+the first-writer's stamp). For `month`/`year` periods, period filtering, hour-of-day / month
+grouping, and `availablePeriods` follow that recorded stamp, so "1pm" means 1pm on the daemon's
+machine **at the moment the activity was recorded** — immune to later DST transitions or
+timezone moves. Rows whose stamp columns are NULL (written while the daemon's local offset was
+indeterminate; pre-migration rows are backfilled) or malformed fall back to shifting
```
panghy: Non-blocking nit: for backfilled pre-migration rows the stamp is derived by migration 0057 via SQLite's `'localtime'` modifier, i.e. it reflects the daemon's timezone **at migration time**, not at the original record time — so the "at the moment the activity was recorded" guarantee two sentences earlier strictly holds only for rows written after the migration. The migration header justifies this as matching what the old `tzOffsetMinutes`-shifted read rendered (no visible history change), so the current phrasing is defensible; if you want to be fully precise, a parenthetical like "(backfilled from `bucket_utc` using the timezone in effect at migration time)" would close the gap.

panghy (commented):
Reviewed the doc text against the merged intentd implementation at `5794b9f` (intent-hq/intentd#544). The updated §5.36 "Timezone semantics" paragraph accurately matches the code, and the submodule bump is correct. Details:

**Doc ↔ implementation verification**

- **Local wall-clock stamp, first-insert wins** — matches `usage_stats_repo.rs`: the upsert's `ON CONFLICT ... DO UPDATE` deliberately leaves `local_date` / `local_hour` untouched, and migration `0057_usage_stats_local_wallclock.sql` documents the same first-writer semantics. Covered by the repo test asserting "first-writer stamp wins".
- **month/year filtering, hour-of-day / month grouping, and `availablePeriods` follow the recorded stamp** — matches `usage_stats_read.rs::local_parts` + `aggregate_usage`: `in_period`, `by_hour`, `by_month`, and the `months`/`years` sets (computed over all rows) all derive from `local_parts`, which prefers the stamp.
- **NULL / malformed fallback to the `tzOffsetMinutes` shift** — matches: `local_parts` falls back when either column is NULL, `parse_local_date` rejects the date, or `local_hour >= 24`; the repo read layer additionally drops out-of-range hours to `None` (test: "hour 99 must not surface"). NULL stamps are genuinely produced when `recording_local_offset()` (i.e. `UtcOffset::current_local_offset()`) fails — e.g. the `time` crate's multi-threaded soundness guard on Linux — and rows are persisted with NULL stamps rather than a misleading UTC stamp.
- **`24h` unchanged** — matches: the `Last24h` branch filters/slots purely on `bucket_utc` against the UTC-hour-floored rolling window, ignoring the stamp; `tzOffsetMinutes` only renders the per-bucket `hour` labels.
- **Wire-shape bullets** (`byHourOfDay`, `byMonth`, `availablePeriods`) are untouched and remain consistent with the new paragraph — "local" now uniformly means stamp-based local with the documented fallback.

**Submodule ref** — `packages/intentd` now points at `5794b9f5be9cb150eea94d45508339da29ec6842`, which is exactly intentd `origin/main` HEAD (the squash commit of intentd#544). Verified via `git fetch` + `git rev-parse origin/main` in the submodule.

One non-blocking wording nit inline. CI is green (Analyze, CodeQL, pr-title). Not approving/merging per review scope.

docs/PROTOCOL.md
panghy: Addressed in 28434644 — the parenthetical now reads 'pre-migration rows are backfilled from bucket_utc using the timezone in effect at migration time'.

Issue #871, opened 2026-07-27
workspace MCP tool surface vanishes from live agent session after request timeout (-32001) and never re-registers


panghy:
While coordinating in a live workspace session (workspace branch: another-add), the workspace MCP tool call surface disappeared mid-session and never recovered, leaving the coordinator unable to delegate, read/update notes, or check task status.

Sequence observed (coordinator agent, 2026-07-27):

1. A workspace_api call that performed two ws.agent.delegate(...) invocations in one script failed with: MCP server "workspace-mcp" is unresponsive (liveness ping failed: MCP error -32001: Request timed out).
2. Immediate retry failed with: MCP client is closing.
3. Every subsequent call across multiple turns failed with: Tool workspace_api_workspace-mcp not found — the tool was removed from the agent toolset entirely and never re-registered, even after long gaps between turns.
4. The daemon itself stayed healthy the whole time: the first of the two delegations from step 1 HAD landed (the child implementor agent ran to completion and its settlement event was delivered to the coordinator conversation), and workspace events kept flowing.

Two distinct problems:

- No reconnect/re-registration: after a transient MCP timeout, the agent-side MCP client shuts down and the workspace tool surface is permanently gone for the rest of the session. Expected: the bridge/client reconnects (or the tool stays registered and returns retryable errors).
- Partial application of multi-call scripts: a workspace_api script that makes several mutating calls can time out after some calls have applied (1 of 2 delegations landed), with the caller receiving only a generic timeout. The coordinator cannot tell what applied without listing agents afterwards — which it also cannot do once the tools are gone. Consider making the timeout error surface which calls completed, or encouraging/enforcing one mutation per script.

Impact: coordinator dead in the water mid-plan; work continues only via events from already-spawned children. Recovery appears to require restarting the agent session (unverified).

panghy:
Daemon-side evidence from the same 2026-07-27 dev session (`make dev`, workspace `another-add`), gathered while investigating this from another workspace:

**Agent log** (`agent-logs/agent-d5e0cfca-5148-445e-8455-2f9f056c0721/2026-07-27.log`) shows both MCP servers report initialized, then the workspace-mcp transport dies:

```
🔌 Waiting for 2 MCP server(s) to initialize...
   ✅ Endara
   ✅ workspace-mcp
⚠️ MCP server startup error: MCP error -32001: Request timed out
[...]

Issue #616, opened 2026-07-24
Delegated agent spawn always fails first attempt: handshake failed: request `initialize` timed out (recovers on restart)


panghy:
## Symptom

While dogfooding the Intent stack (coordinator agent delegating release-ops tasks, 2026-07-24, workspace `kick-off`), **every freshly spawned delegated agent failed once at startup** with:

```text
internal error: handshake failed: request `initialize` timed out
```

Reproduction rate in this session: **6 of 6** fresh spawns failed on the first attempt — 4× `ws.agent.delegate` (implementor specialist), 1× restart-created spawn, 1× `ws.agent.create` (verifier specialist). In **every case**, waking/restarting the errored agent (`ws.agent.wakeOrCreate(taskNoteId, ...)` or `ws.agent.send(agentId, ...)`) succeeded immediately and the agent completed its task normally.

## Observed details

Example failed agent (`agent-d3018f57-5414-4a9d-be8b-df3533d4bc21`):

- `status: "error"`, `stopReason: "internal error: handshake failed: request `initialize` timed out"`
- `messageCount: 1` (only the initial delegation message), no assistant output, no tool calls
- `createdAt 00:50:37Z` → `lastActivity 00:51:01Z` (~23s), consistent with a fixed initialize-handshake timeout on the ACP session spawn
- Failure fires before the agent does any work, so retrying is safe — but the parent coordinator only sees a generic child-failed event and must manually resuscitate each agent

Other affected agents in the same session: `agent-08a0bc52`, `agent-7a61ad9e`, `agent-f316a5c6`, `agent-12ee0fd0`, `agent-b48ce20f` — identical signature each time.

## Impact

P1: delegation still works, but only with a manual restart per agent — every wave of delegation degrades into fail → notice → wake. For multi-task delegation flows this doubles coordination overhead and wakes the parent agent with spurious failure events.

## Notes / hypotheses

- The 100% first-attempt failure + 100% success-on-retry pattern suggests a cold-start cost on first spawn (e.g. provider binary fetch/launch, `npx` resolution, or process warm-up) exceeding the `initialize` request timeout, with subsequent attempts hitting a warm path.
- Possible directions: raise/scale the initialize handshake timeout for fresh spawns, add an automatic one-shot retry on handshake timeout before surfacing the error, or pre-warm the provider process.

## Dedup

[...]

PR #956, opened 2026-07-27
docs: document async sandbox provisioning and pending effectiveIsolation

panghy:
## Summary

Phase-2 monorepo follow-up to intent-hq/intentd#636 (fix: decouple CoW sandbox provisioning from the delegate critical path), which addressed #871 (see also #616).

> Note: the intentd submodule bump to `d16d910` (picking up intent-hq/intentd#636) already landed on main via #955; after rebasing, this PR carries the documentation changes only.

### Docs

**`docs/PROTOCOL.md`**
- §5.5 `agent.delegate`: documented the new asynchronous provisioning contract — an eligible CoW request now returns `effectiveIsolation: "pending"` immediately (the only value the field carries today; `"cow"` / `"direct"` are no longer returned, per the merged code in `agent_ops.rs` and the wire-contract tests). The field stays omitted when no CoW isolation was resolved. The settled outcome is observable via the session's sandbox fields and the `sandbox:created` event; fallback-to-shared on unsupported filesystems / provisioning failure is unchanged (log-only, no sandbox fields, no event). Also documented the `WORKSPACE_API_TIMEOUT` default (30s) and its `INTENTD_WORKSPACE_API_TIMEOUT_MS` env override.
- §5.5a `sandbox.*` intro and the §6.5 event-taxonomy `sandbox:created` row: noted provisioning is asynchronous relative to the delegate result.

**`docs/ARCHITECTURE.md`**
- "Workspace checkouts & agent sandboxes (CoW)" → Agent sandboxes: described the background provisioning task, the per-agent settlement gate awaited before the child's first ACP spawn, and the unchanged fallback-to-shared semantics.

### FE audit (report only, no changes)

Grepped `packages/cloudlands-fe` for consumers of the delegate result's `effectiveIsolation` — none found. The only matches are `resolveEffectiveIsolationMode` in the workspace initializer, which resolves the workspace *checkout* mode (worktree/cow) at creation time and is unrelated to the delegate result field. No FE follow-up task needed.

## Verification

- `make check` (fmt + clippy + build) — green
- `make test` — green


docs/PROTOCOL.md
```diff
@@ -3865,7 +3865,7 @@ All filters on a subscription are combined with **AND**. Delivery is gated *only
...
+| sandbox (new in intentd) | sandbox:created, sandbox:merged | Emitted when `agent.delegate` with effective `isolation: "cow"` successfully provisions a CoW sandbox (§5.5 — asynchronously: provisioning runs in a background task after the delegate result returns `effectiveIsolation: "pending"`) and when sandbox commits are successfully merged back to the canonical repository (§5.5a — auto-merge on completion or manual `sandbox.merge`). `sandbox:created` → data { workspaceId, agentId, sandboxPath, branch, baseCommitSha, snapshotCommitSha } where `sandboxPath` is the absolute filesystem path to the sandbox clone, `branch` is the sandbox snapshot branch (`sb/<agentId>`), `baseCommitSha` is the sandbox HEAD at provisioning, and `snapshotCommitSha` is the WIP-snapshot commit SHA (`null` when the source was clean). `sandbox:merged` → data { workspaceId, agentId, commitRange, canonicalHead } where `commitRange` names the applied sandbox commit range and `canonicalHead` is the canonical repository HEAD SHA after the merge. Both are self-sufficient payloads (§6.7). |
```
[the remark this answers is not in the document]
panghy: Addressed — reworded the §6.5 row to say the event fires when the resolved isolation mode is \cow\ on a sandbox-eligible workspace and the background provisioning task succeeds, and noted explicitly that this is distinct from the effectiveIsolation result field (which only ever reports \pending\).

docs/PROTOCOL.md
```diff
@@ -1072,7 +1072,7 @@ The largest namespace. Every `agent.*` method is served daemon-primary by `inten
...
...rect"`, setting disabled, the worktree-mode ineligibility skip, or no sandbox source). The settled outcome is observable rather than returned: on success the child session's TOP-LEVEL fields (not nested under metadata) `sandbox_id`, `sandbox_path`, and `sandbox_branch` are persisted and served in both `AgentSession` and `AgentLite`, and a `sandbox:created` event is emitted with `data { workspaceId, agentId, sandboxPath, branch, baseCommitSha, snapshotCommitSha }`; when the filesystem does not support CoW reflinks — or provisioning fails — the daemon falls back to shared (`"direct"`) mode exactly as before (no bytes copied, log-only): the session keeps no sandbox fields and no `sandbox:created` fires. The child's first ACP spawn is gated on settlement: its turn worker awaits the in-flight provisioning before spawning, so the child never runs against a half-copied sandbox. The sandbox directory is never auto-cleaned — cleanup is the responsibility of higher-level orchestration. All agent file/exec/terminal/search operations are restricted to the sandbox path when present (logical containment guards in `intent-services`), preventing escape to the main workspace or parent directories |
```
panghy: The retained closing sentence — "The sandbox directory is never auto-cleaned — cleanup is the responsibility of higher-level orchestration" — is no longer strictly true after intentd#636. Because the clone now races `agent.delete`, `settle_provisioned_sandbox` (`agent_ops.rs` L3466–3565) **discards the just-provisioned sandbox (directory + store record)** when the child session is gone or soft-deleted by settlement time, instead of stranding a multi-GB clone on disk.

This also carves a third settlement outcome out of the "settled outcome is observable" claim: in the delete race, provisioning *succeeded* but no sandbox fields persist and no `sandbox:created` fires — indistinguishable from the shared-mode fallback to a client watching for settlement. Suggest amending along the lines of: "The sandbox directory is never auto-cleaned once settled — except when `agent.delete` races the background clone, in which case the daemon discards the just-provisioned sandbox (directory + record; no sandbox fields, no `sandbox:created`) — cleanup is otherwise the responsibility of higher-level orchestration."

docs/ARCHITECTURE.md
```diff
@@ -149,7 +149,18 @@ Wire contract: PROTOCOL.md §5.1 (`checkoutMode`, `cowSupported`), §5.5/§5.5a
...
   that same directory (agent commits land in the workspace's own checkout, never the
   user's repo folder). Worktree-mode workspaces are ineligible (agents share the
-  checkout).
+  checkout). **Provisioning is asynchronous** relative to `agent.delegate`
+  (PROTOCOL §5.5): the delegate path registers a per-agent settlement gate
+  (`Services::begin_sandbox_provisioning`), spawns the clone in a background task,
+  and returns `effectiveIsolation: "pending"` immediately — a large clone can take
+  tens of seconds, which previously starved the `workspace_api` MCP budget. The
+  background half settles the outcome onto the child's session (sandbox fields +
+  `sandbox:created` event) and releases the gate via a drop guard, so it settles
```
panghy, line 158: Same nuance as the PROTOCOL comment: the background half has a third settlement outcome besides success and fallback-to-shared. When `agent.delete` races the clone, `settle_provisioned_sandbox` finds the session missing/soft-deleted and discards the sandbox (removes the directory and best-effort deletes the store record) rather than persisting fields/emitting the event (`agent_ops.rs` L3466–3565). Since this paragraph enumerates what the background half does, a short clause covering the delete-race discard would make it complete.

docs/PROTOCOL.md
panghy: Nit: "provisioning runs in a background task **after** the delegate result returns `effectiveIsolation: \"pending\"`" — the background task is actually spawned *before* the delegate op returns (`agent_ops.rs` L3298; the gate must be registered first so the child's turn worker observes it). It's the *settlement* that lands after the result. Consider "provisioning settles in a background task; the delegate result reports `effectiveIsolation: \"pending\"`" to match the §5.5a phrasing, which gets this right.

panghy (commented):
Reviewed the doc claims against the merged intentd code at `d16d910` (intent-hq/intentd#636). The core contract described in this PR checks out:

- **`effectiveIsolation` is "pending"-or-omitted** — verified in `crates/intent-services/src/agent_ops.rs` (`agent_delegate_op`): `effective_isolation` is only ever set to `Some("pending")` (L3270) inside the eligible-CoW branch, and the field is inserted into the result only when set (L3406–3411). No code path returns `"cow"` or `"direct"` anymore. The wire-contract test `delegate_returns_effective_isolation_in_result` (`intent-services/src/tests.rs` L14033) asserts exactly this, and the WSS e2e suite (`crates/intentd/tests/e2e_wss_workspace_cow.rs` L747/L954/L1071) asserts `"pending"` on the success, concurrency, and failure-fallback paths.
- **Settlement gate + background task + drop guard** — `Services::begin_sandbox_provisioning` / `settle_sandbox_provisioning` / `await_sandbox_provisioning` exist as described (`lib.rs` L17445–17483); the gate is registered before the delegate result returns, the clone runs in a `tokio::spawn`, and the `SettleGuard` drop impl settles the gate even on panic/unpolled-drop (`agent_ops.rs` L3282–3303).
- **Per-agent gate awaited before first ACP spawn** — `AgentManager::ensure_started` calls `self.services.await_sandbox_provisioning(agent_id).await` before reading the session and resolving the spawn cwd (`agent_manager.rs` L3894), matching the ARCHITECTURE claim.
- **`WORKSPACE_API_TIMEOUT` / `INTENTD_WORKSPACE_API_TIMEOUT_MS`** — `dispatch.rs` L22–42: 30s default, override parsed as a positive integer in milliseconds (unset/non-numeric/non-positive keep the default), and `default_workspace_api_timeout()` is read at MCP-server construction (`mcp_server.rs` L75). Matches the §5.5 text exactly.
- **Fallback-to-shared unchanged** — `provision_delegate_sandbox` (`agent_ops.rs` L3422–3464): `Unsupported` → debug log, `Err` → warn log, no sandbox fields, no event, in both cases.
- **FE audit** — confirmed: no consumer of the delegate result's `effectiveIsolation` in `packages/cloudlands-fe`.
- Grepped both docs for other stale mentions of the synchronous contract — the four spots this PR touches are the only ones.

Two inline comments below on a retained sentence that #636 made slightly inaccurate (the delete-race sandbox discard) — worth tightening while this section is being touched. Not blocking.

docs/PROTOCOL.md
panghy: Addressed in fba1948f — verified settle_provisioned_sandbox (agent_ops.rs L3466+): added the delete-race third settlement outcome (daemon discards the just-provisioned sandbox directory + store record; no sandbox fields, no sandbox:created) and reworded the closing sentence to 'never auto-cleaned once settled — except for the delete-race discard above'.

docs/ARCHITECTURE.md
panghy: Addressed in fba1948f — added the delete-race discard clause to the Agent sandboxes bullet (settle_provisioned_sandbox removes the sandbox directory and best-effort deletes the store record when the session is missing/soft-deleted, instead of persisting fields or emitting the event).

docs/PROTOCOL.md
panghy: Already addressed in 1b907e09 (pushed before this comment landed): the §6.5 row no longer says 'after' — it now reads 'the background provisioning task succeeds (§5.5 — asynchronous: the delegate result itself only ever reports effectiveIsolation: "pending" ...)'. The §5.5 row also states the gate is registered and the clone kicked off before the result returns.

Issue #991, opened 2026-07-28
host.providerDiscovery does not expose secondary-binary status for dual-binary providers (unsloth)


panghy:
Follow-up from intent-hq/intentd#653 (which fixed #935 for `intentd doctor`).

Provider discovery now records the secondary-binary resolution status for dual-binary providers (`ProviderAvailability.secondary_binary`, populated for providers with `ProviderConfig.requires_secondary_binary` — currently unsloth: `opencode` + `unsloth`), and doctor uses it to name the actually-missing binary.

However, the `host.providerDiscovery` RPC payload (`discover_providers_with_npx` in `crates/intent-services/src/lib.rs`) does not expose this field, so RPC/FE consumers see only `installed: false` with `command: "opencode"` and cannot attribute unavailability to the actually-missing binary — the same misattribution #935 fixed for doctor can still surface in any client UI built on `host.providerDiscovery`.

Expected: expose the secondary-binary status (e.g. `secondaryCommand` / `secondaryResolved`, or an equivalent attribution string) in the `host.providerDiscovery` provider entries, and document it in `docs/PROTOCOL.md`.

Noted as a non-blocking scope observation during review of intent-hq/intentd#653.

Refs: intent-hq/monorepo#935, intent-hq/intentd#653, intent-hq/intentd#622.

PR #1006, opened 2026-07-28
docs: document host.providerDiscovery incl. secondary-binary fields

panghy:
References #991 (closed by intent-hq/intentd#668).

## What

Documents the previously-undocumented `host.providerDiscovery` method in `docs/PROTOCOL.md` (new subsection after `host.providerAuthStatus`), including:

- The full response payload: `providers[]` (id, displayName, command, installed, optional `resolvedPath`/`gatedOff`/`npxPackage`, `hasNpxFallback`, `npxOnly`) and the `npx` probe object.
- The **additive** `secondaryCommand` / `secondaryResolved` attribution fields for dual-binary providers (unsloth: `opencode` + `unsloth`), which landed in intent-hq/intentd#668 — emitted together, **omitted (never null)** for single-binary providers and for gated-off providers (never probed). Clients detect by presence.

Docs-only; no code changes, no submodule bump (batched separately).


docs/PROTOCOL.md
```diff
@@ -384,6 +384,40 @@ Daemon-owned provider auth probes: reports whether each CLI-backed agent provide
...
+    {
+      "id": "unsloth",
+      "displayName": "Unsloth",
+      "command": "opencode",
+      "installed": false,
+      "resolvedPath": "/usr/local/bin/opencode",  // optional — present when the primary binary resolved
+      "gatedOff": "requires env var SOME_VAR",     // optional — present when the provider is gated off
+      "hasNpxFallback": false,
+      "npxOnly": false,
+      "npxPackage": "@some/pkg",                   // optional — present for npx-only providers
```
panghy: The example shows `npxPackage` alongside `"npxOnly": false`, but per the emission code these are coupled: `npxOnly` is `npx_only_package.is_some()` and `npxPackage` is emitted iff that same option is `Some` — so `npxPackage` is present **iff** `npxOnly` is `true`. As written the example depicts an impossible combination. Suggest changing the comment to `// optional — present iff npxOnly is true` and/or setting `"npxOnly": true` in a separate example entry (the unsloth entry itself is not npx-only, so ideally `npxPackage` would not appear on this object at all).

docs/PROTOCOL.md
```diff
@@ -384,6 +384,40 @@ Daemon-owned provider auth probes: reports whether each CLI-backed agent provide
...
+- **`secondaryCommand` / `secondaryResolved`** *(additive, monorepo#991)* — secondary-binary attribution for **dual-binary providers** (today only `unsloth`, which requires both `opencode` and `unsloth` on the daemon host). `secondaryCommand` names the required secondary CLI and `secondaryResolved` reports whether it resolved via the same discovery precedence as the primary — so when a dual-binary provider shows `installed: false`, clients can attribute the failure to the actually-missing binary instead of the primary `command`. The two fields are always emitted **together**, and are **omitted (never null)** for providers without a secondary requirement and for gated-off providers (gated providers are never probed, so no attribution exists). Clients detect by presence.
+- `gatedOff` is present (with a human-readable reason) only when the provider is feature-gated off; gated providers skip binary probing entirely.
+- `npx` reports the daemon's npx probe for npx-fallback providers: `resolvedPath` / `version` are `null` when npx is missing or the version probe fails, and `versionOk` is whether the resolved version meets the minimum requirement.
```
panghy: Slight imprecision: `resolvedPath` is `null` only when npx itself is missing. When npx resolves but the version probe fails/times out, `resolvedPath` is still populated and only `version` is `null` (with `versionOk: false`) — see `discover_providers_with_npx`: the path is set from `npx_status.resolved_path` independently of the version probe. Suggest: "`resolvedPath` is `null` when npx is missing; `version` is `null` when npx is missing or the version probe fails". Optionally also worth stating the minimum: `versionOk` is npx major version ≥ 7 (`parse_npx_version_ok`).

panghy (commented):
Verified this docs-only change against the merged intentd implementation (`discover_providers_with_npx` in `crates/intent-services/src/lib.rs` and `discover_providers` in `crates/intent-providers/src/discover.rs`, intentd main `0b105bd`, intent-hq/intentd#668).

**Confirmed accurate:**
- Field set and always-vs-optional split match emission exactly (`id`/`displayName`/`command`/`installed`/`hasNpxFallback`/`npxOnly` always; `resolvedPath`/`gatedOff`/`npxPackage`/`secondaryCommand`/`secondaryResolved` conditional).
- Omitted-vs-null semantics for `secondaryCommand`/`secondaryResolved` are correct: emitted together iff `p.secondary_binary` is `Some`, which is `None` for gated-off providers and providers without `requires_secondary_binary` — never `null` on the wire. Only `unsloth` sets `requires_secondary_binary` today (`config.rs`).
- Registry order, `{}` request (handler ignores params), gated-providers-skip-probing, and the resolution-precedence wording (matching the existing `host.providerAuthStatus` bullet and the `discover_providers` doc comment) all check out.
- `npx` object shape matches: `resolvedPath`/`version` are explicit JSON `null` when absent, `versionOk` always boolean.

**Issues found:** two inline comments below (example internal consistency, and the `npx` bullet's null conditions). I also concur with the existing bot comment about `gatedOff` co-existing with `resolvedPath`/`secondary*` in the example — per `discover.rs`, a gated provider emits none of those fields, so the example object depicts an impossible wire shape.

Style is consistent with the surrounding `host.providerAuthStatus` subsection (heading level, Request/Response layout, bold emphasis on semantics).

docs/PROTOCOL.md
```diff
@@ -384,6 +384,40 @@ Daemon-owned provider auth probes: reports whether each CLI-backed agent provide
...
+```jsonc
+{
+  "providers": [
+    {
+      "id": "unsloth",
+      "displayName": "Unsloth",
+      "command": "opencode",
+      "installed": false,
+      "resolvedPath": "/usr/local/bin/opencode",  // optional — present when the primary binary resolved
+      "gatedOff": "requires env var SOME_VAR",     // optional — present when the provider is gated off
```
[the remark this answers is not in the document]
panghy: Addressed in d8e5b200: `gatedOff` is dropped from the probed example, and the bullet now states gated entries never carry `resolvedPath`/`secondaryCommand`/`secondaryResolved` and always report `installed: false`.

docs/PROTOCOL.md
```diff
@@ -384,6 +384,40 @@ Daemon-owned provider auth probes: reports whether each CLI-backed agent provide
...
+      "hasNpxFallback": false,
```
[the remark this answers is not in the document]
panghy: Addressed in b39a62a5: added bullets defining `npxOnly` (npx-launched provider, `npxPackage` present iff true, `resolvedPath` is the npx binary) and `hasNpxFallback` (local binary preferred; when `installed: false` the provider may still be usable if the `npx` probe reports `versionOk: true`).

docs/PROTOCOL.md
panghy: Addressed in d8e5b200: the example is split into a dual-binary entry (no `npxPackage`) and an npx-only claude-code entry with `npxOnly: true` + the pinned `npxPackage` spec, matching the emission coupling.

docs/PROTOCOL.md
panghy: Addressed in d8e5b200: the npx bullet now says `resolvedPath` is null only when npx is missing, a failed version probe leaves it set with `version: null`, and `versionOk` is false whenever `version` is null.

PR #1062, opened 2026-07-29
chore: update intentd and cloudlands-fe submodules for provider path popup fix

panghy:
Bumps both submodules to pick up the provider path popup fix and documents the new wire field.

## Submodule bumps

- **packages/intentd** → `cbc50cf` — intent-hq/intentd#701: surface `secondaryResolvedPath` in `host.providerDiscovery` (dual-binary secondary's resolved absolute path, present only when the secondary resolved; additive wire change).
- **packages/cloudlands-fe** → `6bc0d4fd` — intent-hq/cloudlands-fe#528: settings provider path popup shows full daemon-resolved paths, override state, and both unsloth binaries.

## Docs

- `docs/PROTOCOL.md` §5.14 `host.providerDiscovery`: extended the response sample and dual-binary field notes with the additive `secondaryResolvedPath` field (present only when `secondaryResolved: true`, omitted otherwise, never emitted for gated-off providers).

## Verification

- `make check` (cargo fmt --check, clippy -D warnings) — green
- `make test` (cargo test --workspace) — green

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -400,12 +400,13 @@ Daemon-owned provider discovery: reports which CLI-backed agent providers are in
       "id": "unsloth",
       "displayName": "Unsloth",
       "command": "opencode",
-      "installed": false,
+      "installed": true,
       "resolvedPath": "/usr/local/bin/opencode",  // optional — present when the primary binary resolved
       "hasNpxFallback": false,
       "npxOnly": false,
       "secondaryCommand": "unsloth",               // optional — dual-binary providers only
-      "secondaryResolved": false                   // optional — dual-binary providers only
+      "secondaryResolved": true,                   // optional — dual-binary providers only
+      "secondaryResolvedPath": "/Users/me/.unsloth/bin/unsloth"  // optional — present only when the secondary resolved
     },
     {
       "id": "claude-code",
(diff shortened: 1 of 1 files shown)


panghy:
LGTM — verified as follows:

**Gitlinks**
- `packages/intentd` → `cbc50cf841288d3a08f1a9241333848f1a12d6cd` = squash merge commit of intent-hq/intentd#701, confirmed ancestor of `origin/main` (`git merge-base --is-ancestor` passes).
- `packages/cloudlands-fe` → `6bc0d4fda8fffb9a1905e980a343188e8831d195` = squash merge commit of intent-hq/cloudlands-fe#528, confirmed ancestor of `origin/main`.

**PROTOCOL.md §5.14 accuracy vs. intentd @ cbc50cf**
- `secondaryResolvedPath` is emitted in `crates/intent-services/src/lib.rs` only inside `if let Some(path) = secondary_path` — present only when the secondary resolved, omitted (never null) otherwise. Matches the doc bullet exactly.
- Serialized as `path.display().to_string()`; `find_provider_binary` resolves via native-install dirs / managed bin / PATH scan (explicit overrides must be absolute), and the payload test in `intent-services/src/tests.rs` asserts `Path::new(path).is_absolute()` whenever `secondaryResolved == true` and that the key is absent otherwise. "Absolute path string" claim holds.
- `installed` uses `installed_with_secondary` (discover.rs): true only when the primary resolved AND (no secondary required OR secondary resolved) — so the sample's `installed: true` with both binaries resolved is consistent, with unit tests pinning all four truth-table cases.
- Gated-off providers skip both primary and secondary probing (`secondary_binary = None` when gated), so the updated `gatedOff` bullet (never carries `secondaryResolvedPath`) is accurate.

**Conventions / CI**
- PR title and the single branch commit are valid conventional commits (`chore:`).
- Head-commit checks green: CodeQL ✅, Analyze (actions) ✅, pr-title skipped (nothing failing).

PR #1077, opened 2026-07-29
chore: update intentd submodule for override-aware provider discovery (intentd#717)

### LINKED ISSUE #1065: host.providerDiscovery ignores providers.paths[\unsloth\] override when reporting unsloth installed status
Follow-up from intent-hq/intentd#707 (review comment https://github.com/intent-hq/intentd/pull/707#discussion_r3670612738).

Since intentd#707, `providers.paths["unsloth"]` overrides where the managed-server lifecycle (`unsloth_server.rs`) resolves the `unsloth` CLI. However, provider discovery (`discover_providers` in `crates/intent-providers/src/discover.rs`) is a leaf crate that cannot read settings, so it resolves the secondary binary without the override.

**User-visible consequence:** `host.providerDiscovery` (and the FE's installed badge) reports unsloth as *not installed* even when a valid `providers.paths["unsloth"]` override makes the managed server start fine.

**Feasibility:** there is precedent for applying settings-based path overrides above the leaf crate — `intent-transport/src/host.rs` already reads `providers.paths` from settings for auggie — so the override could be threaded into the discovery surface at the transport/services layer.

Related (closed): #991 (secondary-binary status in `host.providerDiscovery`), #935 (doctor misattribution for unsloth binaries).

panghy:
Wrap-up for the provider-discovery override fix (intent-hq/intentd#717, squash `5ec2aad7`).

## What changed

- **`docs/PROTOCOL.md` §5.14 `host.providerDiscovery`** — documents the override-aware semantics:
  - `installed` / `secondaryResolved` now honor **valid `providers.paths` overrides** (absolute path to an executable file, the same validation as spawn resolution). Override keys follow spawn resolution: for `unsloth` the `opencode` key covers the primary binary and the `unsloth` key covers the secondary CLI.
  - `resolvedPath` / `secondaryResolvedPath` remain **strictly auto-detected** — an override never surfaces there. Consequently `secondaryResolved: true` can now coexist with `secondaryResolvedPath` absent (valid override, nothing auto-detected), and likewise `installed: true` with `resolvedPath` absent.
  - No new wire fields.
- **`packages/intentd` gitlink** → `d4d4e436` (latest main, contains the intentd#717 squash `5ec2aad7`).

## Verification

`make check && make test` from the monorepo root — fmt, clippy (`-D warnings`), build, and the full `cargo test --workspace` suite all green.

Refs intent-hq/intentd#717
Closes #1065

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -387,7 +387,7 @@ Daemon-owned provider auth probes: reports whether each CLI-backed agent provide
 
 #### `host.providerDiscovery`
 
-Daemon-owned provider discovery: reports which CLI-backed agent providers are installed on the daemon host (binary resolution + npx fallback status), so clients render install state without probing `PATH` themselves.
+Daemon-owned provider discovery: reports which CLI-backed agent providers are installed on the daemon host (binary resolution + npx fallback status, honoring valid `providers.paths` overrides), so clients render install state without probing `PATH` themselves.
 
 **Request:** `{}` (no parameters)
 
@@ -401,12 +401,12 @@ Daemon-owned provider discovery: reports which CLI-backed agent providers are in
       "displayName": "Unsloth",
       "command": "opencode",
       "installed": true,
-      "resolvedPath": "/usr/local/bin/opencode",  // optional — present when the primary binary resolved
+      "resolvedPath": "/usr/local/bin/opencode",  // optional — present when the primary binary auto-detected (never an override path)
       "hasNpxFallback": false,
       "npxOnly": false,
       "secondaryCommand": "unsloth",               // optional — dual-binary providers only
       "secondaryResolved": true,                   // optional — dual-binary providers only
-      "secondaryResolvedPath": "/Users/me/.unsloth/bin/unsloth"  // optional — present only when the secondary resolved
+      "secondaryResolvedPath": "/Users/me/.unsloth/bin/unsloth"  // optional — present only when the secondary auto-detected
     },
     {
       "id": "claude-code",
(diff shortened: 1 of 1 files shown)


panghy (commented):
## Review — docs accuracy + gitlink verification

Verified the PROTOCOL.md §5.14 changes against the actual intentd#717 implementation at the bumped commit (`d4d4e436`) and checked the gitlink bump. **Approving.**

### Doc accuracy vs. intentd#717 implementation ✅

Every claim in the updated §5.14 prose traces to code at `d4d4e436`:

- **`installed` / `secondaryResolved` honor valid overrides** — `crates/intent-providers/src/discover.rs` `availability_for()`: `installed` is computed from `resolved_path.is_some() || primary_override.is_some()`, and `SecondaryBinary.resolved` from `auto.is_some() || secondary_override.is_some()`.
- **Same validation as spawn resolution** — overrides go through `resolve_explicit_path()` (trimmed, absolute, executable file), which is the exact function backing `find_provider_binary`'s explicit tier; invalid values (missing/relative/non-executable) return `None` and contribute nothing, matching the doc.
- **Override keys follow spawn resolution** — primary uses `provider.primary_binary_provider_id()` (unsloth's `opencode` primary honors the `opencode` key); secondary uses `override_path(provider.id)` (the `unsloth` key targets the unsloth CLI). Matches the doc sentence verbatim.
- **`resolvedPath` / `secondaryResolvedPath` strictly auto-detected** — `ProviderAvailability.resolved_path` and `SecondaryBinary.resolved_path` are populated only from `resolve_auto(...)`; the override result is never stored. The serializer in `intent-services/src/lib.rs` (`discover_providers_with_npx_overrides`) emits `resolvedPath`/`secondaryResolvedPath` only from those auto-detected fields, so `installed: true` / `secondaryResolved: true` with the path absent is exactly what the wire produces.
- **No new wire fields** — the serializer's field set is unchanged (id, displayName, command, installed, resolvedPath, gatedOff, hasNpxFallback, npxOnly, npxPackage, secondaryCommand, secondaryResolved, secondaryResolvedPath, npx block).
- **Transport threads `providers.paths`** — `intent-transport/src/host.rs` `HostMethod::ProviderDiscovery` arm reads the settings via `read_provider_paths(api)` and passes them into `host_ops::provider_discovery_op`, consistent with the doc's "honoring valid providers.paths overrides" summary line.

### Internal consistency of §5.14 ✅

The section is self-consistent: the summary line, the JSONC example comments ("auto-detected", "never an override path"), and the four bullets (`installed`, `resolvedPath`, `secondaryCommand`/`secondaryResolved`, `secondaryResolvedPath`) all describe the same semantics without contradiction. The `gatedOff` bullet still correctly holds (gated providers skip probing entirely, including overrides — `availability_for` short-circuits both override tiers when gated). The intentd#701 / monorepo#991 attributions on the pre-existing fields are preserved correctly.

### Gitlink bump ✅

- PR touches exactly two paths: `docs/PROTOCOL.md` and the `packages/intentd` gitlink — nothing else staged.
- `d4d4e436` is the current tip of intentd `origin/main`.
- `git merge-base --is-ancestor 5ec2aad7 d4d4e436` confirms the #717 squash (`5ec2aad7 fix: providerDiscovery installed status honors providers.paths overrides (#717)`) is contained in the bump.

### Non-blocking observation (no change requested)

§5.13 `host.providerAuthStatus` says "Installed-ness comes from the daemon's provider discovery". Its actual install gate (`provider_auth.rs` `resolve_probe_binary`) calls `find_provider_binary(provider_id, command, None)` — i.e. **without** `providers.paths` overrides — so an override-only install will show `installed: true` in §5.14 discovery but still be skipped (`authenticated: null`) by the auth probe. That's pre-existing intentd behavior, not something this docs PR introduces or misstates, but it may be worth a follow-up (either thread overrides into the probe gate in intentd, or add a clarifying sentence to §5.13).

panghy:
Re the non-blocking §5.13 observation: agreed that's pre-existing intentd behavior (auth probe's install gate doesn't consult providers.paths) and out of scope for this docs+bump PR — worth a follow-up issue if we want the probe gate to honor overrides too.

PR #1092, opened 2026-07-29
chore: update intentd submodule for override-aware auth-probe gate (intentd#725)

### LINKED ISSUE #1086: host.providerAuthStatus install gate ignores providers.paths overrides
Follow-up from the intent-hq/monorepo#1077 review (same class of gap as #1065, which fixed `host.providerDiscovery`).

**Problem:** the auth-probe install gate (`resolve_probe_binary` in `crates/intent-services/src/provider_auth.rs`) resolves each probe-able provider's binary via `find_provider_binary(provider_id, command, None)` — the explicit-path tier is never populated from settings. A provider that is only reachable through a valid `providers.paths` override is treated as *not installed* and short-circuits to `authenticated: null` without probing, even though spawn resolution (and, since intentd#717, `host.providerDiscovery`) would honor the override.

**User-visible consequence:** with a valid `providers.paths` override and no PATH-resolvable binary, the FE shows the provider's auth state as unknown/not-installed in `host.providerAuthStatus` (PROTOCOL.md §5.13) while discovery reports it installed — inconsistent surfaces.

**Fix direction:** read `providers.paths` at the services layer (precedent: the doctor fix in intentd#717 loads `SettingsFile`; discovery threads overrides from the transport layer) and pass the validated override into `find_provider_binary` as the explicit-path tier. Override keys should match spawn resolution (`ProviderConfig::primary_binary_provider_id`). Invalid overrides keep the existing fall-through semantics.

Related: #1065 (closed, discovery), intentd#717, intentd#707.

panghy:
Bumps the `packages/intentd` submodule to the merged squash of intent-hq/intentd#725 (`b368d3f`) and updates the `host.providerAuthStatus` documentation to match.

## What changed

- **Submodule bump**: `packages/intentd` now points at `b368d3f` — `fix(services): providerAuthStatus install gate honors providers.paths overrides (#725)`.
- **`docs/PROTOCOL.md`** (`host.providerAuthStatus`, install gate bullet only): the not-installed short-circuit now documents that valid `providers.paths` overrides (absolute path to an executable file, same validation as spawn resolution) count toward the gate for providers whose gate command is the registry primary (`auggie`, `opencode`, `droid`, `grok`); invalid overrides fall through to auto-detection. `claude-code`, `codex`, and `pi` gate on the real `claude` / `codex` / `pi` CLIs (distinct from the adapter binaries their override keys describe), so adapter overrides are ignored for their gates. `auggie`'s gate follows the `host.checkAuggie` precedence (`context.auggiePath` → `providers.paths.auggie`) before falling through to auggie auto-detection.

No wire shape changes — only which providers get probed.

## Verification

- `make check` and `make test` green from the monorepo root (fmt, clippy `-D warnings`, full cargo test workspace).

Refs intent-hq/intentd#725.

Closes #1086


panghy:
Reviewed against the intentd#725 squash (`b368d3f`) and the code at that commit. LGTM — approving.

**Verified:**

- **Gitlink**: `packages/intentd` moves `ca2d91c` → `b368d3fff56275c70895c53b894675f90a20d65e`, which is exactly the squash of intent-hq/intentd#725 (`fix(services): providerAuthStatus install gate honors providers.paths overrides`).
- **Doc accuracy** (checked against `crates/intent-services/src/provider_auth.rs`, `crates/intent-transport/src/host.rs`, and the provider registry at `b368d3f`):
  - Override honored only when the gate command IS the registry primary: `resolve_probe_binary` sets `override_applies = cfg.command == command`; registry primaries are `auggie`/`opencode`/`droid`/`grok` for those ids, so the doc's provider list is correct.
  - `claude-code`/`codex`/`pi` gate on the real `claude`/`codex`/`pi` CLIs while their registry primaries are the `claude-agent-acp`/`codex-acp`/`pi-acp` adapters — adapter overrides ignored, pinned by `special_case_gates_ignore_adapter_overrides`.
  - auggie precedence: the transport inserts `context.auggiePath` over `providers.paths.auggie` (`host.rs`, ProviderAuthStatus arm) and `resolve_probe_binary` falls through to `find_auggie` — matches the doc's `host.checkAuggie` precedence claim.
  - Invalid overrides fall through to auto-detection, pinned by `invalid_override_falls_through_to_auto_detection` and the auggie test.
  - No wire shape change: only the install gate's binary resolution changed; the `{ providers: [{ id, authenticated }] }` response is untouched.
- **Scope**: exactly 2 files — the gitlink and the single `host.providerAuthStatus` bullet in `docs/PROTOCOL.md`. No stray edits.
- **Conventional commits**: PR title and the single branch commit are both `chore: update intentd submodule for override-aware auth-probe gate (intentd#725)` — valid type, safe to squash-merge. `pr-title` check is green.

**Non-blocking nit:** the parenthetical "(an absolute path to an executable file — the same validation as spawn resolution)" is grouped over all four providers including `auggie`, but auggie's override validation is checkAuggie parity (existing file or symlink; need not be absolute or executable). The bullet's final sentence does state this explicitly ("same file/symlink validation"), so the semantics are fully documented — just a slight tension with the grouping sentence. Fine to leave as-is or tighten in a follow-up.

panghy:
Addressed the nit in `abfe8f6d`: the absolute+executable parenthetical now applies only to `opencode` / `droid` / `grok`, and `auggie`'s checkAuggie-parity (file/symlink) validation is stated where its precedence is described, so the grouping no longer implies auggie shares the spawn-resolution validation.

PR #1153, opened 2026-07-30
chore: expand .intent/config.json scripts to cover all Makefile targets

panghy:
Expands the saved scripts in `.intent/config.json` to cover all Makefile targets (adds `help`, `ensure-*-submodule`, `clean-dev`, `sweep`, `sweep-all`, `dev-daemon`, `release-daemon`, `run-intentd`, `run-fe`, `build-sidecar`, `dev`, `ios-open`, `ios-info`) and reorders existing entries. `dev` and `dev-daemon` are marked as service-mode scripts.

---

### DIFF
--- a/.intent/config.json
+++ b/.intent/config.json
@@ -112,20 +112,20 @@
     {
       "name": "release-daemon",
       "command": "make release-daemon",
-      "mode": "command",
-      "category": "other"
+      "mode": "service",
+      "category": "dev"
     },
     {
       "name": "run-intentd",
       "command": "make run-intentd",
-      "mode": "command",
-      "category": "other"
+      "mode": "service",
+      "category": "dev"
     },
     {
       "name": "run-fe",
       "command": "make run-fe",
-      "mode": "command",
-      "category": "other"
+      "mode": "service",
+      "category": "dev"
     },
     {
       "name": "build-sidecar",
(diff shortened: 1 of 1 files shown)


panghy:
**Review findings** (cross-checked every entry against the root `Makefile`):

✅ Coverage is complete: all 25 script entries map 1:1 to the 25 Makefile targets, no duplicates, JSON is valid, and `dev` / `dev-daemon` are correctly `service`-mode.

⚠️ One real issue — three long-running targets are marked `"mode": "command"` but should be `"service"` by the same rationale used for `dev`/`dev-daemon`:

| Script | Why it's long-running |
|---|---|
| `release-daemon` | Runs `intentd serve` — a daemon that blocks until stopped |
| `run-intentd` | Deprecated alias that invokes `release-daemon` (same behavior) |
| `run-fe` | Makefile comment says: "Long-running; does not exit until you stop it (Ctrl-C)" |

Command-mode runs are waited on with a timeout (default ~30s), so launching any of these as a `command` script will always time out with partial output instead of running as a supervised process.

Minor (non-blocking): `release-daemon`, `run-intentd`, and `run-fe` could also use `"category": "dev"` rather than `"other"`, consistent with `dev`/`dev-daemon`.

panghy:
Addressed: `release-daemon`, `run-intentd`, and `run-fe` are now `"mode": "service"` with `"category": "dev"` (commit 4956a77a).

PR #1176, opened 2026-07-30

### LINKED ISSUE #1155: script.run cancelled by the workspace_api eval timeout orphans the script process and permanently corrupts its runtime state
## Summary

When an agent calls `ws.script.run` and the 30s `workspace_api` eval budget (`WORKSPACE_API_TIMEOUT` in `intent-acp/src/mcp_server/dispatch.rs`, override `INTENTD_WORKSPACE_API_TIMEOUT_MS`) fires before the script finishes, the in-flight `script_run` future is dropped mid-await. The spawned process is **not** killed, and the script's cached runtime state is left as `Running` forever.

## Cancellation chain (verified in code)

1. **Eval timeout drops the future.** `intent-js/src/engine.rs::eval` wraps the whole QuickJS eval in an outer `tokio::time::timeout`; host futures (`ws.*` calls) are polled inside it. On expiry the pending host future is dropped.
2. **Nothing detaches the API call.** `make_workspace_host` (intent-acp `mcp_server/dispatch.rs`) awaits `workspace_host_dispatch(...)` inline in a `BoxFuture` — no `tokio::spawn`. Cancellation propagates straight into `script_ops::run`.
3. **Drop lands mid-`run_one`.** `intent-services/src/script_ops.rs::run` does `pty.spawn` → `mark_running` (sets `status: Running`, stores `pty_id`) → `run_one().await` → `pty.kill` (its own timeout branch) / `mark_exited`. A drop during `run_one` means **neither the kill nor `mark_exited` ever executes**.
4. **The child survives.** `intent-pty/src/host.rs::spawn` hands the child to dedicated OS threads (`read_loop`, `exit_watch_loop`); there is no kill-on-drop. The process runs to natural completion — or forever, for a non-terminating command.
5. **State is never reconciled.** `script.list` / `script.status` serve the cached `m.state` verbatim (no live `pty.try_exit` check). Only `run_one`→`mark_exited` or the `start()` supervisor flips a script out of `Running` — both cancelled.

## Consequences

| Symptom | Mechanism |
|---|---|
| Orphaned process | Kill branch in `run()` never reached; PTY threads keep the child alive |
| Permanent "running" status (green dot in FE sidebar), even after the process exits | `mark_exited` never called; `list`/`status` serve cached state |
| `script.start` becomes a silent no-op | `start()` returns early when `status == Running` (script_ops.rs ~line 441) |
| Double-orphan on agent retry | `run()` has **no** running-guard: a second `ws.script.run` spawns a new PTY and `mark_running` **overwrites `m.pty_id`** — the first PTY becomes fully untracked; even `script.stop` can no longer reach it |
| Explicit `timeoutSeconds` > 30 doesn't help | The eval budget fires first and drops the script-level `tokio::time::timeout` along with everything else |

Partial recovery: as long as no second `run()` overwrote `pty_id`, `script.stop` kills the stale `pty_id` and resets state. After a second run, the first process is unreachable except by OS-level kill.

## Suggested fix direction

Make `script_run` cancellation-safe in `intent-services`:

- Detach the run onto a spawned task that always executes kill/`mark_exited` (the RPC can return early with `timedOut` when the caller's budget is exceeded), **or** add a drop-guard that reaps the PTY and marks the script exited on cancellation.
- Independently, add a running-guard to `run()` (mirroring `start()`) to close the `pty_id`-overwrite hole.
- Regression test first: a UDS/WSS e2e (plus a unit test in `script_ops`) that cancels a `script.run` mid-flight and asserts the PTY is reaped and `script.status` reports `exited`.

## Related

- #871 (closed) documents the same root pattern — a mutating `workspace_api` host call cancelled mid-flight by the eval budget with partial application — but not this script-orphan/state-corruption consequence.

Secondary UX gaps observed in the same investigation (can be split out if desired): script `mode` (command vs service) is not visible or editable anywhere in the FE (sidebar add/edit forms omit it; `newMode` is hardcoded to `command`), and the script status vocabulary has no "timed out" state — a daemon-killed run renders as signal-stopped, indistinguishable from a manual stop.

panghy:
Bumps the `packages/intentd` submodule to latest main (`7efa04c`), pulling in intent-hq/intentd#777.

## Summary of the fix

- **Cancellation safety**: `script.run` now survives the client request being cancelled mid-run — the run continues to completion in a detached task instead of being aborted with the RPC, so state/output are not lost.
- **Running guard**: a `script.run` on a script that is already running returns a clear error instead of racing the in-flight run.
- **Concurrent-entry reservation**: the running entry is reserved atomically before spawning, closing the window where two concurrent `script.run` calls for the same script could both start.

Fixes #1155


panghy:
## Verification review (submodule bump)

**Verdict: ✅ APPROVED** — all checks pass.

| # | Check | Result | Evidence |
|---|-------|--------|----------|
| 1 | Diff touches only the `packages/intentd` gitlink | ✅ | `changed_files: 1`; sole file is `packages/intentd`, patch `-Subproject commit 227c5466...` → `+Subproject commit 7efa04ce...` (1 addition / 1 deletion) |
| 2 | New gitlink SHA matches latest intentd main / PR #777 merge | ✅ | `7efa04cea7b96e4457e303479585564dbba38a18` == `intent-hq/intentd` `main` HEAD == `merge_commit_sha` of intent-hq/intentd#777 (merged 2026-07-30T23:06:14Z, squash). Old gitlink `227c5466` was that PR's base — a clean fast-forward of exactly one merge commit. |
| 3 | Conventional-commit compliance | ✅ | PR title `chore: update intentd submodule to latest main` is a valid `chore:` type. The single branch commit `c293270` carries the same conventional message, so a squash merge lands a conventional commit on main (per AGENTS.md single-commit PR guidance). |
| 4 | Body references | ✅ | Body links intent-hq/intentd#777 and states `Fixes #1155`. The summary (cancellation-safe detached run, already-running guard, atomic running reservation) accurately reflects PR #777's content. |

Matches the AGENTS.md Phase 2 monorepo-update workflow. No concerns; safe to merge.

PR #1183, opened 2026-07-31
ci: move route job and utility burst to GitHub-hosted ubuntu-latest

panghy:
Swaps the two WarpBuild references in `.github/workflows/ci.yml` to GitHub-hosted `ubuntu-latest`:

- `route` job: `runs-on: warp-ubuntu-latest-x64-2x` → `ubuntu-latest`
- `utilityBurst` label inside the route script: `["warp-ubuntu-latest-x64-2x"]` → `["ubuntu-latest"]`
- Updated the comments that mentioned WarpBuild accordingly

No `warp-` string remains in `.github/workflows/`.

**This PR doubles as the GitHub Actions budget test**: its own CI run's `route` job must acquire a GitHub-hosted runner and complete. If the org budget is still broken, the job will fail to start or fail with a billing error — the observed outcome will be reported on this PR.

Part of the WarpBuild → GitHub-hosted CI migration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,33 +6,31 @@ permissions:
   pull-requests: read
 
 env:
-  # Burst pr-title to WarpBuild when more than this many Linux jobs are
-  # queued/running ahead of this run (see the route job). Bursted jobs
-  # don't carry the self-hosted+Linux labels and stop counting toward
-  # the depth, so the system naturally falls back to tinybox as it
-  # frees up.
+  # Burst pr-title to a GitHub-hosted runner when more than this many
+  # Linux jobs are queued/running ahead of this run (see the route job).
+  # Bursted jobs don't carry the self-hosted+Linux labels and stop
+  # counting toward the depth, so the system naturally falls back to
+  # tinybox as it frees up.
   LINUX_BURST_THRESHOLD: 2
 
 jobs:
   # Decides where pr-title runs: the local tinybox runner normally, or a
-  # WarpBuild ephemeral runner when the tinybox queue is backed up. Runs
-  # on a small WarpBuild ephemeral runner so the burst decision is
-  # independent of the tinybox queue: a tinybox-pinned probe would queue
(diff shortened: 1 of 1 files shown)


panghy (commented):
Reviewed the diff — the change is correct and minimal. Approving.

**Correctness checks**
- ✅ `route` job `runs-on`: `warp-ubuntu-latest-x64-2x` → `ubuntu-latest`
- ✅ `utilityBurst` label in the route script: `["warp-ubuntu-latest-x64-2x"]` → `["ubuntu-latest"]`
- ✅ No `warp` string remains anywhere in `.github/workflows/ci.yml` at head (`grep -i warp` → no matches)
- ✅ YAML parses cleanly (`yaml.safe_load` → OK)
- ✅ Comment updates are accurate: the burst-routing explanation still holds for GitHub-hosted runners (they don't carry `self-hosted`+`Linux` labels, so they don't inflate the measured depth), and dropping the "third-party WarpBuild runners — an accepted trade-off" caveat from the `actions: read` permission comment is correct since GitHub-hosted runners are first-party
- ✅ No scope creep: single file, only the two runner references and their comments

**Budget test result (this PR's own CI run): ❌ FAILED — billing still broken.**

Both GitHub-hosted jobs on head `5232eb7` failed to start in ~9s with zero steps executed. Annotation on `route` (job [91049126008](https://github.com/intent-hq/monorepo/actions/runs/30596247408/job/91049126008)) and `Analyze` (job [91049122810](https://github.com/intent-hq/monorepo/actions/runs/30596245696/job/91049122810)):

> The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings

So the org's Actions billing/spending limit must be fixed before this can merge — as written, merging now would leave `route` (and therefore `pr-title`) unable to run at all, with no tinybox fallback since the probe itself needs a GitHub-hosted runner. Re-run CI after the billing fix to confirm `route` acquires a runner before merging.

PR #1184, opened 2026-07-31
ci: drop route job, pin pr-title to ubuntu-latest

panghy:
Follow-up to #1183: per the directive that seconds-long jobs skip tinybox routing entirely, the monorepo workflow's only routed job is `pr-title`, so the `route` job is pure overhead.

- Delete the `route` job and the `LINUX_BURST_THRESHOLD` env (and related comments).
- `pr-title`: drop `needs: route`, run directly on `ubuntu-latest`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,81 +5,11 @@ permissions:
   contents: read
   pull-requests: read
 
-env:
-  # Burst pr-title to WarpBuild when more than this many Linux jobs are
-  # queued/running ahead of this run (see the route job). Bursted jobs
-  # don't carry the self-hosted+Linux labels and stop counting toward
-  # the depth, so the system naturally falls back to tinybox as it
-  # frees up.
-  LINUX_BURST_THRESHOLD: 2
-
 jobs:
-  # Decides where pr-title runs: the local tinybox runner normally, or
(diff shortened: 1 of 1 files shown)


panghy (commented):
**LGTM** — reviewed the diff against the stated scope (follow-up to #1183: seconds-long jobs skip tinybox routing entirely).

Verified:
- The `route` job, the `LINUX_BURST_THRESHOLD` env block, and all related comments are fully removed; the resulting `ci.yml` has no dangling references to `route`, `needs.route`, or the threshold.
- `pr-title` is pinned to `runs-on: ubuntu-latest` with no `needs`, and retains `if: github.event_name == 'pull_request'`, so push events still skip it as before.
- Top-level `permissions` (`contents: read`, `pull-requests: read`) remain sufficient for `amannn/action-semantic-pull-request@v6`; the job-level `actions: read` that existed only for the route probe is gone with the job.
- YAML structure of the resulting file is valid.

One observation (not caused by this PR): both checks on this PR failed with *"The job was not started because recent account payments have failed or your spending limit needs to be increased"* — GitHub-hosted runners are currently blocked by an account billing issue. This is pre-existing (the deleted `route` job also ran on `ubuntu-latest`, so this PR does not regress anything), but `pr-title` will not run until billing is resolved. Worth fixing billing before relying on this check, or a re-run after it's resolved.

PR #1187, opened 2026-07-31

### LINKED ISSUE #1124: intentd: cow_clone uses COPYFILE_CLONE (best-effort) instead of COPYFILE_CLONE_FORCE — silent byte-copy fallback contradicts the never-byte-copy contract and weakens cow_probe
## Summary

The macOS CoW implementation (`crates/intent-git/src/cow/macos.rs`) passes `COPYFILE_CLONE` (alone for per-file clones, and `COPYFILE_CLONE | COPYFILE_RECURSIVE` for the whole-tree fast path). Per copyfile(3) semantics, `COPYFILE_CLONE` is **best-effort**: when cloning is not possible, copyfile silently falls back to a physical byte copy and returns success. The flag that enforces clone-or-fail is `COPYFILE_CLONE_FORCE`.

This contradicts the module's own documented contract in `crates/intent-git/src/cow/mod.rs`:

> Never falls back to a byte copy — returns `Unsupported` instead.

## Consequences

1. **Silent full byte copies on non-CoW volumes.** On a filesystem that cannot reflink (e.g. HFS+, some network mounts) where the operation doesn't fail with `EXDEV` first, `cow_clone` performs a slow, disk-expensive physical copy instead of returning `Unsupported` — the caller (e.g. `sandbox_ops::provision_sandbox`) believes it got a cheap reflink clone.
2. **`cow_probe` is weaker than designed.** The live probe writes a tiny temp file and "clones" it with the same best-effort flag; a byte-copy fallback succeeds, so the probe can report `Supported` on filesystems that cannot reflink at all. This matters more going forward: the planned microVM backend gates `microvmSupported` on `cowSupported` (FC-D5 in #1120), so a false-positive probe would enable microVM workspaces on hosts where every per-agent "clone" is a full physical copy.

## Proposed fix

- Use `COPYFILE_CLONE_FORCE` (with `COPYFILE_RECURSIVE` for the tree fast path) so non-clonable cases fail with an errno the existing `ENOTSUP | EOPNOTSUPP | EXDEV → Unsupported` mapping catches, preserving the documented contract.
- Verify the best-effort per-entry walk fallback still behaves correctly (skips genuinely non-clonable entries like sockets/FIFOs rather than byte-copying them).
- Add/adjust unit coverage where feasible (the existing tests already skip on non-CoW CI filesystems).

In practice this never bites on APFS same-volume (virtually all macOS hosts), so this is a correctness/contract fix, not a user-visible bug today. **Out of scope for the microVM v1 effort (#1120)** — independent follow-up.

## References

- `crates/intent-git/src/cow/macos.rs` (`clone_file`, `clone_tree_fast`, `probe`)
- `crates/intent-git/src/cow/mod.rs` module contract
- Related: #1120 (microvmSupported gates on cowSupported), #954, intent-hq/intentd#592

panghy:
Bumps the `packages/intentd` submodule to latest main (`9edbdf0`), picking up intent-hq/intentd#782 — `fix(intent-git): use COPYFILE_CLONE_FORCE for per-file CoW clone`.

Fixes #1124


panghy:
## Verification review — ✅ Approved

Verified as PR reviewer (verifier agent):

1. **Diff scope** ✅ — The diff touches only the `packages/intentd` gitlink: `557155c2ff7f17382d6a53b81d04f69b08d3b61d` → `9edbdf09d611ef8670d94b17bb3d60ea277ee0ee` (1 file, +1/−1).
2. **Target commit** ✅ — `9edbdf0` is the current tip of `intent-hq/intentd` `main` and is the squash-merge of intent-hq/intentd#782 (`fix(intent-git): use COPYFILE_CLONE_FORCE for per-file CoW clone`), including the regression test for the clone-or-fail contract. The bump is a clean fast-forward from `557155c` (4 commits ahead, 0 behind).
3. **Title/body** ✅ — `chore: update intentd submodule to latest main` is a valid conventional commit (`chore` is in the CI-allowed set); the body references intentd#782 and contains `Fixes #1124`.

**Note (non-blocking, infra):** the `route` check on head `acf0395` failed with a GitHub Actions billing error ("recent account payments have failed or your spending limit needs to be increased"), and `pr-title` was skipped as a result. This is unrelated to the change; re-run checks once billing is resolved if branch protection requires them.

(Cannot formally approve via review — authenticated account is the PR author.)

PR #1202, opened 2026-07-31
docs: document live-turn preview overlay in PROTOCOL §5.5

panghy:
Documents the live-turn preview overlay (intent-hq/intentd#786) in `docs/PROTOCOL.md` §5.5:

- **`agent.list`**: adds a **Live-turn preview overlay** block to the `AgentLite` row — while a turn is in flight (`isResponding` with the live-turn slot held by a busy worker, the same gate as the STAB-125 turn-liveness reads), `lastAgentResponse`/`digest` are derived from the live turn's streamed-so-far text instead of the persisted last-assistant-message preview, with a **per-field fallback** so an early turn that has streamed no text/digest yet never blanks the previous preview. Read-path only — nothing new is persisted; idle agents serve the persisted preview exactly as before.
- **`agent.get`**: notes it shares the same overlay (same projection as `agent.list`).

Consistent with the content-free `agent:stream:activity` rename (intent-hq/intentd#775, already documented via #1192) — the overlay is served from the daemon's in-memory stream buffer on the read path, not from event content. Adopted client-side by intent-hq/ios#62.

Docs only; monorepo main already contains the intentd implementation via #1199 (no submodule bump needed).

---


panghy (commented):
Reviewed the §5.5 overlay docs against the merged intentd#786 implementation (`project_lite_with_flags_inner` in `intent-services/src/agent_ops.rs`, `live_turn_text_blocks` in `agent_session.rs`).

**Verified accurate:**
- **Gate** — matches the implementation exactly: the overlay is computed only when `is_responding` (the `agent_is_busy` active-worker signal, with the terminal short-circuit) *and* `live_turn_text_blocks` finds an open slot. An orphan slot with no busy worker never overlays, and `is_responding` is literally the same value threaded into `live_turn_liveness_for`, so "the same gate as the STAB-125 turn-liveness reads below" is correct (including "below" — the STAB-125 block follows in the row).
- **Per-field fallback** — the code only overwrites `last_agent_response` / `digest` when the live extraction yields `Some(..)` for that field; a no-text (or no-digest) turn keeps the persisted value, as documented.
- **Read-path only** — no store/schema writes anywhere in the overlay path; idle agents take the `live_overlay = None` branch and serve the persisted-preview projection byte-identical to before.
- **agent.get cross-reference** — `agent_get_op` goes through the same `project_lite_with_flags_from_projection`, so "same projection as agent.list (including the intentd#786 live-turn preview overlay…)" is correct.
- **Consistency with intentd#775 / STAB-125** — no conflict with the §7 content-free `agent:stream:activity` docs: the overlay is served from the daemon's in-memory live-turn slot on the read path, and nothing content-bearing re-enters the `agent:*` event family. The STAB-125 turn-liveness text in the same rows is preserved verbatim.
- **Formatting** — the bold `**Name (ref, additive):**` block, link style, and em-dash prose match the adjacent STAB-125 and monorepo#940 blocks; plain-text `intentd#786`/`monorepo#940` refs in the agent.get row are consistent with each other.

Two minor wording notes inline on the agent.list row — one small accuracy fix suggested, one non-blocking nit. Optional, non-blocking: the same overlay also rides the `AgentLite` returned by `agent.update` (its success payload is built via `project_lite_with_flags`); not worth blocking on, but could be folded in if the row is ever touched again.

docs/PROTOCOL.md
```diff
@@ -1188,8 +1188,8 @@ The largest namespace. Every `agent.*` method is served daemon-primary by `inten
...
...nothing until it ends, so these let a poller tell a long-but-alive turn (timestamp advancing) from a wedged agent (timestamp pinned) while `lastActivity` stays pinned at the last persisted message. Caveat: the stamp only advances on stream traffic, so during a long silent tool call it pins too — combine with `isWaitingOnTool` to avoid misclassifying a healthy-but-slow tool turn. **Corrupted-session flag ([monorepo#940](https://github.com/intent-hq/monorepo/issues/940), additive):** `sessionCorrupted: true` is present only when the session is parked in `error` (`status == "error"` is required for BOTH causes) AND either (a) the failure classifies as session-fatal (provider safety block, deterministic `session/prompt` 400 `invalidArgument` rejection) or (b) the consecutive-identical-failure streak hit the poisoned threshold — the structured signal that `agent.retry` will recreate the provider session (fresh `session/new`) instead of resuming, or that spawning a fresh agent is the right recovery. **Derived on emit** over the persisted (status, stop_reason) + the in-memory failure streak — never persisted as a column — and **omitted when `false`** (absent ≠ present-false on the wire) |
```
panghy: **Accuracy nit** — "(the same extraction used at persistence time)": `last_response_and_digest_from_blocks` never runs at persistence time. All three call sites are serve-time reads (`project_lite` over full transcripts, the monorepo#958 projection path, and this overlay); no preview is persisted as such — the "persisted preview" is itself derived on read from the newest persisted assistant row. As written the parenthetical also sits oddly next to this row's own "Read-path only — nothing new is persisted". Suggest: "(the same extraction that derives the persisted-preview fields from the newest assistant row)".

panghy: **Nit, non-blocking** — in the adjacent blocks "additive" marks new wire fields (STAB-125's `turnInFlight`/`lastStreamActivityAt`, the derived `sessionCorrupted?`). This block adds no fields — it changes the mid-turn *values* of two existing fields. If "additive" here is shorthand for "backward-compatible, no shape change" it's defensible; "read-path" (or dropping the tag) would be slightly more precise.

panghy: Adopted verbatim in 03b37557 — the parenthetical now reads "(the same extraction that derives the persisted-preview fields from the newest assistant row)".

panghy: Agreed — swapped the tag to "read-path" in 03b37557 since this block changes mid-turn values of existing fields rather than adding wire fields.

PR #1207, opened 2026-07-31
docs: document push-preview fields on agent:stream:activity/end; bump ios + cloudlands-fe submodules

panghy:
Documents the push-preview enrichment of the agent streaming events ([intent-hq/intentd#792](https://github.com/intent-hq/intentd/pull/792)) in `docs/PROTOCOL.md` §7, and bumps the client submodules that adopted it.

## Docs (§7 + event summary table)

- **`agent:stream:activity`**: payload is now `{ agentId, messageId, lastAgentResponse?, digest? }` — the liveness tick enriched with server-derived live-preview fields so clients (workspace cards, watched-agent footers, iOS previews) render a live preview push-style without an `agent.get` refetch on every tick. `lastAgentResponse` is the trailing **≤500 chars** of the turn's streamed-so-far text blocks (`...`-prefixed when truncated); `digest` is the parsed digest span — same extraction as the §5.5 `AgentLite` live-turn overlay (intentd#786). Each field is **omitted until derivable** (never an empty string). Throttling (leading-edge, 1 s) and transience are unchanged.
- **`agent:stream:end`**: all transcript-bearing terminal emits (turn-worker, interrupt flush, harness-wake finalize) stamp final `lastAgentResponse?` / `digest?` re-derived from the turn's full streamed text; the not-surfaced-by-streaming failure path carries no preview fields.
- Removed the stale "content-free" framing and the legacy `agent:stream:chunk` version-skew note — no current client handles the old name ([intent-hq/cloudlands-fe#579](https://github.com/intent-hq/cloudlands-fe/pull/579) dropped the FE compat path; [intent-hq/ios#66](https://github.com/intent-hq/ios/pull/66) dropped the iOS activity-ping refetch fallback).

## Submodule bumps

- `packages/ios` → `9ec15d4` ([intent-hq/ios#65](https://github.com/intent-hq/ios/pull/65) + [intent-hq/ios#66](https://github.com/intent-hq/ios/pull/66) — apply the preview payload directly, drop the refetch fallback)
- `packages/cloudlands-fe` → `a4cacc09` ([intent-hq/cloudlands-fe#579](https://github.com/intent-hq/cloudlands-fe/pull/579) — live watched-agent footer previews from the activity payload)
- `packages/intentd` — no bump needed; monorepo main already contains `efe0547` (intentd#792).


panghy:
## Review

Verified the documented contract against the intentd#792 implementation (merged as `efe0547`, confirmed an ancestor of the currently-pinned `packages/intentd` commit `dd0bd04`) and the submodule ref updates. Everything checks out.

### `docs/PROTOCOL.md` §7 / event table vs. implementation

- `PREVIEW_RESPONSE_CAP = 500` in `agent_session.rs` matches the documented "≤500 trailing chars" cap, and the `...`-prefix-when-truncated behavior matches `stamp_preview_fields`'s tail-slice + `format!("...{tail}")` logic.
- "Omitted until derivable (never an empty string)" matches: `stamp_preview_fields` only sets `data["lastAgentResponse"]` / `data["digest"]` inside the `if let Some(...)` arms — no field is ever set to `""`.
- `agent:stream:activity` payload composition (`agentId`, `messageId`, then `stamp_preview_fields(&mut activity_data, &transcript.text_block_strings())`) matches the doc's claim that it's still content-free (no raw `content`) but now carries the derived preview.
- All three transcript-bearing terminal `agent:stream:end` emit sites stamp final preview fields re-derived from the *full* turn text, matching the doc exactly:
  - `run_prompt_turn` normal/error completion — `stamp_preview_fields(&mut end_data, &preview_text_blocks)` where `preview_text_blocks` comes from the full persisted/attempted blocks.
  - `run_harness_wake_turn` finalize — same helper, full wake-turn blocks, payload correctly shaped `{ agentId, messageId?, lastAgentResponse?, digest? }`.
  - **Interrupt flush** (`agent_manager.rs` `interrupt_inner`) — stamps preview fields from `interrupted_text_blocks` (the flushed partial-turn snapshot) alongside `stopReason: "interrupted"` and conditional `messageId`, with no `trailingBlocks`/`turnId` — matches the doc's interrupt-path carve-out precisely.
- The **not-surfaced-by-streaming failure path** (`publish_terminal_failure_events`) does *not* call `stamp_preview_fields` — `end_data` only ever gets `agentId` (+ `turnId?`) — matching the doc's "carries no preview fields" claim for that arm.
- Unit tests in `agent_session/tests.rs` assert exactly the behaviors documented: first-activity preview derived from the first chunk with `digest` omitted absent an `<agent_digest>` span, and the terminal `stream:end` carrying the final preview from the full turn text (not just the last throttled tick).

The removed "content-free" framing and the legacy `agent:stream:chunk` version-skew note are both accurately superseded — no current client (cloudlands-fe#579 dropped the FE compat path, ios#66 dropped the refetch fallback) handles the old event name, so the removal is factually correct, not just aspirational.

### Submodule bumps

- `packages/ios` → `9ec15d4`: `main` HEAD is exactly `9ec15d4`, and `compare 15ac4183...9ec15d4` reports `status: ahead` (clean fast-forward, 2 commits: ios#65 then ios#66).
- `packages/cloudlands-fe` → `a4cacc09`: `main` HEAD is exactly `a4cacc09`, and `compare a80626c...a4cacc09` reports `status: ahead` (clean fast-forward, 1 commit: cloudlands-fe#579).
- Both referenced PRs (ios#65/#66, cloudlands-fe#579) are merged, and their descriptions independently corroborate the same contract (500-char cap, omit-until-derivable, push-apply with no `agent.get` refetch), cross-checking the daemon-side description in this PR.
- `packages/intentd` is correctly left unbumped — `efe0547` (intentd#792) is confirmed an ancestor of the currently-pinned `dd0bd04`.

**No issues found.** Docs, code, and submodule pins are all consistent. LGTM — (note: couldn't submit a formal "Approve" review since GitHub disallows approving one's own PR; posting as a comment instead).

panghy:
Addressed in 1c4d50ad — reworded both spots to say the rename dropped the per-chunk transcript delta (the incremental content firehose, now on the internal chat:stream:delta), and that lastAgentResponse is a capped server-derived preview rather than the raw delta stream.

Issue #588, opened 2026-07-23
EXECUTE_COMMAND_STREAMING schema carries cwd but no workspaceId — cwd-bearing streaming payloads always fail daemon containment guard


panghy:
Follow-up spotted during review of intent-hq/cloudlands-fe#321 (monorepo#578).

## Problem

`SystemExecuteCommandStreamingSchema` (`src/main/ipc-schemas.ts`) still carries an optional `cwd` but has **no `workspaceId` field at all**:

```ts
export const SystemExecuteCommandStreamingSchema = z.object({
  sessionId: z.string().min(1, 'Session ID is required'),
  command: z.string().min(1, 'Command is required'),
  cwd: z.string().optional(),
  stdin: z.string().optional(),
});
```

The `EXECUTE_COMMAND_STREAMING` handler (`src/features/system/main/system.ipc.ts`) forwards `cwd` to `hostExecStream` → `host.execStream`, which per PROTOCOL.md §5.14 reuses the same within-workspace containment guard on `cwd` as `host.exec`. So any cwd-bearing streaming payload is guaranteed to fail at the daemon with `-32602 "cwd requires workspaceId"` — the schema cannot express a valid cwd-bearing request.

monorepo#578 named only `EXECUTE_COMMAND`, so #321 deliberately left the streaming schema untouched.

## Fix options

1. Add `workspaceId` to `SystemExecuteCommandStreamingSchema` + the same cwd⇒workspaceId refinement `SystemExecuteCommandSchema` now has, and forward it through the handler to `hostExecStream`; or
2. Drop `cwd` from the streaming schema entirely if no caller uses it (audit `EXECUTE_COMMAND_STREAMING` call sites first).

## References

- intent-hq/cloudlands-fe#321 (review comment noting the gap)
- monorepo#578, monorepo#537 (the `host.exec` counterpart)
- PROTOCOL.md §5.14 (containment guard applies to both `host.exec` and `host.execStream`)

PR #1230, opened 2026-08-01
feat: surface lastMessageRole for conversation previews

panghy:
Surfaces the additive `lastMessageRole` AgentLite field end-to-end: documents the wire contract in `docs/PROTOCOL.md` §5.5 and bumps all three submodules to the merged main commits that implement/consume it.

## Changes

- **docs/PROTOCOL.md §5.5** — documents `lastMessageRole?: "user" | "assistant"` on the AgentLite projection (`agent.list` / `agent.get`): role of the session's newest user/assistant transcript message, system rows transparent, omitted when the transcript has neither (never `null`), denormalized at message-write time so both projection paths agree, with the live-turn read-path overlay flipping it to `"assistant"` on the same per-field gate as the live `lastAgentResponse` overlay.
- **packages/intentd** → `c9af83d11468d8ff0b8000c8e95b085358875ce7`
- **packages/cloudlands-fe** → `18bb73c4434e8301fa24eccfa27d6cab01b4e9f2`
- **packages/ios** → `1a627ac01a3b01684c1d54f4188ed04d3e7b941a`

## References

- intent-hq/intentd#807 — daemon: derive/persist `lastMessageRole` + live-turn overlay
- intent-hq/cloudlands-fe#586 — FE: consume `lastMessageRole` for conversation previews
- intent-hq/cloudlands-fe#588 — FE follow-up: session upsert no-op guard swallowed `lastMessageRole`/`lastUserMessage`-only updates
- intent-hq/ios#67 — iOS: consume `lastMessageRole`

## Verification

`make check` (fmt, clippy `-D warnings`, build) and `make test` (full workspace suite) pass at the monorepo root with all three bumps in place.


panghy:
## Review — verified, approving ✅

Reviewed the §5.5 doc change against the intentd implementation at `c9af83d1` and verified all three submodule bumps. Everything checks out.

### docs/PROTOCOL.md §5.5 vs. intentd implementation

Every documented semantic maps to concrete code at `c9af83d1`:

| Documented claim | Evidence |
| --- | --- |
| Role of the session's newest **user/assistant** transcript message; system (and any other) rows transparent | `intent-services/src/agent_ops.rs` `last_message_role()` (rev-scan filtered to `user`/`assistant`); store-side `batch_preview_col_values()` and the single-append path in `intent-store/src/agent_repo.rs` only stamp the column on user/assistant writes; 0070 backfill filters `role IN ('user','assistant')` — `last_message_role_backfill_and_self_heal` asserts a system tail is transparent |
| **Omitted** when absent, never `null` | `AgentLite.last_message_role: Option<String>` with `#[serde(skip_serializing_if = "Option::is_none")]` (`intent-core/src/model.rs`); `wss_integration.rs` asserts `.get("lastMessageRole").is_none()` on a fresh agent |
| **Denormalized at message-write time** so both projection paths agree | Migration `0070_agent_session_last_message_role.sql` adds the column + backfill; maintained in the same transaction on single append, batch insert, and `replace_agent_messages`; both `get_agent_session_message_projection(s)` read it, with a guarded NULL-column self-heal (`heal_last_message_role`) for pre-0070 rows |
| Live-turn overlay flips to `"assistant"` on the **same per-field gate** as the live `lastAgentResponse` overlay | `agent_ops.rs`: `live_overlay` is gated on `is_responding` + the live-turn slot, and the flip sits inside `if live_response.is_some()` — exactly the branch that overlays `lastAgentResponse`; pre-first-completed-line the persisted value is served unchanged |

Coverage: unit tests in `agent_ops/tests.rs` + store tests (`last_message_role_maintained_on_writes`, backfill/self-heal), and wire-level assertions in the WSS e2e (`wss_integration.rs`) including the omitted-when-absent case — consistent with the intentd e2e requirement.

### Submodule bumps

All three SHAs verified on their repo's `main`:

- `packages/intentd` → `c9af83d1` — `feat: persist and serve lastMessageRole on AgentLite (#807)`, ancestor of `main` ✅
- `packages/cloudlands-fe` → `18bb73c4` — `fix: include preview fields in session no-op guard snapshot (#588)`, ancestor of `main`; the bump range also contains #586 (consume `lastMessageRole`), matching the PR body ✅
- `packages/ios` → `1a627ac0` — `feat: prefer the user's first line in previews when the user message is newest (#67)`, currently `main` HEAD ✅

### Notes (non-blocking)

- The doc's "denormalized … so both projection paths agree" is accurate in effect: the full-transcript path (`project_lite`) re-derives the role from loaded messages while the transcript-free path serves the persisted column, and the derivation and column maintenance implement identical semantics (store tests confirm both paths agree).
- No CI checks had reported on the head commit at review time; the diff is docs + three gitlink bumps, and the PR body reports `make check` / `make test` green at the root with the bumps in place.

**Verdict: APPROVED** — docs match the implementation, all gitlinks point at merged main commits.

PR #1458, opened 2026-08-04
chore: update submodules and document archive/delete active-work teardown

panghy:
## Summary

Final monorepo PR for the "stop agents on workspace archive" feature: submodule bumps plus the protocol docs update.

### Submodule bumps

- **packages/intentd** → `4da40a6` — includes [intent-hq/intentd#896](https://github.com/intent-hq/intentd/pull/896): interrupt agents and cancel hooks on workspace archive; eager hook-task abort on delete.
- **packages/cloudlands-fe** → `e78e3973` — includes [intent-hq/cloudlands-fe#727](https://github.com/intent-hq/cloudlands-fe/pull/727): confirm archive/delete when agents or background hooks are active.

### Docs (`docs/PROTOCOL.md`)

- New **archive active-work teardown** block under §5.1: `workspace.archive` gracefully interrupts in-flight agent turns (provider child, ACP session, session rows, transcripts, completion watches, and pending queues all survive for unarchive; terminal `agent:stream:end` with `stopReason: "interrupted"`), parks queued messages/wakes behind archived gates (re-kicked by `workspace.unarchive`), and cancels ACTIVE background hooks (owner woken with an archive notice; unarchive does not resurrect them).
- **Delete cascade** extended with the eager hook-task abort (task-abort only, no event/wake, before the store cascade drops the rows).
- Cross-references from the §5.1 `workspace.archive`/`workspace.unarchive` table rows and §5.40, plus a within-4.2 behavior-only note in the version paragraph (no wire-shape change).
- `docs/ARCHITECTURE.md` untouched: its only delete mention is the checkout-mode git-metadata phase; it does not describe archive or agent/hook teardown semantics.

---


panghy (commented):
Reviewed the docs against the intentd implementation at `4da40a6` (intent-hq/intentd#896) and verified both submodule refs.

**Verified accurate** ✅

- **Submodule refs**: `packages/intentd` → `4da40a6` is the squash-merge of intent-hq/intentd#896 on intentd `main`; `packages/cloudlands-fe` → `e78e3973` is the squash-merge of intent-hq/cloudlands-fe#727 on cloudlands-fe `main`. Both are the current HEADs.
- **Archive sweep order**: `archive_workspace` persists the archived row first, then interrupts busy agents via `AgentManager::interrupt` (the `agent.stop` keep-alive path — turn cancelled, worker aborted, `agent:stream:end` with `stopReason: "interrupted"`, no `agent:deleted`), then runs the hook cancel sweep. Read-only wiring (no manager) archives without error. Matches the docs.
- **Parked queues**: the archived gate exists in both `AgentManager::try_drain_queue` and `Services::deliver_wake_message`, including the chief skip and the fail-open lookup-error arm; `unarchive_workspace` re-kicks the drain per agent. Matches the docs.
- **Hook cancel sweep**: `cancel_workspace_hooks` routes each ACTIVE hook through the shared `cancel_active_hook` machinery (task abort, `cancelled` persisted, `hook:cancelled` emitted, owner wake with archive notice riding the archived gate), skips terminal hooks, and is best-effort per hook. `abort_workspace_hook_tasks` on delete is task-abort only — no event, no wake, runs before the store cascade. Matches the docs.
- **Residual race paragraph** matches the implementation comment in `archive_workspace` (accepted one-stray-turn window, no post-claim re-check on the drain path).
- **Style**: bold-lede blocks, §-cross-references, version-paragraph placement, and the "behavior only, no wire change" convention all match existing PROTOCOL.md precedent.

**Issues found** — 1 accuracy issue + 1 wording nit, both inline on §5.1. The completion-watch-wake claim is the substantive one: that delivery arm is not actually gated on the archived flag in the code.

docs/PROTOCOL.md
```diff
@@ -687,6 +687,46 @@ asynchronously (fire-and-forget) but the create call is not idempotent unless a
...
+  attached) still archives without error.
+- **Queued messages and wakes park while archived.** Both delivery arms gate on the
+  archived flag: the automatic queue drain and wake delivery (hook wakes,
+  completion-watch wakes, `agent.wakeOrCreate` context messages) refuse to start a turn
+  in an archived workspace — entries stay parked in the pending queue (still visible
+  via `agent.getQueue`, §5.5). The virtual chief workspace skips the row read (never
+  archived), and a row-lookup error fails open so a transient store error cannot strand
+  a queue. `workspace.unarchive` re-kicks the drain for every workspace agent with
+  ready-to-send work (the drain re-checks its own gates), so parked queues deliver
+  after unarchive without a manual kick.
```
panghy: **Accuracy: completion-watch wakes are not gated on the archived flag.** The two gates that exist in the code are `AgentManager::try_drain_queue` (the automatic queue drain) and `Services::deliver_wake_message` (hook wakes + `agent.wakeOrCreate` context messages). Completion-watch wakes take a different arm: `deliver_completion_to_watches` → `deliver_parent_wake` → `AgentManager::send_message`, and `send_message` has no archived check — an idle parent in an archived workspace whose watched child completes elsewhere will claim the slot via `try_begin` and run a real turn. (The archive sweep interrupts the parent, leaving it idle, so the enqueue-when-busy path — which *would* park behind the drain gate — does not apply.)

Note the doc comment on `deliver_wake_message` in `agent_ops.rs` doesn't claim completion-watch coverage either; the gate comment there names only the drain mirror.

Suggest either (a) dropping "completion-watch wakes" from the parenthetical so the doc matches the shipped gates, or (b) keeping the claim and filing a follow-up intentd issue to gate `deliver_parent_wake` (or `send_message`) on the archived flag — in which case the doc should land with the fix, not before it.

docs/PROTOCOL.md
```diff
@@ -687,6 +687,46 @@ asynchronously (fire-and-forget) but the create call is not idempotent unless a
...
+- **Active background hooks are cancelled.** Every ACTIVE (`scheduled`/`running`) hook
+  in the workspace (§5.40) goes through the `hook.cancel` machinery: scheduler task
+  aborted, state persisted to `cancelled`, `hook:cancelled` emitted (§6.5), and the
+  owner woken with an archive-specific notice — the wake itself parks behind the
+  archived gate above, so it queues at most and never starts a turn while archived.
+  Terminal hooks are untouched, and **unarchive does not resurrect cancelled hooks** —
+  the notice tells the owner to reschedule if the condition still matters. Best-effort
```
panghy: **Nit: the notice does not actually tell the owner to reschedule.** The shipped wake text is `This hook was cancelled because its workspace was archived.` (§`hook_manager.rs::cancel_workspace_hooks`) — it explains why the hook stopped but contains no reschedule instruction. Suggest softening to something like “the owner is expected to reschedule if the condition still matters” (framing it as the contract, not the notice text), or quoting the actual notice.

docs/PROTOCOL.md
panghy: Fixed in b62e618d — went with option (a): the parked-queues bullet now names only the two shipped gates (queue drain + `deliver_wake_message`: hook wakes and `agent.wakeOrCreate` context messages) and explicitly documents that completion-watch wakes take the ungated `send_message` arm, so a watched child completing elsewhere can still wake an idle parent in an archived workspace.

docs/PROTOCOL.md
panghy: Fixed in b62e618d — the bullet now quotes the shipped notice text verbatim and frames rescheduling as the contract ("the owner is expected to reschedule if the condition still matters") rather than attributing it to the notice.

PR #1558, opened 2026-08-06
chore: update intentd submodule and pr.snapshot protocol docs

panghy:
## Summary

Phase-2 monorepo follow-up to intent-hq/intentd#942.

- Bumps the `packages/intentd` submodule to `d98e4016` (latest `main`), which fixes `ws.pr.snapshot`'s `reviews.decision` to derive from the provider's authoritative GraphQL `reviewDecision` instead of inferring from REST `mergeable_state == "blocked"`.
- Updates `docs/PROTOCOL.md`'s `ws.pr.snapshot` block (§5.7) to document:
  - The new `reviews.decision` derivation (provider `reviewDecision` when available, falling back to the aggregated actionable reviews — never inferred from `mergeable_state`).
  - The new `comments.threadResolutionUnknown` result field, which flags when thread resolution fell back to REST comments (making `unresolvedThreadCount` an unreliable overcount).

Refs intent-hq/intentd#942
Refs intent-hq/monorepo#1524

---

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -1961,16 +1961,36 @@ The namespace holds the **two** workspace/active-PR-scoped methods that survived
 > updatedAt, mergeable, mergeableState, mergeBlockedReason, checks: { total, passed,
 > failed, pending, failedNames }, reviews: { decision, approvals, changesRequested },
 > comments: { conversationCount, reviewCommentCount, unresolvedThreadCount,
-> totalCount } }`. `mergeBlockedReason` is a human-readable reason and is non-`null`
-> exactly when the PR is open (draft included) and cannot be merged. `checks` tallies
-> the runs on the PR head (SHA, else source branch); a provider without the
-> `checkRuns` capability — or a PR whose head cannot be determined — reports an empty
-> tally rather than failing the snapshot. `comments.reviewCommentCount` counts every
-> inline thread comment **including replies** (threads come from GraphQL when
-> available, else the REST list grouped by reply parent — resolution state is
-> unavailable there, so every fallback thread counts as unresolved), and `totalCount =
-> conversationCount + reviewCommentCount`, so a new reply anywhere moves the counter a
-> hook can diff.
+> threadResolutionUnknown, totalCount } }`. `mergeBlockedReason` is a human-readable
+> reason and is non-`null` exactly when the PR is open (draft included) and cannot be
+> merged. `checks` tallies the runs on the PR head (SHA, else source branch); a
+> provider without the `checkRuns` capability — or a PR whose head cannot be
+> determined — reports an empty tally rather than failing the snapshot.
+> **`reviews.decision`** *(changed in intentd,
+> [intentd#942](https://github.com/intent-hq/intentd/pull/942);
+> [intent-hq/monorepo#1524](https://github.com/intent-hq/monorepo/issues/1524))*
+> is derived from the forge's authoritative review-requirement verdict (GitHub's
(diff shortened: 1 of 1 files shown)


docs/PROTOCOL.md
```diff
@@ -1961,16 +1961,36 @@ The namespace holds the **two** workspace/active-PR-scoped methods that survived
...
+> **`reviews.decision`** *(changed in intentd,
+> [intentd#942](https://github.com/intent-hq/intentd/pull/942);
+> [intent-hq/monorepo#1524](https://github.com/intent-hq/monorepo/issues/1524))*
+> is derived from the forge's authoritative review-requirement verdict (GitHub's
+> GraphQL `reviewDecision`) when available: `approved`, `changes_requested`, or
+> `review_required` (the last only for an open PR, draft included) map directly.
+> When the provider signal is unavailable — no review requirement configured on
+> the base branch, a host without the capability, or a failed fetch — the
+> decision falls back to the aggregated actionable reviews: `changes_requested`,
+> else `approved`, else `none`; the fallback path never yields
```
panghy, line 1980: Non-blocking nit: this enumeration of fallback triggers omits a fourth path in the implementation. When the provider *does* return `review_required` but the PR is merged/closed, `snapshot_review_decision`'s `_` arm also routes to the aggregate fallback — it doesn't yield `none` by fiat. E.g. the unit test at `pr_ops.rs` (closed PR, 1 approval, provider says `ReviewRequired`) asserts `approved`. As written, a reader could infer that a non-open PR with `review_required` reports `none`. Consider appending something like: "a provider `review_required` on a merged/closed PR is likewise discarded and falls back to the aggregate." Fine to defer to a follow-up since the observable behavior for merged/closed PRs is an edge case.

panghy (commented):
Reviewed the doc text against the intentd implementation at the bumped commit `d98e4016` (`crates/intent-services/src/pr_ops.rs::snapshot_review_decision` and the `pr.snapshot` handler in `crates/intent-services/src/lib.rs`). The documentation is accurate on every substantive claim:

- **Direct mapping** — `approved` / `changes_requested` map unconditionally; `review_required` is gated on `state == "open" || "draft"`. Matches the doc's "the last only for an open PR, draft included".
- **Fallback triggers** — null `reviewDecision` (unprotected base branch), hosts without the signal (the trait's default `review_decision` returns `Ok(None)`), and failed fetches (degraded via `.ok().flatten()` with a debug log) all yield `None` → aggregate fallback. Matches the doc's three-item enumeration.
- **Fallback ordering** — `changes_requested`, else `approved`, else `none`; the fallback arm cannot produce `review_required`. Matches.
- **`mergeable_state` no longer consulted for this field** — correct; it is still used for `mergeBlockedReason`, and the doc correctly scopes the claim to `reviews.decision`.
- **`threadResolutionUnknown`** — set `true` exactly when `get_review_threads` fails and the handler falls back to `fetch_all_pages(list_review_comments)` + `fallback_threads` (which hardcodes `is_resolved: false`, so every fallback thread counts as unresolved). `false` means counts come from GraphQL threads. Matches the doc precisely, including the "overcount" characterization.
- **Result shape** — the field order in the signature line (`threadResolutionUnknown` between `unresolvedThreadCount` and `totalCount`) matches the handler's JSON output.

One minor, non-blocking gap noted inline. LGTM otherwise.

docs/PROTOCOL.md
panghy: Addressed in 79df59b8 — added the merged/closed review_required fallback case to the enumeration.

Issue #1582, opened 2026-08-06
intentd: CoW checkouts inherit the source repo's origin verbatim (relative/self-referencing local paths break fetch)


panghy:
## Problem

A CoW checkout (`provision_cow_checkout`, used by both `workspace.create` with `cowIsolation` on and `workspace.duplicate` of a standalone source) reflink-copies the whole repository directory, `.git/config` included. The clone therefore inherits the source repository's `origin` URL **verbatim**, at a new location.

That is correct for network URLs, but wrong for local-path origins:

1. A **relative** local path (e.g. `origin = ../upstream`) resolves against the repository that holds it. In the clone — which lives under the workspaces root, not beside the source — it re-resolves to a different (usually nonexistent) directory, so `fetch`/`push` fail or silently target the wrong repo.
2. A local path naming the **source checkout itself** leaves the clone's `origin` pinned to a directory that can be deleted independently (the metadata half of intent-hq/monorepo#1560).

## Context

[intentd#956](https://github.com/intent-hq/intentd/pull/956) fixed exactly this for the *plain local clone* path (`provision_local_clone_checkout` → `resolve_source_origin`): a relative local origin is absolutized against the source repository, one resolving to the source checkout is dropped, network URLs and already-absolute local paths carry over verbatim. The CoW path has no equivalent step — it was noted as pre-existing and deliberately left out of that PR's scope.

## Suggested fix

Reuse the same resolution in the CoW path: after the reflink clone, run the clone's inherited `origin` through the `resolve_source_origin` logic in `crates/intent-git/src/cow_checkout.rs` (absolutize relative, drop self-referencing, leave network/absolute alone) instead of leaving `.git/config` as copied.

Regression tests belong next to the existing ones in `cow_checkout::tests`, mirroring `local_clone_absolutizes_a_relative_local_origin` and `local_clone_removes_origin_pointing_at_the_source_itself` for the CoW entry point (guarded by the existing `cow_available` skip helper).

## Impact

Low frequency — most repos have a network `origin` — but when it hits, the affected workspace's fetch/push is broken in a way that is hard to attribute, and it undercuts the self-contained-checkout guarantee for CoW workspaces the same way it did for local clones.

Filed while reviewing intentd#956 (review nit 2).

panghy:
Fixed in intentd v0.6.1 (beta).

panghy:
intentd v0.6.2 promoted to stable.

PR #1592, opened 2026-08-07
docs: document standalone-source duplication; bump intentd

### LINKED ISSUE #1560: workspace.duplicate of a direct-mode workspace provisions a fragile linked worktree rooted in the original's checkout
Found during review of intent-hq/intentd#944 (cache-hydrated `workspace.create` + `checkoutMode: "direct"`).

## Problem

The `workspace.duplicate` decision matrix only yields `cow` or `worktree` checkouts. On exactly the filesystems where `direct` arises (CoW unsupported), duplicating a hydrated `direct` workspace falls back to a **linked worktree whose source repo is the original workspace's own standalone checkout**.

Deleting the original workspace then renames/removes that checkout, orphaning the duplicate's worktree (its `.git` gitfile points into a directory that no longer exists).

## Expected

A `direct`-mode source should duplicate as another standalone checkout — e.g. a plain local `git clone` of the source checkout (mirroring `provision_direct_checkout`) — so the duplicate never holds a dangling reference to a sibling workspace's directory.

## Notes

- On CoW-capable filesystems the duplicate is a standalone CoW clone, so the issue only bites where CoW is unsupported (ext4 without reflink, some network filesystems).
- `isNewRepo` `direct` workspaces have the same shape: the "repo" is the user's own folder, so a worktree-based duplicate rooted there is fragile against the user moving/deleting it, though that folder is not daemon-managed.

panghy:
Phase 2 for #1560. Bumps `packages/intentd` to `fd033b2` and lands the matching
protocol/architecture docs.

Submodule PR: intent-hq/intentd#956 (merged).

## What the daemon change does

Duplicating a workspace whose source is a **standalone** checkout (`cow` or `direct`)
now provisions a standalone checkout instead of a linked worktree, so the duplicate
survives deletion of the source. Previously the duplicate got a linked worktree rooted
in the source workspace's checkout: deleting the source detached that directory and
orphaned the duplicate, and deleting the duplicate mutated the source.

- Standalone source → always standalone duplicate, cloned from the source's **own
  checkout** (not its `repositoryPath`). CoW probe decides `cow` vs. a plain local
  clone persisted as `direct` (committed state only). `workspace.cowIsolation` is
  ignored, matching cache-hydrated create.
- `origin` is **resolved**, not copied verbatim: network URL and absolute local path
  carry over; a relative local path is absolutized against the source repository; it is
  removed when the source has none or when it resolves to the source checkout itself.
- On provisioning failure the inherited `repositoryPath` is cleared for standalone
  sources, so a checkout-less duplicate row never points at the source's directory.
- Shared-checkout (worktree-mode) sources keep the existing `workspace.create`
  decision matrix.

## Docs

- `docs/PROTOCOL.md` §5.1 — duplication rules split by source kind (standalone vs.
  shared-checkout), the origin-resolution rules, the provisioning-failure
  `repositoryPath` clearing, and the standalone exemption noted on the
  `workspace.cowIsolation` settings entry.
- `docs/ARCHITECTURE.md` — `workspace.duplicate` split out of the create bullet into
  its own duplication bullet, including the failure-cleanup behavior.

## Follow-up

#1582 tracks a pre-existing CoW-clone limitation surfaced while reviewing this change
(not a regression from it).

Fixes #1560

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -568,17 +568,19 @@ standalone:
   is the source's **own checkout** (`worktreePath`, or `repositoryPath` when the
   repository itself is the checkout), not the source's `repositoryPath`. A CoW probe from
   that checkout to `<root>/<newId>` decides the mode: supported ⇒ CoW clone
-  (`checkoutMode: "cow"`); Unsupported, probe error, or a CoW clone that still reports
-  Unsupported ⇒ a **plain local `git clone`** of the source checkout
-  (`checkoutMode: "direct"`) with a logged warning. A source checkout whose `.git` is a
-  gitfile skips the probe and goes straight to the local clone. `workspace.cowIsolation`
+  (`checkoutMode: "cow"`); Unsupported ⇒ a **plain local `git clone`** of the source
+  checkout (`checkoutMode: "direct"`), and the same local-clone fallback is taken — with a
+  logged warning — on a probe error or when a CoW clone still reports Unsupported despite a
+  passing probe. A source checkout whose `.git` is a gitfile skips the probe and goes
+  straight to the local clone (also warned). `workspace.cowIsolation`
   is **ignored** for standalone sources (parity with cache-hydrated create). Like a
   hydrated create, the checkout *is* the repository: the duplicate persists
-  `repositoryPath` = its **own** checkout path, and `origin` is retargeted at the source's
-  own resolved `origin` — a network URL and an absolute local path carry over verbatim, a
-  relative local path is absolutized against the source repository so it still names the
-  same upstream, and `origin` is **removed** when the source has none or when it resolves
-  to the source checkout itself — so nothing references the source workspace's directory.
+  `repositoryPath` = its **own** checkout path, so nothing in the row references the source
+  workspace's directory. The `direct` local-clone path additionally **resolves `origin`**
+  (a CoW clone is a byte copy and keeps the source's `.git/config` verbatim): a network URL
+  and an absolute local path carry over as-is, a relative local path is absolutized against
+  the source repository so it still names the same upstream, and the remote is **removed**
+  when the source has no `origin` or when `origin` resolves to the source checkout itself.
   The local-clone fallback carries **committed state only** —
   uncommitted/untracked work in the source is not copied. Rationale
   ([intent-hq/monorepo#1560](https://github.com/intent-hq/monorepo/issues/1560)): a linked
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -559,15 +559,45 @@ removed in the background as usual.
 **Duplication (`workspace.duplicate`).** Duplicating a local workspace off a local git
 repository provisions a fresh checkout for the copy at `<root>/<newId>/<repo-slug>` on a
 branch named for the new id (uniquified against the source repo's local and
-remote-tracking branches), using the **same decision matrix as `workspace.create`**:
-`workspace.cowIsolation` off ⇒ linked worktree (`checkoutMode: "worktree"`); on ⇒ CoW
-probe from the repository directory to `<root>/<newId>` — supported ⇒ standalone CoW
-clone (`checkoutMode: "cow"`), Unsupported/probe error ⇒ the duplicate **falls back to
-a linked worktree** with a logged warning (same fallback semantics as create).
-Provisioning is skipped for remote / skip-isolation sources and for a `repositoryPath`
-that is not a local git repository. An ordinary provisioning failure is logged and
+remote-tracking branches). The decision depends on whether the **source** checkout is
+standalone:
+
+- **Standalone source** (the source's own `checkoutMode` is `"cow"` or `"direct"`, which
+  includes every cache-hydrated workspace and an `isNewRepo` `direct` workspace): the
+  duplicate is **always standalone too, and never a linked worktree**. The clone source
+  is the source's **own checkout** (`worktreePath`, or `repositoryPath` when the
+  repository itself is the checkout), not the source's `repositoryPath`. A CoW probe from
+  that checkout to `<root>/<newId>` decides the mode: supported ⇒ CoW clone
+  (`checkoutMode: "cow"`); Unsupported, probe error, or a CoW clone that still reports
+  Unsupported ⇒ a **plain local `git clone`** of the source checkout
+  (`checkoutMode: "direct"`) with a logged warning. A source checkout whose `.git` is a
+  gitfile skips the probe and goes straight to the local clone. `workspace.cowIsolation`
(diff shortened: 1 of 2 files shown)


docs/PROTOCOL.md
```diff
@@ -559,15 +559,45 @@ removed in the background as usual.
...
+  Unsupported ⇒ a **plain local `git clone`** of the source checkout
+  (`checkoutMode: "direct"`) with a logged warning. A source checkout whose `.git` is a
+  gitfile skips the probe and goes straight to the local clone. `workspace.cowIsolation`
+  is **ignored** for standalone sources (parity with cache-hydrated create). Like a
+  hydrated create, the checkout *is* the repository: the duplicate persists
+  `repositoryPath` = its **own** checkout path, and `origin` is retargeted at the source's
+  own resolved `origin` — a network URL and an absolute local path carry over verbatim, a
+  relative local path is absolutized against the source repository so it still names the
+  same upstream, and `origin` is **removed** when the source has none or when it resolves
+  to the source checkout itself — so nothing references the source workspace's directory.
```
panghy: **Accuracy — should fix (overstates the CoW path).** These `origin`-resolution rules are written as applying to the whole standalone-source bullet, but in the code they run **only in the local-clone (`direct`) fallback**: `resolve_source_origin` is called solely from `provision_local_clone_checkout` (`intent-git/src/cow_checkout.rs` L288–295), and the services duplicate path never retargets remotes after a CoW clone (the only `set_remote_url` call in `intent-services/src/lib.rs` is in the cache-hydrated *create* arm, L10434). A CoW clone copies the source checkout's `.git/config` verbatim, so its `origin` is the source's origin **unresolved**: a network URL, an absent origin, and an absolute local path coincide with the documented outcome, but a *relative* local-path origin is carried verbatim (not absolutized — it then mis-resolves from the duplicate's directory), and an origin resolving to the source checkout itself is *not* removed. Suggested minimal fix: scope the resolution rules to the local-clone fallback (e.g. "`origin` carries over from the source's config (a CoW clone copies it verbatim); the local-clone fallback retargets it at the source's own **resolved** `origin` — …"), or note the CoW-mode caveat. As written, "so nothing references the source workspace's directory" is also not guaranteed on the CoW path for a self-referencing source origin (contrived, but the sentence claims it unconditionally).

docs/PROTOCOL.md
```diff
@@ -559,15 +559,45 @@ removed in the background as usual.
...
+
+- **Standalone source** (the source's own `checkoutMode` is `"cow"` or `"direct"`, which
+  includes every cache-hydrated workspace and an `isNewRepo` `direct` workspace): the
+  duplicate is **always standalone too, and never a linked worktree**. The clone source
+  is the source's **own checkout** (`worktreePath`, or `repositoryPath` when the
+  repository itself is the checkout), not the source's `repositoryPath`. A CoW probe from
+  that checkout to `<root>/<newId>` decides the mode: supported ⇒ CoW clone
+  (`checkoutMode: "cow"`); Unsupported, probe error, or a CoW clone that still reports
+  Unsupported ⇒ a **plain local `git clone`** of the source checkout
+  (`checkoutMode: "direct"`) with a logged warning. A source checkout whose `.git` is a
```
panghy: **Nit — "with a logged warning" is only true for two of the three fallback causes.** In `intent-services/src/lib.rs` L12239–12254, a probe **error** warns and the CoW-clone-still-Unsupported retry warns (L12433), but a clean `Ok(CowSupport::Unsupported)` probe result falls through to `direct` **silently** (L12243–12245) — reasonable, since `direct` is not a degradation of any user preference for a standalone source, unlike the shared-checkout path where Unsupported warns before the worktree fallback. Consider dropping the clause or scoping it ("probe error or a CoW clone that still reports Unsupported log a warning").

panghy (commented):
Verifier review of the Phase 2 bump + docs against the merged intentd code (`fd033b2`, intent-hq/intentd#956).

**Verified clean:**
- **Gitlink**: `packages/intentd` moves e2db1a9 → `fd033b25c369bd4cf2558b8824365a6f16e68f4e`, which is the squash commit of intentd#956 and the current tip of intentd `origin/main` (`git merge-base --is-ancestor` passes). No other submodule gitlink is touched.
- **Docs vs. code** (checked against `intent-services/src/lib.rs` duplicate path ~L12163–12550, `clear_standalone_repository_path` L5970, and `intent-git/src/cow_checkout.rs`): the standalone predicate (`checkoutMode` cow|direct), clone source (`worktreePath`, else `repositoryPath` when the repo is the checkout), gitfile-skips-probe, cowIsolation ignored for standalone sources, `repositoryPath` repointed at the duplicate's own checkout on success (L12513), cleared on provisioning failure (L12525/12545), committed-state-only local-clone fallback, the CoW-clone-still-Unsupported → direct retry (L12428), and the unchanged shared-checkout matrix all match the code exactly. `checkoutMode`-immutability and cache-hydrated-create text remain consistent; § cross-references and the full-form #1560 issue link are correct; no "wave" terminology.
- **Conventional commit / PR title**: single commit `docs: document standalone-source duplication; bump intentd` — valid type, matches the PR title.

Two doc-accuracy comments inline (one should-fix scoping issue on the `origin` rules, one nit on "with a logged warning"). Nothing else blocking.

docs/PROTOCOL.md
panghy: Confirmed and fixed in a4d8df9. You're right: `resolve_source_origin` is called only from `provision_local_clone_checkout` (cow_checkout.rs:290), so the CoW path — a reflink byte copy — keeps the source's `.git/config` verbatim and neither absolutizes a relative `origin` nor drops a self-referencing one.

The paragraph now scopes those rules explicitly to the `direct` path and says why: "The `direct` local-clone path additionally **resolves `origin`** (a CoW clone is a byte copy and keeps the source's `.git/config` verbatim)". The unconditional claim that survives is only the one the code actually guarantees for both modes — `repositoryPath` points at the duplicate's own checkout.

`docs/ARCHITECTURE.md` got the matching scoping on the duplication bullet.

Worth noting the residual gap this exposes is the same class of pre-existing CoW-clone limitation already tracked in intent-hq/monorepo#1582, not something introduced here.

docs/PROTOCOL.md
panghy: Correct — fixed in a4d8df9. A clean `Ok(CowSupport::Unsupported)` falls through to `Direct` with no log (lib.rs:12243-12245); only the probe `Err(e)` arm (lib.rs:12246), the clone-retry arm (lib.rs:12433), and the gitfile arm (lib.rs:12221) warn.

The sentence is now split so the warning attaches only to the paths that emit one: plain `Unsupported` ⇒ local clone, and "the same local-clone fallback is taken — with a logged warning — on a probe error or when a CoW clone still reports Unsupported despite a passing probe". Also marked the gitfile case as warned, which the previous text omitted.

PR #1594, opened 2026-08-07
docs: document owner-scoped ws.hook.cancel and bump intentd

### LINKED ISSUE #1563: intentd: ws.hook.cancel is not scoped to the owning agent — one agent can cancel another agent's hooks
## What happened

While shepherding the ACP session-cost feature (intentd#943 / cloudlands-fe#782 / monorepo#1561), the intentd implementor agent ran a cleanup loop over ws.hook.list() + ws.hook.cancel(hookId) to remove its own PR-watch hooks before merging. Because ws.hook.list() returns ALL hooks in the workspace and ws.hook.cancel() accepts any of them, the loop also cancelled a sibling agent's hook (the fe implementor's fe782-checks CI watch), silently killing that agent's wake condition.

In this instance the impact was benign (the fe CI had already been re-checked by the coordinator), but in general this can strand an agent indefinitely: it ends its turn expecting a hook wake that never comes, and nothing surfaces the cancellation to the hook's owner.

## Why it matters

- Hooks are the sanctioned mechanism for waiting on external conditions; an agent whose hook is cancelled by a sibling loses its only wake path.
- The natural cleanup idiom (list + cancel-all) is a foot-gun: nothing in the API signals ownership, so a well-meaning agent doing hygiene can break siblings.
- The cancelled agent gets no notification, so the failure mode is a silent stall rather than an error.

## Suggested fix (any of, in preference order)

1. Scope ws.hook.cancel to the calling agent's own hooks by default; reject cross-agent cancellation (or require an explicit force/agentId parameter for coordinator-style overrides).
2. Include an ownerAgentId field in ws.hook.list() results so callers can filter, and document the ownership semantics.
3. When a hook is cancelled by a non-owner, wake the owning agent with a cancellation notice instead of leaving it waiting silently.

Option 1 + 2 together would make the accidental case impossible while keeping deliberate coordinator intervention available.

## Repro sketch

1. Agent A schedules a hook (ws.hook.schedule) and ends its turn to wait.
2. Agent B runs: for (const h of await ws.hook.list()) await ws.hook.cancel(h.hookId)
3. Agent A's hook is gone; A is never woken and no one is notified.

panghy:
Phase 2 for [intent-hq/intentd#953](https://github.com/intent-hq/intentd/pull/953) (`fix: scope ws.hook.cancel to the owning agent`, squash-merged as `ea90d19`), which fixes #1563.

## Docs

`docs/PROTOCOL.md`:
- §5.40 — new **"Cancel is ownership-scoped on the MCP side"** block: MCP `ws.hook.cancel` may only cancel the calling agent's own hooks (non-owner cancel fails with a tool error naming the owner, before any state change — hook stays active, no `hook:cancelled`); the binding requires an agent caller context, mirroring `ws.hook.schedule`; owner self-cancel still delivers no self-wake. The caller-less wire/FE path is unchanged (cancels any hook, wakes the owner with the notice) and the archive sweep rides the same unscoped path.
- §5.40 `hook.cancel` table row clarified as the any-hook wire path.
- Version history: a "Within 5.1 (behavior only, no method-catalog or wire-shape change)" entry, following the existing convention.

`docs/ARCHITECTURE.md` — Background hooks: new **Ownership scoping** bullet covering `hook_cancel(caller: Option<AgentId>)`, the `Some`/`None` split, and the wake asymmetry.

Wire contract is untouched — `hook.cancel` params/result unchanged, so no protocol version bump.

## Submodule

`packages/intentd` gitlink bumped to `ea90d191837ad9651aeb12b0f7e940b09e22fb37` (intentd `main`).

## Verification

- Semantics checked against the merged intentd code (`hook_manager::hook_cancel_op`, `intent-core` `WorkspaceApi::hook_cancel`, the MCP `hook.rs` binding, and the `intent-transport` router FE arm) rather than the PR description alone.
- `git submodule status packages/intentd` → `ea90d191…` (matches intentd main).
- `make check` green.


docs/PROTOCOL.md
```diff
@@ -23,7 +23,7 @@ This document is the canonical wire contract between Intent clients (desktop, iO
...
...suffix AND codex evidence (provider column, `codex:` compound prefix, or a known effort-variant base model) so slash-bearing non-codex ids (e.g. HuggingFace-style unsloth ids) are untouched. The codex spawn path applies the session field as the `-c model_reasoning_effort=…` config override (an effort still embedded in a compound model id wins over the session field; the `CODEX_REASONING_EFFORT` env seam remains the last-resort fallback). No method-catalog change. Within 5.1 (behavior only, no method-catalog or wire-shape change — the wire `hook.cancel` params/result are untouched; [intent-hq/intentd#953](https://github.com/intent-hq/intentd/pull/953), monorepo#1563): **hook cancel is ownership-scoped on the MCP side** (§5.40) — hooks are agent-owned, so `ws.hook.cancel` now only cancels the calling agent's own hooks (a non-owner cancel fails with a tool error naming the owner, leaving the hook active and emitting no `hook:cancelled`; the binding requires an agent caller context, mirroring `ws.hook.schedule`), while the caller-less wire/FE path still cancels any hook in the workspace and wakes its owner with the cancellation notice; an owner's own cancel still delivers no self-wake.
```
panghy: **Version label is now wrong: this shipped within 5.2, not 5.1.** At `ea90d191` the daemon advertises `PROTOCOL_VERSION = "5.2"` (`crates/intent-transport/src/protocol.rs:44`), and the 5.2 bump commit (`5cab7c4`, intentd#946) is an *ancestor* of `ea90d191` — so no daemon build advertising 5.1 ever had the ownership-scoped cancel behavior. When this branch was cut the doc header still said 5.1, but the merge from `main` (9c5025d) brought in the Version 5.2 entry, and this sentence now (a) carries a stale label and (b) sits *after* the `Version 5.2` entry, breaking the grouping convention (the two existing `Within 5.1` entries — hook-name cap, was-running marker — correctly precede `Version 5.2`).

Suggested fix: relabel to `Within 5.2 (behavior only, no method-catalog or wire-shape change — the wire hook.cancel params/result are untouched; …)`. With that relabel the current end-of-paragraph position becomes convention-correct, so no move is needed.

docs/PROTOCOL.md
```diff
@@ -4976,6 +4976,29 @@ line is appended to that run's `lastLogs`). The schedule-time validation run per
 its returned state too. `hook:*` event payloads (§6.5) stay light
 and do **not** carry `lastLogs` or `lastState`; clients read them via `hook.list`.
 
+**Cancel is ownership-scoped on the MCP side** (within v5.1, behavior only;
```
panghy: Same version-label issue as the history entry: `(within v5.1, behavior only; …)` should be `(within v5.2, behavior only; …)` — the daemon already advertised 5.2 when intentd#953 merged (`PROTOCOL_VERSION = "5.2"` at `crates/intent-transport/src/protocol.rs:44`, with the 5.2 bump `5cab7c4` an ancestor of `ea90d191`).

Everything else in this block is accurate against the merged code: the non-owner rejection before any state change with the owner named in the error (`hook_manager.rs:704-711`), no `hook:cancelled` on rejection (emit only happens inside `cancel_active_hook`, line 744), the caller-context requirement mirroring `ws.hook.schedule` (`bindings/hook.rs:81-85`), the no-self-wake owner cancel vs. FE-cancel owner wake (`hook_manager.rs:718-722, 745-748`), the unchanged wire path (`router.rs:2958-2966` passes `None`), and the archive sweep riding the shared unscoped transition (`cancel_workspace_hooks` → `cancel_active_hook` with a wake notice, lines 768-800).

panghy (commented):
**Verification verdict: NOT APPROVED as-is — one fix required** (submitted as a comment review because the reviewing account authored the PR; treat this as request-changes).

Verified every factual claim against the merged intentd code at `ea90d191` (the intentd#953 squash-merge; gitlink matches exactly, commit confirmed on intentd `main`, and the diff contains only `docs/PROTOCOL.md`, `docs/ARCHITECTURE.md`, and the gitlink). One factual issue introduced by the merge-from-main: the change is labeled **within 5.1** but ships within protocol **5.2** — see the two inline comments.

**Claims verified against code (all accurate):**

- **Ownership check before any state change, error names the owner** — `hook_manager.rs:704-711`: the `caller != hook.agent_id` guard returns `InvalidParams("hook.cancel: hook {id} is owned by agent {owner} — you can only cancel your own hooks")` before the active-state check and before `cancel_active_hook` (which is what aborts the task, persists `cancelled`, and emits `hook:cancelled` at lines 737-744). Regression test `cross_agent_cancel_is_rejected_and_leaves_hook_active` (lines 2048-2093) asserts the error names the owner, state stays `Scheduled`, task stays alive, no `hook:cancelled`, no owner wake.
- **MCP binding requires an agent caller context, mirroring `ws.hook.schedule`** — `intent-acp/src/mcp_server/bindings/hook.rs:81-85` rejects `caller: None` before touching the service (tests at lines 154-201 prove both `cancel` and `schedule` guards reject pre-service).
- **Wake asymmetry** — `hook_manager.rs:720-722`: `notice = caller.is_none().then_some(...)`; owner cancel takes the no-wake branch (`resettle_owner_after_hook_terminal`, line 747), FE cancel wakes via `wake_hook_owner` (line 746). Tests: `owner_cancel_does_not_wake_owner` (2018-2041), `cancel_stops_task_and_fe_cancel_wakes_owner` (1975-2015).
- **Wire path unchanged / cancels any hook** — `intent-transport/src/router.rs:2958-2966`: the `hook.cancel` arm passes `caller: None` with identical params (`workspaceId`, `hookId`) and result `{ ok: true, hook }` (`hook_manager.rs:724`). No wire-shape change, so no version bump is correct.
- **Archive sweep rides the unscoped path** — `cancel_workspace_hooks` (`hook_manager.rs:768-800`) goes through the shared `cancel_active_hook` transition with an owner-wake notice, same as the `None`-caller path.
- **Cross-agent cleanup framing** — MCP `ws.hook.list` passes `agent_id: None` (`hook.rs:64`), so it does return every hook in the workspace with `agentId`, as §5.40 line 4998-4999 claims.
- **`docs/ARCHITECTURE.md` Ownership-scoping bullet (lines 367-376)** — matches `WorkspaceApi::hook_cancel(caller: Option<AgentId>)` (`intent-core/src/traits.rs:5285-5290`) and the `Some`/`None` split exactly.

**Style** — the §5.40 callout block, the table-row amendment, and the version-history sentence all follow the existing §5.40/PROTOCOL.md conventions (bold callout block, `Within N.M (behavior only, …)` phrasing with the intentd PR + monorepo issue references). The only convention break is the post-merge placement/label flagged inline.

Not pushing fixes per my review-only scope — the two inline comments are the only requested changes.

docs/PROTOCOL.md
panghy: Fixed in 3aa1b8e — relabeled to 5.2. Confirmed: doc header is now **Version: 5.2**, and the 5.2 entry precedes this note, so the end-of-paragraph position is convention-correct without a move.

docs/PROTOCOL.md
panghy: Fixed in 3aa1b8e — relabeled to 5.2. Confirmed: doc header is now **Version: 5.2**, and the 5.2 entry precedes this note, so the end-of-paragraph position is convention-correct without a move.

panghy:
Correction to my thread replies above: the fix commit is **c29adac6** (`docs: relabel hook-cancel scoping note as within v5.2`), not the placeholder SHA I typed.

Both spots now read 5.2 — the version-history entry (line 26) and the §5.40 block header (`(within v5.2, behavior only; …)`). The doc header is `**Version:** 5.2`, and the `Version 5.2 is an **additive** minor bump over 5.1` entry precedes this note in the paragraph, so its end-of-paragraph position is convention-correct with no move needed.

Issue #816, opened 2026-07-26
Remaining settings indicator toggles lack descriptive ariaLabel (Notifications, WebSocket API, Additional Agents)


panghy:
## Problem

Follow-up to #812 (fixed for the RTK and Custom MCP Servers toggles by intent-hq/cloudlands-fe#409). Several other indicator-variant `Toggle` instances on the Settings page still don't pass an `ariaLabel`, so their accessible name falls back to the generic "On"/"Off" labels from the shared component and screen readers can't tell which setting the switch controls.

Affected call sites (verified on cloudlands-fe `main` at `7ce9fa1b`, all in `packages/cloudlands-fe`):

- `src/lib/components/settings/NotificationSettings.svelte` — 3 toggles (~L71, ~L86, ~L101)
- `src/lib/components/settings/WebSocketApiSettings.svelte` — 1 toggle (~L246)
- `src/lib/components/settings/AdditionalAgentsSettings.svelte` — 1 per-provider toggle (~L56); needs a dynamic label, e.g. `ariaLabel={`Enable ${provider.name}`}`

`OpenInAppsSettings.svelte` already does this correctly (`ariaLabel={`Show ${editor.name} in Open In`}`) and can serve as the reference for the dynamic case.

## Suggested fix

Pass a descriptive `ariaLabel` at each call site matching the visible setting label (as done in RtkSettings/McpServersSettings/OpenInAppsSettings). No changes needed to the shared `Toggle` component.

## Context

Flagged as a non-blocking note during review of intent-hq/cloudlands-fe#409 (see monorepo PR #815 for the submodule bump).

Issue #824, opened 2026-07-26
Directory picker: typing a `~` path fails to navigate (FE expansion depends on listing.home; host.listDirectory rejects raw tilde)


panghy:
## Symptom

Typing `~/src` into the onboarding `DirectoryPickerModal` path bar and pressing Enter does not navigate there.

## Root cause

Two layers:

1. **FE**: `DirectoryPickerModal.svelte#expandTypedPath` only expands `~` when `listing?.home` is available; if the current listing hasn't loaded (or the initial `~/Developer` path failed to list), the raw `~/src` string is sent to the daemon unchanged.
2. **intentd**: `host.listDirectory` does not expand a leading `~`, so the raw path fails with a not-found error, leaving the picker stuck with a path error.

The picker's `initialPath` is also passed to the daemon unexpanded (`~/Developer`), which fails when that literal path can't be resolved and forces the home-directory fallback.

## Proposed fix

- **intentd**: expand a leading `~`/`~/` in `host.listDirectory` (same daemon-boundary expansion as the workspace.create/git.clone tilde issue filed alongside this one).
- **FE**: keep the client-side expansion as a fast path, but it must not be a correctness requirement — typed tilde paths should work even when `listing.home` is unavailable.

## Acceptance criteria

- Typing `~/src` + Enter navigates to `$HOME/src` even on a fresh picker whose initial listing failed.
- `host.listDirectory` with `path: "~"` and `path: "~/src"` returns the expanded listing.

Observed while dogfooding onboarding, 2026-07-26.

PR #1641, opened 2026-08-07
chore: fold auggie into generic provider path (submodule bumps + protocol docs)

panghy:
Phase 2 monorepo update for the auggie generic-provider folding.

## Submodule bumps

- `packages/intentd` → `558ce3e` — includes [intent-hq/intentd#977](https://github.com/intent-hq/intentd/pull/977) (`feat(providers): probe auggie auth via token print, drop checkAuggie version`).
- `packages/cloudlands-fe` → `6b8c4747` — includes [intent-hq/cloudlands-fe#824](https://github.com/intent-hq/cloudlands-fe/pull/824) (`refactor: fold auggie into the generic provider path`), plus the intervening `#816` fix already on main.

## docs/PROTOCOL.md

- New `host.checkAuggie` section under the §5 fast-path notes: resolution-only, payload `{ available, path? }`. The `version` field is retired and there is no `--version` spawn; resolution precedence (`context.auggiePath` → `providers.paths.auggie` → auto-detection) is unchanged.
- `host.providerAuthStatus`: documents that probe commands come from the registry `auth_check_args` and ride the generic exit-code arm with stdout/stderr discarded. auggie now probes with `auggie token print` — the discarded output matters because the command prints the auth session secret. The bespoke `auggie model list` output-sniffing probe is retired.
- `providers.catalog` auggie example row gains `loginDocsUrl` (the generic provider row's login CTA reads it).
- Protocol 6.0 summary records the folding as a within-6.0, no-method-catalog change (no version bump).

## Verification

- `make check` — clean (`cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`).
- `make test` — 2586 passed, 1 flaky failure in `events::skills_watcher::tests::resume_catch_up_survives_a_discovery_cache_refresh_while_suspended` (filesystem-watcher timing, unrelated to this docs/gitlink-only change); passes in isolation on re-run.


panghy:
Rebased on `main`: the `packages/cloudlands-fe` bump landed independently via monorepo `8cf8b9d1` (main now points at `67c5f8b4`, which already contains cloudlands-fe#824 / `e7c2d83f`). This PR therefore now carries only the `packages/intentd` bump to `558ce3e` plus the `docs/PROTOCOL.md` updates.

docs/PROTOCOL.md
```diff
@@ -390,6 +405,7 @@ Daemon-owned provider auth probes: reports whether each CLI-backed agent provide
...
+- **Probe commands** come from the provider registry's `auth_check_args` and ride a **generic exit-code arm** — exit 0 ⇒ `true`, non-zero ⇒ `false` — with the child's stdout and stderr **discarded**, never captured, logged, or surfaced. `auggie` probes with `auggie token print` ([intent-hq/intentd#977](https://github.com/intent-hq/intentd/pull/977)): it has no bespoke probe any more, and the discarded output matters here because the command prints the auth session secret. The former `auggie model list` output-sniffing probe is retired.
```
panghy: This bullet reads as a blanket statement about all `host.providerAuthStatus` probes, but in `check_provider_auth_cli` (intentd `558ce3e`, `crates/intent-services/src/provider_auth.rs`) only the default arm nulls stdout/stderr and maps exit 0 ⇒ `true` / non-zero ⇒ `false`. It doesn't hold for four of the seven probe-able providers:

- **grok** — stdout is **piped** and parsed for auth markers; exit 0 does *not* imply `true` (exit 0 with no markers/models is `null`).
- **opencode** — stdout is **piped**; `true` additionally requires at least one `provider/model` line beyond exit 0.
- **droid** — probed via ACP (`probe_droid_auth`), not `auth_check_args` at all.
- **pi** — probed via the pinned npx adapter (`probe_pi_auth`), not `auth_check_args`.

Since this bullet is the only probe-mechanics description in the section (nothing elsewhere in PROTOCOL.md covers the grok/opencode/droid/pi bespoke arms), suggest scoping the first sentence, e.g.: "CLI-probed providers run their registry `auth_check_args`; most (`auggie`, `claude-code`, `codex`) ride a generic exit-code arm — exit 0 ⇒ `true`, non-zero ⇒ `false` — with the child's stdout and stderr discarded, never captured, logged, or surfaced; `grok` and `opencode` keep bespoke output-sniffing arms, and `droid` / `pi` probe via their adapters instead." The auggie-specific sentences that follow are accurate as-is.

panghy (commented):
Verified this Phase 2 bump against the merged submodule code (intentd `558ce3e`, cloudlands-fe `6b8c4747`). Summary:

**Gitlinks — both on main** ✅
- `packages/intentd` → `558ce3e` is exactly the squash-merge commit of [intent-hq/intentd#977](https://github.com/intent-hq/intentd/pull/977) and an ancestor of `intentd/main` (currently 1 behind tip — intentd#979 landed after this PR was cut; not a problem).
- `packages/cloudlands-fe` → `6b8c4747` is the merge commit of cloudlands-fe#816 and an ancestor of `cloudlands-fe/main`; [intent-hq/cloudlands-fe#824](https://github.com/intent-hq/cloudlands-fe/pull/824)'s merge commit `e7c2d83f` is an ancestor of it, so the gitlink includes #824 + #816 exactly as the PR body says (currently 1 behind tip — fe#817 landed after; not a problem).

**docs/PROTOCOL.md claims verified against `558ce3e`** ✅
- `host.checkAuggie` resolution-only, `{ available, path? }`, no `version` field, no `--version` spawn: matches `check_auggie_with` in `crates/intent-transport/src/host_ops.rs` (`available:true` iff a path resolved, `path` omitted otherwise; nothing is executed). Covered by `check_auggie_is_resolution_only` / `handle_check_auggie_uses_configured_path` (the latter asserts `version` is absent).
- Resolution precedence `context.auggiePath` → `providers.paths.auggie` → auto-detection: matches `configured_auggie_path` (host.rs) + `resolve_auggie_path` (host_ops.rs, file-or-symlink acceptance, falls through to `auggie_discovery::find_auggie`).
- auggie auth probe = registry `auth_check_args` `["token", "print"]` (config.rs:269) on the generic exit-code arm of `check_provider_auth_cli` with stdin/stdout/stderr all `Stdio::null()` — auggie has no bespoke match arm anymore; the former `auggie model list` output-sniffing probe is gone. ✅ (one scoping nit on the doc bullet — see inline comment)
- `providers.catalog` auggie row `loginDocsUrl`: `login_docs_url: Some("https://docs.augmentcode.com/cli/overview")` in the registry, serialized as `loginDocsUrl` by `provider_catalog.rs` — matches the doc example verbatim.
- Protocol 6.0 summary addition: accurate — within-6.0, no method-catalog change (`host.checkAuggie` remains in `catalog.rs`), and the fe#824 claim checks out (the version gate — `settings_providers_needsUpdate` strings and the git/auggie version-check plumbing — was removed in that PR).

**Informational, non-blocking:** the FE at `6b8c4747` still carries a vestigial `version?: string` in the two `host.checkAuggie` response type annotations (`auggie-path.ts`, `auggie.ipc.ts`). It is never read, so the doc's "only consumer removed" statement stays true; cleanup could ride any future fe change.

One inline comment on the probe bullet's phrasing; everything else reflects the merged code accurately. Not merging per reviewer scope.

docs/PROTOCOL.md
panghy: Good catch — the blanket phrasing was wrong. Fixed in `943605c5` with the scoped wording: `auggie` / `claude-code` / `codex` on the generic exit-code arm with output discarded, `grok` and `opencode` called out as bespoke output-sniffing arms (piped stdout; grok exits 0 in both auth states, opencode needs a `provider/model` line beyond exit 0), and `droid` / `pi` noted as adapter probes rather than `auth_check_args`.

The vestigial `version?: string` in the FE's checkAuggie type annotations is out of scope for this monorepo PR (docs + gitlinks only) — worth a follow-up on cloudlands-fe if it bothers anyone.

PR #1784, opened 2026-08-09
ci: remove zero-job push trigger from CI workflow

panghy:
Part of the CI audit to stop redundant runs on push to main (workspace: audit-cis).

## What

Change the `CI` workflow trigger from `on: [push, pull_request]` to `on: pull_request`.

## Why

The only job (`pr-title`) is gated with `if: github.event_name == 'pull_request'`, so every push to main spawned a zero-job, all-skipped noise run. Dropping the push trigger removes that noise with no behavior change for PRs — `pr-title` still runs on every pull request.

## Verification

- YAML remains valid (single-line trigger change).
- `pr-title` check runs on this PR itself.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,5 @@
 name: CI
-on: [push, pull_request]
+on: pull_request
 
 permissions:
   contents: read


panghy:
**Review (verifier agent): no concerns — approve-equivalent.**

Verified:

- **YAML validity** — `.github/workflows/ci.yml` parses cleanly; `on` resolves to `pull_request` (scalar form is valid workflow syntax).
- **PR behavior unchanged** — `pr-title` is still triggered by every `pull_request` event and its `if: github.event_name == 'pull_request'` gate remains satisfied. Evidence: the `pr-title` check ran and **passed on this PR itself** (run 31302056747).
- **No push-trigger dependents** —
  - `ci.yml` is the only workflow in this repo; no `workflow_run` triggers reference the `CI` workflow.
  - No README badges or docs reference `workflows/ci.yml`.
  - Branch protection: `main` has no classic protection, and both active rulesets ("Default Branch", "Default") contain **no** `required_status_checks` rules — so nothing depends on a push-event `pr-title` status on `main`.

Non-blocking follow-up: with the push trigger gone, the `if: github.event_name == 'pull_request'` gate on `pr-title` is now redundant and could be dropped in a future cleanup. Harmless to keep.

panghy:
Ack — keeping the pr-title if-guard as-is per the task's minimal-diff preference; noted as a possible future cleanup. Merging.

PR #1798, opened 2026-08-09

### LINKED ISSUE #1776: intentd: repo cache never initializes/updates submodules — cache-hydrated workspace checkouts (CoW/direct) arrive with empty submodule dirs
## Problem

The daemon-managed repo cache (`intent-git/src/repo_cache.rs`, `ensure_cached_repo`) is submodule-unaware end to end:

- **Cache miss (clone):** `clone()` runs a plain `git clone <url>` with no `--recurse-submodules`, so submodule paths in the cache are empty directories with no content.
- **Cache hit (refresh):** the refresh path runs `git fetch --prune`, `remote set-head origin --auto`, hard reset to the default branch, and `git clean -fdx` — but never `git submodule sync` / `git submodule update --init --recursive`. Even if submodules were somehow populated, `clean -fdx` (without submodule awareness) and the reset would leave them stale or wiped relative to the recorded gitlinks.

Because `workspace.create` hydration provisions a **standalone** checkout by CoW-cloning (or plain local-cloning) the cache directory byte-for-byte, every cache-hydrated workspace inherits the uninitialized submodules. For a repo like `intent-hq/monorepo` (three submodules: `packages/intentd`, `packages/cloudlands-fe`, `packages/ios`), a freshly created workspace lands with all package directories empty, and the first agent in the workspace has to pay the full network cost of `git submodule update --init --recursive` itself — on every new workspace, defeating the point of the cache.

## Expected

When a repo is cached (and on each refresh), submodules should also be initialized and updated as needed — e.g. `--recurse-submodules` on the initial clone, and `git submodule sync --recursive` + `git submodule update --init --recursive` (with the same fail-fast env, wall-clock deadline, and token-via-credential-helper handling as the existing network git calls) during refresh. Then the CoW/local clone of the cache immediately carries fully populated submodule working trees, and new workspaces are usable at creation time with no per-workspace network fetch.

## Notes / considerations

- Refresh must stay fail-soft in the established way: a submodule anomaly should at worst wipe-and-re-clone the cache, never fail `workspace.create`.
- `git clean -fdx` in the refresh path needs to be reconciled with populated submodules (it currently would treat their content as removable, or conversely leave stale content the reset does not touch); `git submodule update` after the reset should make each submodule match the recorded gitlink.
- Submodule remotes may need the same auth token as the superproject (same-host GitHub submodules should work via the existing env-backed credential helper); private submodules on other hosts can be best-effort.
- The hydrated checkout's `origin` is retargeted to the real URL after provisioning; submodule URLs inside the clone are absolute (from `.gitmodules`) so they should already be correct, but relative `.gitmodules` URLs should be verified against the retargeting step.
- Related but distinct: #1714 (attributed commit helper flattening a submodule), #1125 (CoW clone speed).

## Where

- `packages/intentd/crates/intent-git/src/repo_cache.rs` — `clone()`, `refresh()` (no submodule handling today)
- `packages/intentd/crates/intent-git/src/cow_checkout.rs` — hydration clone of the cache (byte-copies whatever the cache has)
- `docs/ARCHITECTURE.md` §"Repo cache & cache-hydrated creation" documents the current (submodule-unaware) behavior

panghy:
## Summary

- Bumps `packages/intentd` to latest main (`26f08b6`), picking up intent-hq/intentd#1024 — `feat(git): submodule-aware repo cache and hydration` (fixed intent-hq/monorepo#1776 and intent-hq/monorepo#1782).
- Updates `docs/ARCHITECTURE.md` §"Repo cache & cache-hydrated creation" to describe the submodule-aware cache clone/refresh (recursive clone, submodule sync/update on refresh, orphan-aware clean) and the strict no-network hydration guarantee (`--no-fetch`, clone transports refused; anomalies degrade with a warning and never fail `workspace.create`).

## Verification

- `make check` and `make test` green on this branch (5725/5725 tests passed).

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -158,22 +158,33 @@ Wire contract: PROTOCOL.md §5.1 (`checkoutMode`, `cowSupported`), §5.5/§5.5a
   `<workspaces_root>/.repo-cache/<owner>/<repo>` (dot-prefixed so it stays invisible to
   users and to recent-repo derivation; the module never reads config — the caller passes
   the cache root). `ensure_cached_repo` is the single entry point: it serializes callers
-  on a per-repo lock, then clones fresh (miss) or refreshes (`git fetch --prune`,
-  `remote set-head origin --auto` so an upstream default-branch change is re-resolved,
-  hard reset to that branch, then `git clean -fdx` so untracked pollution is never
-  byte-copied into hydrated checkouts). **Refresh never fails the flow** — any anomaly
+  on a per-repo lock, then clones fresh (miss; `--recurse-submodules`, so the cache
+  carries populated submodule work trees and their module git dirs) or refreshes
+  (`git fetch --prune`, `remote set-head origin --auto` so an upstream default-branch
+  change is re-resolved, hard reset to that branch, `submodule sync` + `submodule update
+  --init --recursive --force` so gitlink bumps, URL changes, and newly added submodules
+  are followed, then `git clean -ffdx` plus a per-submodule recursive clean so untracked
+  pollution — including orphaned submodule checkouts — is never byte-copied into
+  hydrated checkouts). **Refresh never fails the flow** — any anomaly
   (diverged history, corrupt object store, an interrupted prior clone, a vanished
   `origin/HEAD`, a mismatched `origin`) deletes the cache dir and re-clones; only a
   failed clone surfaces as an error. `intent-services` uses the cache to hydrate
   `workspace.create` when a `githubUrl` arrives without a `clonePath` (PROTOCOL §5.1):
   the checkout is always **standalone** — a CoW clone of the cache (`cow`) or a plain
   local `git clone` of it (`direct`) — never a linked worktree against the cache, whose
-  hard-reset/re-clone refresh would corrupt linked worktrees. Provisioning holds the
-  per-repo cache lock, and afterwards `origin` is retargeted at the real URL so the
-  checkout is fully self-contained and the cache is always safe to delete. Network git
-  here shells out to system `git` (fail-fast `GIT_TERMINAL_PROMPT=0`, wall-clock
-  deadline kill) with any token offered through the env-backed credential helper, never
-  argv.
+  hard-reset/re-clone refresh would corrupt linked worktrees. Both paths populate
+  submodule work trees from the cache's local module git dirs alone (the CoW byte copy
+  carries them; `direct` copies `.git/modules` before the update), and the populating
+  `submodule update` runs **strictly offline** (`--no-fetch`, every clone transport
+  refused): hydration never touches the network — a gitlink the cache does not hold
+  degrades to an unpopulated submodule with a warning, and no submodule anomaly ever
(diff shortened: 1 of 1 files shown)


docs/ARCHITECTURE.md
```diff
@@ -158,22 +158,33 @@ Wire contract: PROTOCOL.md §5.1 (`checkoutMode`, `cowSupported`), §5.5/§5.5a
...
+  `submodule update` runs **strictly offline** (`--no-fetch`, every clone transport
+  refused): hydration never touches the network — a gitlink the cache does not hold
+  degrades to an unpopulated submodule with a warning, and no submodule anomaly ever
+  fails `workspace.create`. Provisioning holds the per-repo cache lock, and afterwards
+  `origin` is retargeted at the real URL (submodule URLs re-synced to their
+  `.gitmodules` resolution) so the checkout is fully self-contained and the cache is
+  always safe to delete. Network git here shells out to system `git` (fail-fast
+  `GIT_TERMINAL_PROMPT=0`, wall-clock deadline kill) with any token offered through the
+  env-backed credential helper, never argv; the token propagates to submodule child
+  fetches during cache clone/refresh via `GIT_CONFIG_PARAMETERS`.
```
panghy: Minor precision nit (non-blocking): per `intent-git/src/auth.rs`, what propagates to submodule child fetches via `GIT_CONFIG_PARAMETERS` is the credential-helper *config entry* — "The config value never contains token bytes; the token travels only under `TOKEN_ENV`" (the inherited env var). As written, "the token propagates … via `GIT_CONFIG_PARAMETERS`" could be read as token bytes living in that variable, which contradicts the preceding "never argv / env-backed" clause. Suggest e.g. "…never argv; the helper config propagates to submodule child fetches during cache clone/refresh via `GIT_CONFIG_PARAMETERS`, the token via the inherited env var."

panghy (commented):
Reviewed the submodule bump and the ARCHITECTURE.md update.

- **Gitlink**: `26f08b6d40b90867cfa30860a0cb4cdc211e5598` verified as the squash-merge of intent-hq/intentd#1024 and identical to the current tip of intentd `main`.
- **Doc accuracy**: checked every claim in the updated §"Repo cache & cache-hydrated creation" against `crates/intent-git/src/repo_cache.rs` and `cow_checkout.rs` at that commit — recursive clone, refresh sequence (`submodule sync --recursive` + `update --init --recursive --force`, `clean -ffdx` + per-submodule recursive clean), anomaly self-heal, strictly-offline hydration (`--no-fetch` + `protocol.allow=never`), warning-degrade never failing `workspace.create`, and origin retarget / submodule URL re-sync all match the implementation. Style and density are consistent with the surrounding bullets.
- **Conventions**: conventional title, single conventional branch commit, body references the intentd PR and the fixed monorepo issues.

One minor wording nit inline (non-blocking). Otherwise looks good.

docs/ARCHITECTURE.md
panghy: Fixed in the follow-up commit: reworded to distinguish the helper config (via `GIT_CONFIG_PARAMETERS`) from the token bytes (inherited env var only).

Issue #1044, opened 2026-07-28
fe: stale 'Response failed' banner persists after a successful queued retry (no error clear on enqueue success or drain promotion)


panghy:
## Symptom

After a turn fails ("Response failed" banner up), retrying while the daemon still reports the agent busy leaves the stale banner visible — and since the error state also suppresses the Thinking/streaming indicator, the UI looks dead even though the daemon accepted the retry and will run it.

## Root cause — two gaps in `cloudlands-fe`

1. **Queue-on-send success branch never clears the error** (`src/features/agent/chat-send-service.ts`, `dispatchToLifecycle`): when `selectAgentIsResponding` is true, the retry routes through `agent.queueMessage`. On a *successful* enqueue, nothing dispatched `chatErrorCleared` / `chatModelUnavailableCleared`, so the previous turn's failure banner stayed up. (The direct-send path clears via `chatSendStarted`; the queue path had no equivalent.)

2. **Drain promotion never clears the error** (`src/store/renderer/slices/chat-state/chat-state-slice.ts`, `reduceQueueSnapshotDiff`): when a queued entry drains and its parked retry record is promoted into `lastAttemptedMessage` (#999), the promoted record's turn is now the active turn — but a stale `error` / `modelUnavailable` from a previous turn persisted over it.

## Fix

- On **successful enqueue** in the queue-on-send branch, dispatch `chatErrorCleared` + `chatModelUnavailableCleared`. The enqueue-**failure** branches keep setting the error (#969 semantics preserved).
- In `reduceQueueSnapshotDiff`, a **genuine drain promotion** also clears `error` and `modelUnavailable`. The clear-queue signature branch (empty snapshot, >1 vanished id — those entries never run) and the pure text-sync path (#1011) deliberately do NOT clear.

## Related but distinct

- #941 (stale `lastAttemptedMessage` after edit-and-regenerate)
- #969 (enqueue-failure branches must pair the banner with the failed message)
- #984 (disposition-neutral `agent:stream:end` vs retry-record clearing)
- #999 (turn-scoped retry records / drain promotion)
- #1022 (missing turn-correlation ids forcing FE-side attribution)

Fixed by intent-hq/cloudlands-fe PR (referenced below).

panghy:
[...]

PR #1843, opened 2026-08-09
chore: update intentd submodule to latest main (pi CLI version gate)

panghy:
Phase 2 follow-up to [intent-hq/intentd#1044](https://github.com/intent-hq/intentd/pull/1044) (pi CLI version gate, >= 0.80.4; see intent-hq/monorepo#1662 — already auto-closed by the intentd merge).

## Changes

- **`packages/intentd`**: bump the submodule to `9df2fd0d` (latest `main`, the intentd#1044 squash commit).
- **`docs/PROTOCOL.md`**: document the shipped `host.providerDiscovery` pi-row CLI verdict fields exactly as landed:
  - Always-present on the probed pi row: `cliCommand`, `cliResolved`, `cliVersionOk`, `cliRequirement` (`"Pi CLI 0.80.4+"`).
  - Present when known (omitted, never null, otherwise): `cliResolvedPath`, `cliVersion`.
  - Gating semantics: a missing or confirmed-too-old `pi` CLI forces `installed: false` with an actionable `unavailableReason` (found version, requirement, adapter pin); an inconclusive probe (Unknown) is permissive — WARN only, never gated.
  - Fields appear only on the pi row, and only when it is not `gatedOff` (gated rows are never probed); resolution scans the spawn-time enhanced PATH so the probe reports the same binary the pi-acp child would exec.
  - Adds the matching "Also within 6.2 (additive response fields …)" entry to the Protocol Version & Compatibility narrative (no method-catalog change, so no version bump).

## References

- intent-hq/intentd#1044
- intent-hq/monorepo#1662


docs/PROTOCOL.md
```diff
@@ -453,6 +470,8 @@ Daemon-owned provider discovery: reports which CLI-backed agent providers are in
...
+- **`pi` CLI gating** — a **missing** or **confirmed-too-old** CLI gates the pi row off: `installed` is forced to `false` and `unavailableReason` (optional string, pi row only) carries an actionable message naming the found version (when too old), the requirement, and the adapter pin (e.g. `"pi CLI not found — Pi CLI 0.80.4+ is required by pi-acp@0.0.33"`). An **inconclusive** probe — spawn failure, timeout, unparseable `--version` output, or a relative separator-carrying `PI_ACP_PI_COMMAND` override that did not resolve from the daemon's CWD — is **permissive**: the daemon logs a WARN and does not gate, so `cliVersionOk` is `false` but `installed` is untouched and `unavailableReason` is omitted (a changed `--version` format never false-negatives the provider). Invariant: `unavailableReason` present ⇒ `installed: false` and `cliVersionOk: false`. The same gate fails agent creation fast with a clear error (instead of a silent hang) when a Pi agent is spawned against a missing/too-old CLI.
```
panghy: Non-blocking nit: "gates the pi row off" risks conflation with the distinct `gatedOff` field documented a few bullets up (and referenced in the previous bullet's "only when it is not `gatedOff`"). The CLI gate forces `installed: false` + `unavailableReason` but never sets `gatedOff` — the shipped code and tests confirm the CLI-gated row carries no `gatedOff`. Consider phrasing like "marks the pi row unavailable" / "forces the pi row uninstallable" to keep "gated off" reserved for the `gatedOff` mechanism. Same phrase appears in the §Version narrative entry. Fine to leave as-is since the sentence immediately defines what it means.

panghy (commented):
Verified this Phase 2 follow-up against the merged intent-hq/intentd#1044 diff. Everything checks out; one non-blocking wording nit inline.

**Submodule gitlink** ✅
- `packages/intentd` now points at `9df2fd0db147911056c612fd410114a29e6a9a75`, which is exactly the intentd#1044 squash-merge commit and the current `main` HEAD of intent-hq/intentd.

**Docs vs. shipped payload** ✅ (checked field-by-field against the #1044 diff)
- Field names/presence match `apply_pi_cli_fields`: always-present `cliCommand` / `cliResolved` / `cliVersionOk` / `cliRequirement`; `cliResolvedPath` / `cliVersion` inserted only when known (omitted, never null).
- `cliRequirement` string `"Pi CLI 0.80.4+"` matches `PI_CLI_REQUIREMENT`; the example's `pi-acp@0.0.33` matches `PI_ACP_NPX_PACKAGE`.
- Both `unavailableReason` examples are byte-identical to the `pi_gate_reason` format strings (too-old: `"pi CLI 0.79.0 is too old — Pi CLI 0.80.4+ is required by pi-acp@0.0.33"`, asserted verbatim in the unit tests; not-found variant likewise).
- Gating semantics match: Missing/TooOld force `installed: false` + reason; `PiCliGate::Unknown` is permissive (WARN, `installed` untouched, no reason). The documented invariant (`unavailableReason` ⇒ `installed: false` ∧ `cliVersionOk: false`) is asserted in tests.
- Inconclusive conditions match `probe_pi_cli` / `run_version_probe`: spawn failure, 3s timeout, unparseable `--version`, and the relative separator-carrying `PI_ACP_PI_COMMAND` override that didn't resolve (degraded to Failed → Unknown by design).
- Pi-row-only + not-`gatedOff` scoping matches the discovery guard (`p.id == "pi" && p.gated_off.is_none()`), probed once per discovery call as documented.
- `cliCommand` resolution (non-empty `PI_ACP_PI_COMMAND` else bare `pi`) and the spawn-time enhanced-PATH scan (npx's parent dir + `~/.augment/bin` ahead of enriched/inherited dirs, via `enhanced_path_spawn_dirs`) match the code.
- `cliVersion` = trimmed first stdout line of `pi --version` matches `run_version_probe`.
- Spawn-side fail-fast (`check_pi_cli_for_spawn`, InvalidInput so the message survives the RPC envelope) backs the "fails agent creation fast" claim, with an e2e test.

**Version narrative** ✅
- The "Also within 6.2 (additive response fields, presence-detected per the convention below; no method-catalog change, so no version bump …)" entry follows the established convention (matches the intentd#1041/#1039 entries in form), is placed chronologically after the #1040 entry that carries the 6.2 constant bump, and links both intentd#1044 and monorepo#1662. #1044 adds no router methods, so no-bump is correct.

**Style** ✅ — the two new bullets follow the section's existing bold-lead-in additive-field convention (cf. `secondaryCommand`/`secondaryResolvedPath`), and the example-JSON comments match the surrounding `// optional — …` style.

CI: `pr-title` passes. Cannot approve my own PR — posting as a comment review; content-wise this is an approval.

docs/PROTOCOL.md
panghy: Addressed in ccb16af — reworded to "marks the pi row unavailable (it never sets gatedOff, which stays reserved for the env-var/feature-code mechanism above)" here and "marks the row unavailable (never via gatedOff)" in the version narrative.

PR #1877, opened 2026-08-10
fix: skip private packages/ios submodule for public clones

### LINKED ISSUE #418: Handle private ios submodule in public monorepo
iOS will NOT launch on Aug 17 and `intent-hq/ios` stays private. A public monorepo would carry a broken `packages/ios` submodule pointer for external users (`git clone --recursive` fails or 404s).

Options:
- Remove `packages/ios` from `.gitmodules` until iOS goes public (re-add later)
- Keep gitlink but document it as optional/internal and make setup scripts skip it

Decide + implement before flipping monorepo public. Launch blocker.

panghy:
## Summary

Prepares the monorepo for going public while `intent-hq/ios` stays private: keeps the `packages/ios` gitlink (and its version pin) but marks it `update = none` in `.gitmodules`, so external users' recursive clones skip it cleanly instead of failing with a 404/credential prompt.

Fixes intent-hq/monorepo#418

## Changes

- **.gitmodules**: `update = none` for `packages/ios` — `git clone --recurse-submodules` and `git submodule update --init --recursive` now skip it (exit 0, empty dir).
- **Makefile**:
  - `ensure-ios-submodule` uses `--checkout` to force-init despite `update = none` (internal devs with access).
  - `ensure-submodules` ios leg is `--checkout` + fail-soft warning when the clone fails due to no access.
  - `update` skips uninitialized `update = none` submodules with a notice instead of erroring (incl. the summary loop).
- **Docs** (README, CONTRIBUTING, AGENTS): `packages/ios` documented as private/optional, skipped by default, initializable via `make ensure-ios-submodule`; README build-from-source simplified to plain `git clone --recurse-submodules`.

## Verification

- Fresh `file://` recursive clone: exit 0, "Skipping submodule 'packages/ios'", ios dir empty, intentd + fe checked out; `git submodule update --init --recursive` in the clone: exit 0, ios still empty.
- `make ensure-submodules` / `make ensure-ios-submodule`: succeed both with ios uninitialized and as no-ops when initialized.
- Independently re-verified end-to-end by a verifier agent.

## Notes

- Follow-up when iOS goes public: delete the `update = none` line.
- Known git quirk (outside the Makefile): a global `submodule.recurse=true` makes `git pull` itself try to clone ios during `make update`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,4 @@
 	path = packages/ios
 	url = https://github.com/intent-hq/ios.git
 	branch = main
+	update = none
--- a/Makefile
+++ b/Makefile
@@ -82,11 +82,21 @@ all: build
 help: ## List documented targets
 	@awk 'BEGIN {FS = ":.*## "} /^[a-zA-Z0-9_-]+:.*## / {printf "  %-16s %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
-ensure-submodules: ## Initialize any missing submodules — intentd, FE, iOS (idempotent)
+# The iOS submodule is private and marked `update = none` in .gitmodules, so
+# the generic `git submodule update --init` would silently skip it while
+# claiming success. Its leg passes `--checkout` to override `update = none`,
+# and fails soft (warning, not error) when the clone fails — e.g. no access to
+# the private repo.
+ensure-submodules: ## Initialize any missing submodules — intentd, FE, iOS (idempotent; iOS is fail-soft)
 	@for sm in $(SUBMODULES); do \
 		if [ ! -e "$$sm/.git" ]; then \
 			echo "[ensure-submodules] initializing $$sm"; \
-			git submodule update --init --recursive "$$sm" || exit 1; \
+			if [ "$$sm" = "$(IOS_DIR)" ]; then \
+				git submodule update --init --checkout --recursive "$$sm" \
+					|| echo "[ensure-submodules] WARNING: could not initialize $$sm (private repo; skipping — check GitHub access if you need it)"; \
+			else \
+				git submodule update --init --recursive "$$sm" || exit 1; \
+			fi; \
 		else \
 			echo "[ensure-submodules] $$sm already initialized — leaving as-is"; \
 		fi; \
@@ -114,11 +124,13 @@ ensure-fe-submodule:
 	fi
 
 # On-demand init for the iOS submodule — pulled in only by targets that need
-# it (`ios-open`); backend-only workflows stay fast.
+# it (`ios-open`); backend-only workflows stay fast. `--checkout` overrides the
+# `update = none` in .gitmodules (set so external clones skip this private
+# repo) so it still initializes on machines with access.
 ensure-ios-submodule:
 	@if [ ! -e "$(IOS_DIR)/.git" ]; then \
 		echo "[ensure-ios-submodule] initializing $(IOS_DIR)"; \
-		git submodule update --init --recursive "$(IOS_DIR)"; \
+		git submodule update --init --checkout --recursive "$(IOS_DIR)"; \
 	else \
 		echo "[ensure-ios-submodule] $(IOS_DIR) already initialized — leaving as-is"; \
 	fi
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -82,11 +82,21 @@ all: build
...
+# claiming success. Its leg passes `--checkout` to override `update = none`,
+# and fails soft (warning, not error) when the clone fails — e.g. no access to
+# the private repo.
+ensure-submodules: ## Initialize any missing submodules — intentd, FE, iOS (idempotent; iOS is fail-soft)
 	@for sm in $(SUBMODULES); do \
 		if [ ! -e "$$sm/.git" ]; then \
 			echo "[ensure-submodules] initializing $$sm"; \
-			git submodule update --init --recursive "$$sm" || exit 1; \
+			if [ "$$sm" = "$(IOS_DIR)" ]; then \
+				git submodule update --init --checkout --recursive "$$sm" \
```
panghy: **Non-blocking:** for an external user on an interactive terminal with no cached GitHub credentials, this `--checkout` clone of the private HTTPS repo will stop at `Username for 'https://github.com':` before it can fail — the fail-soft `|| echo` only kicks in after the prompt is dismissed (blank input → auth failure → warning). Since this leg is explicitly designed to fail soft, consider prefixing it with `GIT_TERMINAL_PROMPT=0` so it fails fast non-interactively:

```make
GIT_TERMINAL_PROMPT=0 git submodule update --init --checkout --recursive "$$sm" \
```

Internal devs who want to enter credentials interactively still have `make ensure-ios-submodule` (which intentionally hard-fails and can keep prompting).

.gitmodules
```diff
@@ -10,3 +10,4 @@
 	path = packages/ios
 	url = https://github.com/intent-hq/ios.git
 	branch = main
+	update = none
```
panghy, line 13: **Non-blocking, for the documented follow-up:** `git submodule init` copies `update = none` into the clone's local `.git/config` (verified on this branch: after `--init` the local `submodule.packages/ios.update` is `none`), and `git submodule sync` does **not** refresh the `update` key. So when iOS goes public and this line is deleted from `.gitmodules`, any clone that registered the submodule during this period will keep skipping it until `git config --unset submodule.packages/ios.update` is run (or `--checkout` is passed). Worth capturing in the follow-up note on #418 / the spec so the un-flip includes that step.

panghy (commented):
## Review: ✅ Approve (submitted as COMMENT — GitHub rejects self-approval since this account authored the PR)

Verified against the spec's acceptance criteria; the mechanism and shell logic are correct. Two non-blocking suggestions inline.

### What I checked

**`.gitmodules` / skip mechanism (empirically re-verified on this branch via a fresh `file://` clone):**
- `git submodule update --init --recursive packages/ios` → registers the submodule, prints `Skipping submodule 'packages/ios'`, **exit 0**, `packages/ios` left empty. Since a skipped submodule is never fetched, no credential prompt / 404 can occur for external users regardless of access.
- `--checkout` override re-verified with a public stand-in: with `update = none` set in local config, plain `--init` skips; `--init --checkout` initializes successfully. So `ensure-ios-submodule` and the `ensure-submodules` iOS leg work for internal devs.

**Makefile shell logic (static review):**
- `ensure-submodules`: quoting is sound; iOS leg fail-softs via `|| echo` (exit 0) while intentd/FE legs keep `|| exit 1`; loop exit status is correct in all orderings.
- `update`: the `update = none` skip guard is correctly placed *before* the not-initialized hard error and only fires when `.git` is absent, so an internal dev with ios initialized still gets the pull leg; `git config -f .gitmodules --get "submodule.$sm.update"` resolves correctly (section names match paths); command substitution inside `[ ... ]` doesn't trip `set -e`; the summary loop and the final `git diff --quiet -- $(SUBMODULES)` both tolerate an uninitialized ios (git treats uninitialized submodules as unchanged).
- The `echo "[update] $sm → $branch"` move below the guard avoids announcing a pull that then gets skipped — good.
- `ios-open` (via `ensure-ios-submodule`) hard-fails for users without access, which is appropriate: the target is unusable without the checkout anyway.

**Docs:** README/CONTRIBUTING/AGENTS are mutually consistent (private, `update = none`, skipped by default, `make ensure-ios-submodule` to opt in). README's simplified `git clone --recurse-submodules` flow works because `make check/test/build` only require intentd, which the clone now brings in.

**Acceptance criteria:** all six spec criteria map to the diff and pass (recursive clone exits 0 with empty `packages/ios`; `--init --recursive` in the clone exits 0; `ensure-ios-submodule` still force-inits; `ensure-submodules`/`update` tolerate missing ios; docs consistent; PR body carries `Fixes intent-hq/monorepo#418`).

Makefile
panghy: Addressed in f9ab943: the ensure-submodules iOS leg is now prefixed with GIT_TERMINAL_PROMPT=0, so with no cached credentials it fails fast ('could not read Username ... terminal prompts disabled') and hits the fail-soft warning instead of prompting. Verified in a temp clone with credential helpers disabled: the warning prints and make exits 0 with no prompt. ensure-ios-submodule is left prompting on purpose so internal devs can authenticate interactively there.

.gitmodules
panghy: Addressed in f9ab943: documented in the AGENTS.md repository-structure section (the single durable place describing the update = none posture) — when iOS goes public, deleting the .gitmodules line is not enough for clones registered during the private period; they also need 'git config --unset submodule.packages/ios.update', since 'git submodule init' copies the key into local .git/config and 'git submodule sync' does not refresh it.

PR #1966, opened 2026-08-11
chore: update cloudlands-fe submodule to latest main

panghy:
Bumps the `packages/cloudlands-fe` submodule to latest main (`166bbd7e7`), picking up intent-hq/cloudlands-fe#1023 (fix: fall back to top-level transport when building first stats).


panghy:
Closing: superseded — main's packages/cloudlands-fe gitlink (via #1967/#1970/#1972) already points past 166bbd7e7 (cloudlands-fe#1023), so this bump is a no-op that now conflicts on the gitlink.

PR #2072, opened 2026-08-12
fix: probe cargo subcommands instead of command -v in Makefile guards

### LINKED ISSUE #2066: make test refuses to run when cargo-nextest is installed but ~/.cargo/bin is off the non-interactive PATH
## Problem

`make test` aborts with

```
[test-intentd] ERROR: cargo-nextest is not installed — run 'cargo install cargo-nextest --locked'
```

on machines where cargo-nextest **is** installed and working. The guard in the root `Makefile` (`test-intentd`, ~L236) probes the binary by name:

```make
@command -v cargo-nextest >/dev/null 2>&1 || { \
	echo "[test-intentd] ERROR: cargo-nextest is not installed — run 'cargo install cargo-nextest --locked'"; \
	exit 1; \
}
```

That only succeeds when `~/.cargo/bin` is on `PATH`. It commonly is not in non-interactive contexts — agent shells, CI steps that do not source the cargo env, editors, `make` invoked from a GUI-launched process — because `~/.cargo/env` is sourced from an interactive shell profile.

## Why the guard is a false negative

Cargo resolves subcommands from `$CARGO_HOME/bin` **regardless of `PATH`**, so the command the guard is protecting works fine even when the guard fails. Verified on a machine where the guard reports "not installed": `which cargo` resolves to `/opt/homebrew/bin/cargo`, `~/.cargo/bin` is absent from the inherited `PATH`, `command -v cargo-nextest` fails — and `cargo nextest --version` prints `cargo-nextest 0.9.143 (60fa45f63 2026-08-04)`.

So the guard blocks a working `make test` outright. It is not a warning — it `exit 1`s.

## Impact

`make test` is the documented local test command in `AGENTS.md`. Agents and contributors hitting this either skip the suite or silently substitute plain `cargo test`, which does **not** apply `.config/nextest.toml`'s timing-serial test group, retries, and slow-timeout — exactly the protections that keep the flaky provider-binary modules from producing spurious failures. Several recent agent runs reported substituting `cargo test` for this reason and then had to baseline-diff ~24 unrelated failures by hand.

## Suggested fix

Probe the capability rather than the binary name:

```make
@cargo nextest --version >/dev/null 2>&1 || { \
	echo "[test-intentd] ERROR: cargo-nextest is not installed — run 'cargo install cargo-nextest --locked'"; \
	exit 1; \
}
```

Optionally also add `$(HOME)/.cargo/bin` to `PATH` for the recipe so a direct `cargo-nextest` invocation would work too.

Same consideration applies to the two `command -v cargo-sweep` guards (~L252, ~L263).

panghy:
The three tool guards in the root Makefile (`test-intentd`, `sweep`, `sweep-all`) probed the binary name with `command -v cargo-nextest` / `command -v cargo-sweep`, which false-negatives on working setups: cargo resolves external subcommands from `$CARGO_HOME/bin` regardless of `PATH`, so the guard could block a run that would have succeeded.

This changes the guards to probe the capability instead:

- `test-intentd`: `command -v cargo-nextest` → `cargo nextest --version >/dev/null 2>&1`
- `sweep`: `command -v cargo-sweep` → `cargo sweep --version >/dev/null 2>&1`
- `sweep-all`: same as `sweep`

Error messages and `exit 1` semantics are unchanged. No `PATH` mutation in recipes.

## Verification

- `cargo sweep --version` confirmed supported (cargo-sweep 0.8.0 prints `cargo-sweep-sweep 0.8.0`, exit 0).
- Positive, PATH-stripped (`PATH="/usr/bin:/bin:/tmp/cargo-only"` where `/tmp/cargo-only` holds only a `cargo` symlink, so neither `cargo-nextest` nor `cargo-sweep` is on `PATH`):
  - `command -v cargo-nextest` / `command -v cargo-sweep` exit 127 (old guards would have tripped);
  - `cargo nextest --version` / `cargo sweep --version` exit 0 (new probes pass);
  - `make sweep` passes the guard (no-ops: no `target/` dir);
  - `make test-intentd` passes the guard and reaches `cargo nextest run --workspace`.
- Negative: `cargo definitely-not-a-subcommand --version` exits 101 (non-zero), so the guard still trips when the tool is absent.
- Full `make test` green: 6266 tests run, 6266 passed, 1 skipped.

Fixes intent-hq/monorepo#2066

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,7 +233,7 @@ test: test-intentd ## Run the Rust test suite (cargo nextest; needs cargo-nextes
 # retries, slow-timeout). nextest does not run doctests; the workspace has
 # none, so nothing is lost.
 test-intentd: ensure-intentd-submodule
-	@command -v cargo-nextest >/dev/null 2>&1 || { \
+	@cargo nextest --version >/dev/null 2>&1 || { \
 		echo "[test-intentd] ERROR: cargo-nextest is not installed — run 'cargo install cargo-nextest --locked'"; \
 		exit 1; \
 	}
@@ -249,7 +249,7 @@ clean-dev: ## Wipe the dev-seat state dir (.dev/)
 # initializing the submodule just to sweep a target/ dir that cannot exist yet
 # would be overkill, so it short-circuits with a friendly no-op instead.
 sweep: ## Prune intentd build artifacts older than $(SWEEP_DAYS) days (needs cargo-sweep)
-	@command -v cargo-sweep >/dev/null 2>&1 || { \
+	@cargo sweep --version >/dev/null 2>&1 || { \
 		echo "[sweep] ERROR: cargo-sweep is not installed — run 'cargo install cargo-sweep'"; \
 		exit 1; \
 	}
@@ -260,7 +260,7 @@ sweep: ## Prune intentd build artifacts older than $(SWEEP_DAYS) days (needs car
 	fi
 
 sweep-all: ## Sweep intentd build artifacts in every worktree under $(WORKSPACES_DIR)
-	@command -v cargo-sweep >/dev/null 2>&1 || { \
+	@cargo sweep --version >/dev/null 2>&1 || { \
 		echo "[sweep-all] ERROR: cargo-sweep is not installed — run 'cargo install cargo-sweep'"; \
 		exit 1; \
 	}


panghy (commented):
## Verification review (verifier agent) — ✅ approved

Reviewed the diff and independently re-ran the verification for intent-hq/monorepo#2066 at commit 8eb0854.

**Diff review**
- Only the three root-Makefile guards changed (`test-intentd`, `sweep`, `sweep-all`): `command -v cargo-<tool>` → `cargo <tool> --version >/dev/null 2>&1`.
- Error messages and `exit 1` semantics unchanged; no submodule pin changes; no `PATH` mutation in recipes.

**Independent re-verification** (workspace at 8eb0854, `PATH="/usr/bin:/bin:/tmp/verify-cargo-only"` where the extra dir holds only a `cargo` symlink):
- Old guards would trip: `command -v cargo-nextest` / `command -v cargo-sweep` → exit 127.
- New probes pass: `cargo nextest --version` / `cargo sweep --version` → exit 0.
- `make sweep` on the stripped PATH passes the guard and runs cargo-sweep (exit 0).
- `make test-intentd` on the stripped PATH passes the guard and reaches `cargo nextest run --workspace` (tests started; run cut off by timeout as intended).
- Negative: `cargo definitely-not-a-subcommand --version` → exit 101, so a genuinely missing tool still trips the guard.
- Full `make test`: 6266 tests run, 6266 passed (1 flaky passed on retry), 1 skipped — exit 0.

**PR hygiene**: conventional title, `Fixes intent-hq/monorepo#2066` in body, single conventional commit, `pr-title` check green, bot review left no suggestions.

Everything checks out; no changes requested.

Issue #2129, opened 2026-08-12
intentd: note.create drops @@@task conversion warnings and createdTasks — silent relation loss on new notes


panghy:
## What happened

Dogfooding #2018 on .27: created a NEW note via MCP `create_note` whose initial content contained 8 `@@@task` blocks, including deliberate warning probes (`dependsOn=nosuchkey`, unknown attribute `mysteryattr=42`, and a cycle g↔h whose closing edge must be rejected).

- Auto-conversion at note creation worked: all 8 blocks converted, relations/efforts seeded correctly, bad edges correctly dropped.
- But the `note.create` result carried NO `createdTasks` / `warnings` fields — so the cycle-rejection and unresolvable-reference warnings were never observable by the caller.
- Re-running `task.convertBlocks` on the note is idempotent (`convertedCount: 0`, empty `warnings`), so the warnings from the original conversion are gone for good.

## Why it matters

Per #2018's design, warnings are the only signal that an author's intended relation was dropped (typo'd key, cycle). The four content-write ops (`setContent`/`add`/`edit`/`editLines`) carry `createdTasks` + `warnings`, but note CREATION with task blocks in the initial content silently discards them. An agent creating a planning note with a typo'd `dependsOn` gets a silently-degraded graph and no way to notice.

## Proposal

Thread the conversion result (`createdTasks` + `warnings`) through `note.create`'s result the same way the content-write ops do.

Observed on .27 with intentd containing #1128/#1130/#1133.

panghy:
Fixed in intentd v0.6.12 (alpha).

Issue #2128, opened 2026-08-12
intentd: unlockPlan omits criticalPathMinutes when any task in the graph lacks an effort estimate


panghy:
## What happened

Dogfooding the shipped #1974/#2018/#2044 stack on .27 with an 8-task graph (diamond A,B→D→E plus C(dep A, conflict B) and warning/cycle probes F/G/H):

- Wave 1 batch delegate: started A,B,F,H; held C,D,E,G. Efforts present on A(5m), B(10m), C(5m), D(15m), E(5m); F/G/H have none. `unlockPlan` named the settlement-unlocked set correctly but carried **no** `criticalPathMinutes`.
- Wave 2 (C,D,E,G): same — correct `unlockedBySettlement`, no `criticalPathMinutes`.
- Wave 3 (E alone, effort 5m): `criticalPathMinutes: 5` present ("~5 min of serial work remains on the critical path").

## Expected

The delegate docs say estimates default to 30 min when missing/unparseable, so a mixed graph (some tasks with estimates, some without) should still produce a critical-path projection — wave 1's B(10m)→D(15m)→E(5m) chain should have reported ~30 min.

## Suspicion

`criticalPathMinutes` appears to be emitted only when EVERY task involved carries a parseable estimate, instead of defaulting the missing ones to 30 min as the admission-ordering path does. If the omission is intentional (avoid misleading numbers when estimates are partial), the docs should say so; otherwise the plan should apply the same 30-min default and always report.

Observed on .27 with intentd containing #1138/#1144/#1147.

panghy:
Fixed in intentd v0.6.11 (beta).

Issue #1162, opened 2026-07-30
intentd: delegated agent session dies repeatedly with 'JSON-RPC error 0: agent stdout closed'; wakeOrCreate re-attaches to the dead session instead of recreating


panghy:
A delegated implementor agent (fable-5, background, delegationDepth 1) crashed twice in a row with:

```
internal error: session/prompt failed: JSON-RPC error 0: agent stdout closed
```

Timeline:
1. First crash ~4 minutes into the session, mid tool-use (reading files; no edits yet). Coordinator received a child-failed event.
2. Coordinator used `agent.wakeOrCreate(taskNoteId, ...)` which returned `{ action: "woke_existing" }` for the same agent id.
3. The woken session appended the context message (messageCount 2→3) but then crashed the same way again within ~30s; the delegation group settled with completionStatus: partial. Notably `agent.status` afterwards showed `status: "active"` with no stopReason, while the parent had received a failure event for it — status and event stream disagree.

Impact: the task loses its session repeatedly; wake-based recovery did not stick. Recovery required spawning a brand-new agent on the same task note.

Questions/asks:
- What closes the child agent's stdout? (provider/ACP process exit? daemon-side pipe teardown? resource limits?) Two crashes in the same workspace within minutes while 3+ sibling agents ran unaffected suggests per-session state, not global load.
- `wakeOrCreate` reporting woke_existing but the session dying again immediately suggests the underlying ACP session was already unhealthy; wake should probably health-check the session (or recreate) instead of re-attaching to a dead pipe.
- Status/event divergence: after the second crash, `agent.status` = active with no stopReason while the delegation group had already recorded the failure.

Context: same dogfood workspace as #1148/#1150 (daemon possibly restarted under make dev earlier in the session, though the first crash occurred well after).

panghy:
Follow-up detail for this issue from the same workspace, ~10 minutes later: the "crashed" session was NOT actually dead.

Sequence:
1. Parent received TWO child-failed events for agent-60c336c5 ("JSON-RPC error 0: agent stdout closed"), the second after a wakeOrCreate woke_existing; the delegation group settled with completionStatus: partial.
2. Parent spawned a replacement agent on the same task note.
3. The replacement discovered the original session was alive and actively editing its worktree (uncommitted changes appearing minutes after the second "failure"), and stood down without edits.

[...]

Issue #1160, opened 2026-07-30
cloudlands-fe: stale stream-owned ghost message survives hydrates forever after daemon crash mid-turn


panghy:
## Problem

`withPreservedStreamOwnedMessages` in `src/features/agent/chat-read-service.ts` (added in cloudlands-fe#565 to fix the mid-turn re-entry clobber) has no eviction path for a **stale** stream-owned row.

If the daemon dies mid-turn without persisting the partial (crash/restart), the renderer store keeps a message with `isStreaming: true` forever:

- every future hydrate preserves it (its id is never in the fetched pages), and
- `applyTranscript` in `chat-subscribe-service` also preserves store-only rows,

so the ghost survives reconnects until app restart. Before #565, the wholesale replace in `loadChatTranscript` accidentally healed this case.

The module comment says orphan healing belongs in the daemon, but this particular ghost is renderer-local state the daemon has no knowledge of, so it cannot heal it.

## Suggested fix

The fresh `session` object is available at the `withPreservedStreamOwnedMessages` call site — gate preservation on the daemon actually reporting a turn in flight, and drop stream-owned rows when it doesn't. A wrongly-dropped live partial self-heals on the next delta (deltas carry full accumulated blocks); a wrongly-kept ghost never recovers.

## Origin

Raised as a non-blocking review comment on cloudlands-fe#565 (https://github.com/intent-hq/cloudlands-fe/pull/565#discussion_r3684753192).

panghy:
Fixed in cloudlands-fe v2.12.0 (beta), bundles intentd v0.2.16.

panghy:
cloudlands-fe v2.12.0 promoted to stable, bundles intentd v0.2.16.

PR #2141, opened 2026-08-12
docs: document note.create conversion fields and criticalPathMinutes mixed-estimate emission in PROTOCOL.md

panghy:
Documents two shipped intentd wire changes in the canonical contract (both "within v6.14" — no protocol-constant bump; the daemon's `PROTOCOL_VERSION` on main is `6.14`):

## `note.create` conversion fields (refs #2129; intent-hq/intentd#1162)

- §5.2 method table: `note.create` result documented as `{ note, convertedCount, createdTaskNoteIds, createdTasks, warnings }` — additive over the old `{ note }` shape, same shapes/warning contract as the four content-write ops, all four fields always present. Also documents the `idempotencyKey?` param and the legacy-replay decode (a key recorded before the fields existed decodes as a zeroed conversion outcome).
- §5.2 "`@@@task` auto-conversion on note writes" paragraph now lists `note.create` (initial content) among the converting writes, with a note that earlier daemons converted but returned only `{ note }`.

## `criticalPathMinutes` mixed-estimate emission (refs #2128; intent-hq/intentd#1160)

- §5.5 batch `agent.delegate` row: the `unlockPlan.criticalPathMinutes` emission condition updated — the max is taken over requested tasks whose max-attaining chain carries at least one parsed estimate, so a longer pure-defaults chain no longer suppresses an estimated chain's minutes; pure-defaults-only graphs still omit the field.

## Version history

- Two "Also within 6.14" entries appended to the §Protocol Version & Compatibility narrative, following the established convention for no-bump behavior/additive-field changes.

Docs only; no code. Both intentd PRs are merged (intentd-first rule satisfied).


panghy (commented):
**Verifier review: ✅ APPROVED** (posted as a comment — GitHub blocks a formal approval from the PR author's own account). Verified against the merged intentd diffs (origin/main @ 654b34f4); no findings.

**(a) Field names & semantics match the merged code.**
- intentd#1162 (654b34f4): `NoteCreateResult` serializes `{ note, convertedCount, createdTaskNoteIds, createdTasks, warnings }` (camelCase, no `skip_serializing_if` → all four fields always present; `convertedCount: 0` + empty arrays on no-op conversion). The router `note.create` arm returns `to_result_value(&result)` over the old `{ note }`, the MCP `ws.note.create` binding appends the same four fields, `note` is the refetched post-conversion row (`outcome.refetched_note` replaces it before the result is built), and the custom `Deserialize` decodes a legacy bare-`Note` idempotency record as a zeroed conversion outcome (covered by `note_create_result_decodes_legacy_bare_note_idempotency_record`). The `idempotencyKey?` param is pre-existing on both the router and MCP paths, so documenting it in the table row is accurate.
- intentd#1160 (4558c57a): `serial_remaining_minutes` now filters requested tasks' `PathDetail`s to estimated ones before taking the max (`.filter(|(_, estimated)| *estimated).map(...).max()?`), and `critical_path_details` propagates the estimate flag along max-attaining chains only — exactly the doc's "the max is taken over requested tasks whose max-attaining chain carries at least one parsed estimate". Pure-defaults-only graphs still return `None` (field omitted), a longer unestimated chain neither suppresses nor inflates, and understatement is possible — all confirmed by the updated regression test (`serial_remaining_reports_estimated_chain_despite_longer_defaults_chain`, 30 vs a 60-min defaults chain).

**(b) "within v6.14" anchoring is correct.** `PROTOCOL_VERSION` on intentd main is `6.14` (`crates/intent-transport/src/protocol.rs:133`); `protocol.rs` was last touched by #1146 — neither #1160 nor #1162 modified it. The two new entries append immediately after the "Version 6.14 is an **additive** minor bump over 6.13 … 287 router methods, 326 total." entry, so the placement is right.

**(c) Internally consistent with PROTOCOL.md conventions.** Both entries follow the established "Also within X (…, presence-detected per the convention below; no method-catalog change, so no version bump — refs)" style; leaving the historical "Also within 6.8" `criticalPathMinutes` entry untouched while the new 6.14 entry names the previous behavior matches the doc's supersession convention (cf. the v4.3/v2.8 dismiss-notice precedent). The §5.2 table row, the auto-conversion paragraph ("`note.create` has always converted, but its result carries the outcome only since v6.14 — earlier daemons converted and returned only `{ note }`, silently discarding the warnings"), and the §5.5 row all agree with each other and with the version-history entries; no other spot in the doc still asserts the bare `{ note }` result.

**(d) Scope is clean.** Diff is `docs/PROTOCOL.md` only, +12/−9; the remaining textual churn is line re-wrapping within the edited paragraphs. `pr-title` check passes.

Issue #2149, opened 2026-08-12
docs: PROTOCOL.md §5.7 describes childMemoryPeakBytes as a max over childMemoryBytes, and omits what it is blind to


panghy:
Follow-up to #2107, fixed by [intentd#1167](https://github.com/intent-hq/intentd/pull/1167) (merged). The daemon behaviour changed; `docs/PROTOCOL.md` §5.7 has not been updated to match, and its current wording is now imprecise in a way that invites the same wrong conclusion #2107 was about.

**No version bump.** Sampling behaviour changed, not the wire shape — no new field, no shape change. This is a doc-only amendment.

## What §5.7 says today (v6.14, line ~284)

> `childMemoryPeakBytes` is the high-water mark of `childMemoryBytes` since daemon start. It is carried separately because ephemeral quick-action (`agent.completeOnce`, `agent.enhancePrompt`) and model-probe adapter chains live only seconds, so by the time a debug bundle is captured the instantaneous value has drained back to baseline.

Two things are now wrong or missing:

1. **"high-water mark of `childMemoryBytes`"** — the peak is no longer a maximum over published `childMemoryBytes` values. It is sampled every 500 ms while an ephemeral adapter chain is live, against the 5 s cadence `childProcesses` / `childMemoryBytes` keep, so it can and routinely does exceed every value `childMemoryBytes` ever publishes. (Measured: a 16-chain burst peaked at 6.97 GB with `childMemoryBytes` reading 0 at every published sample.)

2. **It does not say what the peak is still blind to.** The fast cadence keys off held permits in the daemon-wide adapter bound (`agents.maxConcurrentAdapters`), so it covers exactly the two paths that go through `spawn_adapter` — the one-shot ACP runner and the model probe — and nothing else. Not covered: the **auggie route of the same quick actions** (spawns its CLI directly, takes no slot), `host.exec` children, PTY sessions, MCP bridge servers, the Unsloth server, and the tool children a long-lived agent runs. Those are mostly long-lived enough for the 5 s baseline to catch — an agent subtree lives minutes to hours — but a short, sharp excursion from one of them can still be missed the way an adapter burst used to be.

Leaving (2) unstated is precisely the shape of #2107: a reader who assumes the peak is exact would draw a confident wrong conclusion from a field that is exact only for part of the tree.

## Proposed replacement

[...]

PR #2220, opened 2026-08-13
docs(protocol): describe childMemoryPeakBytes as an independently-sampled peak with bounded coverage

### LINKED ISSUE #2107: intentd: childMemoryPeakBytes aliases on transients — understated a burst peak 2x and made an unbounded run look cheaper than a bounded one
Found while validating the bounds shipped in intentd#1146 / #1145 (see #2062, #2063).

## Problem

`childMemoryPeakBytes` was added in intentd#1139 specifically so that a debug bundle captured *after* the fact still shows a transient overshoot — the instantaneous `childMemoryBytes` drains back to baseline within seconds, so without a high-water mark a bundle reports nothing. From that PR's own rationale:

> ephemeral quick-action and model-probe adapter chains live only seconds, so by the time a debug bundle is captured the instantaneous value has drained back to baseline.

**It does not reliably do this.** The high-water mark is computed over 5 s samples, so a burst that peaks between ticks is never seen, and the "peak" is only the maximum of what the ticks happened to catch.

## Measurement

Validating the one-shot adapter bound, an unbounded 16-call burst (`cap64-n16-a`) was measured two ways at once:

| Source | Peak reported |
|---|---:|
| `system.status.childMemoryPeakBytes` (5 s sampling) | **3.48 GB** |
| `ps` descendant walk at 1 Hz | **7.00 GB** |

The telemetry understated the real peak by **~2x**. Worse, the same run *bounded* (cap 6) measured **3.57 GB** by `ps` — so **read from telemetry alone, the unbounded run looks cheaper than the bounded one.** Anyone using this field to judge whether a bound is working would reach the opposite of the correct conclusion.

Steady state is fine: in a running soak both `system.status` and `ps` read 5.85 GB. The field only aliases on transients — which is precisely the case it was introduced for.

## Why it matters

Every conclusion in the #2062/#2063 validation had to be taken from a 1 Hz `ps` walk rather than from `system.status`. That option is not available to a user submitting a debug bundle, which is the intended consumer.

## Suggested direction

Options, roughly in order of cost:

1. **Sample on the event, not the clock** — take a tree reading when an adapter chain is spawned/reaped, so bursts are observed by construction rather than by luck.
2. **Adaptive cadence** — drop to sub-second sampling while `childProcesses` is non-zero and rising, back off when flat.
3. **Document the limitation honestly** — if the cadence stays at 5 s, `childMemoryPeakBytes` should be described as "peak of sampled values, may miss sub-5s transients" in `docs/PROTOCOL.md` (§5.7, currently v6.14), so nobody draws the inverted conclusion above.

(3) is the minimum; on its own it makes the field close to useless for its stated purpose, so it is worth pairing with (1) or (2).

### LINKED ISSUE #2150: docs(protocol): §5.7 childMemoryPeakBytes — peak can exceed every published childMemoryBytes, and its fast cadence only covers slot-holding adapters
Follow-up from intentd#1167 (fixing #2107), flagged by its author rather than left in a PR body.

## 1. §5.7's description of `childMemoryPeakBytes` is now imprecise

The doc describes it as the "high-water mark of `childMemoryBytes`". After #1167 that is no longer strictly true: the peak is sampled on a faster cadence (500 ms while an adapter chain holds a slot in the daemon-wide bound) than `childMemoryBytes` is published at (5 s), so **the peak can legitimately exceed every value `childMemoryBytes` ever reports**. A reader reconciling the two fields in a bundle would currently conclude one of them is wrong.

## 2. The peak's coverage is bounded, and that should be stated

The fast cadence keys off held permits in the adapter concurrency bound, so it observes:

- the one-shot ACP runner, and
- the model probe.

It does **not** observe, and a transient from any of these can still be missed:

- the **auggie** route of the same quick actions — it spawns its CLI directly and takes no slot,
- `host.exec` children,
- PTY sessions,
- MCP bridges,
- the Unsloth server,
- an agent's own tool children.

This is written down at `CHILD_TREE_BURST_PERIOD` in the intentd source, but a client reading `docs/PROTOCOL.md` §5.7 has no way to know it. Given that #2107 was *caused* by an unstated sampling caveat producing an inverted conclusion, leaving this one unstated repeats the mistake in a smaller way.

## Suggested fix

Amend §5.7 to describe `childMemoryPeakBytes` as the high-water mark of the daemon's **sampled** child-tree memory — sampled faster than `childMemoryBytes` is published, hence potentially exceeding it — and list what the fast cadence covers and does not.

No version bump: #1167 changed sampling behaviour, not the wire shape, and the fields themselves have been documented since 6.13.

panghy:
Fixes intent-hq/monorepo#2150

Doc-only precision fix for PROTOCOL §5.7's `system.status` child-process tree fields, following up [intent-hq/intentd#1167](https://github.com/intent-hq/intentd/pull/1167) (which fixed #2107 by adding the 500 ms burst sampling cadence but left §5.7 describing the pre-#1167 contract).

## What changed

1. **`childMemoryPeakBytes` is no longer described as "the high-water mark of `childMemoryBytes`"** — it is now described as the high-water mark of the daemon's **sampled** descendant-tree memory: swept every 500 ms while an ephemeral ACP adapter chain holds a slot in the daemon-wide adapter bound (one-shot `agent.completeOnce` / `agent.enhancePrompt` on ACP providers, and model probes), against the 5 s cadence `childProcesses` / `childMemoryBytes` are published at — so the peak can legitimately exceed every instantaneous value ever published. Measured backing (from the #1167 work): a 16-chain burst peaked at 6.97 GB while `childMemoryBytes` read 0 at every published sample.
2. **The fast cadence's bounded coverage is now stated**: it applies only to descendants spawned through the ephemeral-adapter bound (the one-shot ACP runner and the model probe). The auggie route of the same quick actions, `host.exec` children, PTY sessions, MCP bridge servers, the Unsloth server, and an agent's own tool children remain baseline-sampled (5 s) only, so a short, sharp burst from one of those may not appear in the peak.

Wording follows the exact replacement paragraph proposed on the issue (carried over from #2149), adapted to the section's bullet style, and matches the `CHILD_TREE_BURST_PERIOD` source comment in intentd.

No version bump: intentd#1167 changed sampling behaviour, not the wire shape; the fields have been documented since 6.13.


docs/PROTOCOL.md
```diff
@@ -274,15 +274,16 @@ The `system.status` result additionally reports the daemon's **whole descendant
...
+- `childMemoryPeakBytes` is the high-water mark of the daemon's **sampled** descendant-tree memory since daemon start. It is **not** simply the maximum of the `childMemoryBytes` values a client has seen: the tree is swept every **500 ms** while an ephemeral ACP adapter chain holds a slot in the daemon-wide adapter bound (`agents.maxConcurrentAdapters`, §5.12) — one-shot `agent.completeOnce` / `agent.enhancePrompt` on ACP providers, and model probes — against the 5 s cadence `childProcesses` / `childMemoryBytes` are published at, so **the peak can legitimately exceed every instantaneous value ever published**. That is deliberate: those chains live only seconds, so by the time a debug bundle is captured the instantaneous value has drained back to baseline (intentd#1139, monorepo#2107; fast cadence introduced by [intent-hq/intentd#1167](https://github.com/intent-hq/intentd/pull/1167) — measured, a 16-chain burst peaked at 6.97 GB while `childMemoryBytes` read 0 at every published sample).
```
panghy: **Accuracy:** `agent.enhancePrompt` should not be in this slot-holder list. It is auggie-only — §5.31 gates it on auggie being the effective default provider (any other provider returns `{ available: false }`), §5.32 explicitly contrasts it ("Unlike `agent.enhancePrompt` (§5.31, auggie-only)"), and in the source `enhance_ops.rs` never touches the adapter bound while the one-shot ACP runner (`one_shot_acp.rs`) is consumed only by `complete_ops.rs`. So "`agent.enhancePrompt` on ACP providers" names a route that does not exist, and implies the fast cadence can cover it. The `CHILD_TREE_BURST_PERIOD` comment names only "the one-shot ACP runner and the model probe". (The wording was carried over from the #2149 paragraph on the issue, but that paragraph is imprecise on this point.)

Suggested: `— one-shot \`agent.completeOnce\` on ACP providers, and model probes —`. The next bullet's "the **auggie** route of the same quick actions" can then stay as-is, since for `agent.enhancePrompt` the auggie route is the only route and it is correctly described there as taking no slot.

docs/PROTOCOL.md
```diff
@@ -274,15 +274,16 @@ The `system.status` result additionally reports the daemon's **whole descendant
...
+- The fast cadence applies **only** to descendants spawned through the ephemeral-adapter bound — the one-shot ACP runner and the model probe. Everything else in the tree — the **auggie** route of the same quick actions (it spawns its CLI directly and takes no slot), `host.exec` children, PTY sessions, MCP bridge servers, the Unsloth server, and the tool children a long-lived agent runs — is sampled at the 5 s baseline only, so a burst from one of those that spikes and drains inside one baseline interval may not appear in the peak.
```
panghy, line 285: **Version-history convention (non-blocking but worth doing here):** the doc consistently records behavior-only changes as "Also within 6.x (behavior only, no method-catalog or wire-shape change, so no version bump — intentd#NNNN, monorepo#NNNN)" notes in the version-history paragraph, and intentd#1167's burst cadence is exactly that class of change — yet `grep 1167` finds it only in this section. As it stands, the v6.13 entry still defines `childMemoryPeakBytes` as "the high-water mark of `childMemoryBytes` since daemon start" with nothing in the history recording that #1167 changed the semantics, so a reader of the version history alone is contradicted by §5.7 with no bridge. Suggest appending an "Also within 6.14 (behavior only, no wire change — intentd#1167, monorepo#2107)" note summarizing the burst cadence and noting it supersedes the v6.13 description (precedent: the v4.3 note that supersedes the pre-#892 dismissal contract "documented since v2.8").

panghy (commented):
Reviewed against issue #2150 (body + the #2149 replacement paragraph carried in its comments), the `CHILD_TREE_BURST_PERIOD` doc comment in `packages/intentd/crates/intentd/src/main.rs`, and the surrounding contract (§5.12, §5.31, §5.32, v6.13/v6.14 version-history entries).

**Verified accurate:**
- The 500 ms burst / 5 s baseline cadences and the trigger condition (sweep while an ephemeral chain holds a slot in the adapter bound) match `CHILD_TREE_BURST_PERIOD` / `CHILD_TREE_BASE_PERIOD` and `child_tree_sweep` exactly.
- The bounded-coverage bullet (auggie route spawns its CLI directly and takes no slot; `host.exec` children, PTY sessions, MCP bridge servers, the Unsloth server, and a long-lived agent's tool children stay baseline-sampled) matches blind-spot #2 of the source comment nearly verbatim.
- The measured backing (16-chain burst peaking at 6.97 GB while `childMemoryBytes` read 0 at every published sample) matches the #2149 paragraph carried on the issue.
- The `agents.maxConcurrentAdapters` → §5.12 cross-reference is correct, and the intentd#1139 / monorepo#2107 / intentd#1167 references are all accurate.
- Markdown renders cleanly (the long bullets are single lines; bold/code spans/link are well-formed), and "no version bump" is right — behavior only, wire shape unchanged.

Two issues inline: one accuracy fix (`agent.enhancePrompt` listed as an ACP slot-holder) and one version-history convention gap.

docs/PROTOCOL.md
panghy: Fixed in ea359de — dropped `agent.enhancePrompt` from the slot-holder list exactly as suggested (`— one-shot \`agent.completeOnce\` on ACP providers, and model probes —`), and left the next bullet's auggie-route wording as-is. Verified against §5.31 (auggie-only gate), §5.32 ("Unlike `agent.enhancePrompt` (§5.31, auggie-only)"), and the source (`enhance_ops.rs` never touches the adapter bound; `one_shot_acp.rs` is consumed only by `complete_ops.rs`).

docs/PROTOCOL.md
panghy: Done in ea359de — appended an "Also within 6.14 (behavior only, no method-catalog or wire-shape change, so no version bump — intentd#1167, monorepo#2107)" note to the version-history paragraph, summarizing the 500 ms burst cadence vs the 5 s publish cadence and explicitly noting it supersedes the v6.13 description of `childMemoryPeakBytes` as "the high-water mark of `childMemoryBytes` since daemon start" (following the v4.3 supersession precedent).

PR #2244, opened 2026-08-13
docs: document workspace:setup lifecycle events and watcher deferral

panghy:
Documents the `workspace:setup` lifecycle events and the watcher-deferral behavior shipped in intent-hq/intentd#1183 (`feat: defer workspace watcher start until setup script completes`).

## Changes (docs/PROTOCOL.md only)

- **§5.1 (setup script execution):** new paragraph describing the setup-stage lifecycle events — `workspace:setup:started` fires iff an effective script was resolved and a spawn will be attempted; exactly one `workspace:setup:completed` fires per logical create on every terminal path (script ran to exit, spawn/pre-spawn failure, no effective script, `skipIsolation`/no worktree, `workspace.duplicate`). Also documents that a created workspace's watcher registration is deferred until the completion event is observed (60s backstop), so setup-script churn never surfaces as `file:*` events.
- **§6.5 (event taxonomy):** new `workspace setup` row documenting both event types, their self-sufficient payloads (`{ workspaceId }` / `{ workspaceId, ranScript, exitCode? }`), the exit-code omission semantics, idempotent-replay behavior, and the internal watcher-registry consumption.

No submodule pointer changes — pin advancement is owned by the auto-bump workflow.


docs/PROTOCOL.md
```diff
@@ -924,6 +924,20 @@ a newline-prefixed completion summary to the scrollback —
...
+`exitCode`), spawn/pre-spawn failure (`ranScript: false`), no effective script,
+`skipIsolation`/no worktree, and `workspace.duplicate` (immediate
+`completed { ranScript: false }`). Idempotent replays publish nothing (same as
+`workspace:created`). The daemon's own file watcher keys off the completion: a created
+workspace's watcher registration (file watcher, git-metadata watcher,
+skills/specialists) is **deferred** until `workspace:setup:completed` is observed for
+it, so setup-script churn is dropped — never published, never buffered (no `file:*`
+events exist for the setup window) — with a 60s backstop that starts the watchers
+anyway (WARN) if no completion is observed; a script outliving the backstop emits its
+remaining churn.
```
panghy: **Non-blocking accuracy nuance:** the 60s backstop is not the only path that ends the no-`file:*` window early. The watcher registry's `workspace:opened` arm **supersedes the deferral** — an open during the setup window removes the pending entry and starts the watchers immediately (`events/registry.rs`, `WORKSPACE_OPENED` arm: "An open during the setup window supersedes the deferral (the user is in the workspace; watch it now)"), so a still-running script's remaining churn does surface as `file:*` from that point on.

As written, "never published, never buffered (no `file:*` events exist for the setup window)" reads as absolute with the backstop as the sole exception. Suggest appending a short clause, e.g.: "…; a `workspace.open` during the setup window likewise supersedes the deferral and starts the watchers immediately." Same applies to the §6.5 row's "setup-window file churn never surfaces as `file:*`". Fine to land as-is and amend in a follow-up if preferred.

panghy (commented):
## Verification review (against merged intent-hq/intentd#1183)

**Verdict: accurate — no blocking issues.** Every claim in both edits was checked against the merged implementation (intentd main `80478fd8`, the #1183 squash; tree-identical to the feature branch tip).

### Scope check
- ✅ PR contains **only** `docs/PROTOCOL.md` (+15/−0). No submodule pointer changes.
- ✅ CI: `pr-title` passes; PR is mergeable.

### Accuracy — verified against code
- ✅ **`workspace:setup:started` fires iff an effective script was resolved and a spawn will be attempted** — `intent-services/src/lib.rs` publishes it only after the effective-script resolution (explicit param > repo config > legacy row) yields a script; the no-script arm publishes `completed { ranScript: false }` directly with no `started`.
- ✅ **Exactly one `completed` per logical create, every terminal path** — all enumerated paths have emit sites and unit coverage (`tests.rs` `setup_lifecycle_events` module): script exit (`ranScript: true` + `exitCode`, exit 0 and exit 7 both tested), pre-spawn failures (`.intent` symlink guard, dir-create/script-write/wrapper-write failures) and PTY spawn failure all `break 'setup (false, None)`, no-script, `skipIsolation`/no-worktree (else-arm at the `setup_worktree` check), and `workspace.duplicate` (immediate `completed { ranScript: false }` under the duplicate's id, no `started` — tested).
- ✅ **`exitCode` omitted, never `null`** — `workspace_setup_completed_event` inserts the key only when `Some`; type is `u32` (`intent-pty` `ExitStatus.exit_code: u32`); "present only when the script ran to exit and its status was observable" exactly matches `(true, exit.ok().map(|e| e.exit_code))` — a `wait()` error yields `ranScript: true` with `exitCode` omitted.
- ✅ **Idempotent replays publish neither event** — the whole setup stage lives inside the idempotency closure; a cached response returns without re-running it (matches the `workspace:created` analogy).
- ✅ **Watcher deferral** — `events/registry.rs`: `workspace:created` inserts a `PendingSetup`; `start_watches` + skills/specialists registration run only on the matching `workspace:setup:completed`; `SETUP_COMPLETION_BACKSTOP = 60s` starts them anyway with a `tracing::warn!`. "Never published, never buffered" and "a script outliving the backstop emits its remaining churn" both match the module docs and tests.
- ✅ **Actor: system; payload shapes** — `system_actor()`; WSS e2e (`e2e_wss_setup_lifecycle.rs`) asserts the exact `{ workspaceId }` / `{ workspaceId, ranScript, exitCode }` wire shapes and the no-`file:*`-before-completion invariant.

### Conventions
- ✅ §5.1 paragraph sits in the setup-script-execution block, before initial-agent orchestration; §6.5 row sits between the `workspace` and `spec/goal` rows; "(new in intentd)" family naming, "additive" wording, and "self-sufficient payloads (§6.7)" phrasing all match neighboring rows.

### One non-blocking nuance (inline comment below)
The absolute "setup-script churn never surfaces as `file:*`" claim has a second exception besides the 60s backstop: a `workspace:opened` during the setup window supersedes the deferral. See inline comment.

Also noted (informational, no change needed): the legacy-importer path (`publish_workspace_created`) pairs its `workspace:created` with an immediate `completed { ranScript: false }` too — out of §5.1 `workspace.create` scope, so its omission from the enumerated paths is fine.

docs/PROTOCOL.md
panghy: Addressed in 859f91d — appended the workspace.open-supersedes-deferral clause to the §5.1 paragraph and qualified the §6.5 row's "never surfaces as file:*" accordingly.

PR #2296, opened 2026-08-13
docs(protocol): document reason on agent:process:* events in 6.5

panghy:
## Summary

Documents the additive `reason` field on the `agent:process:queued` / `agent:process:resumed` / `agent:process:evicted` events in the §6.5 event taxonomy (monorepo#2063 A4). These events previously had no taxonomy row at all; the new "agent (process registry)" row documents the full payload `{ agentId, used, cap, reason }`, where `reason` is `"slots"` or `"memory-budget"` naming the admission constraint that drove the event.

Daemon side: [intent-hq/intentd#1196](https://github.com/intent-hq/intentd/pull/1196) (intentd-first ordering — this docs PR should merge after it).

Refs intent-hq/monorepo#2063


panghy:
**Review (verifier): docs row is accurate on the payload/semantics — two corrections needed.**

Checked the §6.5 row against the implementation in intent-hq/intentd#1196.

**Accurate as written**

- Payload shape `data { agentId, used, cap, reason }`, the two `reason` values and when each applies, the queued-waiter echo on `resumed`, the budget-wins-when-both-bind rule, and the best-effort skip when the agent's session row is missing — all match the implementation.

**Corrections**

1. **"Actor: system" is wrong.** The events are published via `publish_agent_event` with `agent_actor(agent_id)`, so the emitted frames carry `actorType: "agent"` with the affected agent's id (and the agent id as the session id). The row should read something like "Actor: agent (the affected agent's id)."
2. **Cross-reference "(§5.5 concurrency cap)" doesn't resolve.** §5.5 has no concurrency-cap text; the cap is the `agents.maxConcurrent` setting documented in §5.12. Suggest pointing the reference there (e.g. "the `agents.maxConcurrent` cap, §5.12").

panghy:
Both corrections applied in f1b6cfe: the row now reads "Actor: agent (the affected agent's id)" (publish_agent_event uses agent_actor), and the cap cross-reference points at agents.maxConcurrent in §5.12 instead of §5.5.

PR #2332, opened 2026-08-14
docs: document FE-side loopback-hostname interpretation for browser.exec

panghy:
Documents the FE-side loopback-hostname convention for `browser.exec` in `docs/PROTOCOL.md` (part of intent-hq/monorepo#2323; docs-only, no wire change).

## What

- **§5.9 (`browser.exec` contract block)** — new paragraph: URL hostnames in `navigate`/`openTab` action URLs are interpreted on the frontend that serves the reverse RPC, per the reserved-hostname convention:
  - `daemon.localhost` → the daemon machine (`127.0.0.1` locally, the sanitized transport-target host on a remote daemon)
  - `client.localhost` → the client (user's) machine (always `127.0.0.1`)
  - bare loopback (`127.0.0.1` / `localhost` / `[::1]`) → ambiguous, defaults to the daemon frame of reference (unchanged locally; rewritten to the daemon host remotely, with a `warning` echoed)

  Rewritten actions echo additive result fields `requestedUrl` / `finalUrl` / `rewritten: true` / `reason` (+ `warning` on bare-loopback); non-rewritten URLs keep a byte-identical result shape. Only the hostname of top-level `navigate`/`openTab` URLs is rewritten.

- **§5.14 (execution locus / remote behavior)** — cross-referencing bullet in the FE-served-intents list summarizing the convention and pointing at §5.9 for the full contract.

The daemon remains a thin proxy — the `browser.exec` request / reverse-RPC wire shape is unchanged, so no protocol version bump.

## Why

The FE implementation (cloudlands-fe `loopback-rewrite.ts` + `browser-action-executor.ts` wiring) landed the rewrite; PROTOCOL.md is the canonical wire contract and needs to record that hostname interpretation is an FE concern with additive result-echo fields.

## Verification

Doc-only change; reviewed against the FE implementation for field names and rewrite-table semantics.

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -2862,6 +2862,26 @@ The namespace holds the **two** workspace/active-PR-scoped methods that survived
 > Client-triggered `browser.exec` is **unchanged**: it still reverse-dispatches on the
 > caller's own connection.
 >
+> **Loopback-hostname interpretation — FE-side, wire shape unchanged (monorepo#2323).**
+> URL hostnames in `navigate` / `openTab` action URLs are interpreted **on the frontend
+> that serves the reverse RPC**, per the reserved-hostname convention (RFC 6761
+> `*.localhost` names): `daemon.localhost` targets the **daemon machine** (rewritten to
+> `127.0.0.1` on a local daemon, to the daemon host — the sanitized transport target —
+> on a remote one); `client.localhost` targets the **client (user's) machine** (always
+> rewritten to `127.0.0.1`); **bare loopback** (`127.0.0.1` / `localhost` / `[::1]`) is
+> ambiguous and defaults to the agent's frame of reference — the daemon: unchanged on a
+> local daemon, rewritten to the daemon host on a remote one. The daemon remains a
+> **thin proxy**: no rewrite happens daemon-side and the `browser.exec` request /
+> reverse-RPC wire shape is unchanged — the convention is entirely FE-served. Rewritten
+> actions echo **additive fields** in their result payload: `requestedUrl` (the URL as
+> requested), `finalUrl` (the URL actually loaded), `rewritten: true`, and a
+> human-readable `reason`; ambiguous bare-loopback rewrites additionally carry a
+> `warning` naming the explicit `daemon.localhost` / `client.localhost` forms.
+> Non-rewritten URLs keep a byte-identical result shape (no echo fields). Only the
+> hostname is rewritten (scheme, port, path, query, and hash are preserved), and only
+> top-level `navigate` / `openTab` URLs are interpreted — never URLs inside pages
+> (redirects, fetches, links).
+>
 > **`browser.docs` — not exposed.** The `browser_docs` MCP tool that
 > served static reference docs on-demand has no consumer in the daemon surface (skills-style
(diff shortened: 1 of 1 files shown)


panghy (commented):
**Verifier review — no changes needed.** Verified the new §5.9 paragraph and §5.14 bullet line-by-line against the FE implementation (cloudlands-fe `src/features/browser/main/loopback-rewrite.ts` + `browser-action-executor.ts` + the `browser.ipc.ts` / `browser-exec-reverse.ts` wiring):

- **Hostnames & rewrite table** — match `classifyLoopbackHost` / `rewriteLoopbackUrl` exactly: `daemon.localhost` → `127.0.0.1` locally / sanitized transport-target host remotely; `client.localhost` → always `127.0.0.1`; bare loopback (`127.0.0.1` / `localhost` / `[::1]`, incl. unbracketed `::1`) unchanged locally, rewritten to the daemon host remotely with a `warning` naming both explicit forms.
- **Echo fields** — `rewriteEcho()` emits exactly `requestedUrl` / `finalUrl` / `rewritten: true` / `reason` (+ `warning` when set), and returns `{}` for non-rewritten URLs, so the "byte-identical result shape" claim holds (the internal degraded-case `reason` on non-rewritten results is never echoed).
- **Hostname-only rewrite** — `withHostname()` replaces only `URL.hostname`; scheme/port/path/query/hash preserved. ✓
- **Top-level `navigate`/`openTab` only** — the rewrite is applied solely in those two action arms; no page-internal URL interception. ✓
- **FE-served claim** — confirmed the reverse-RPC handler (`browser-exec-reverse.ts`) routes into `executeBrowserActions`, which supplies `getDaemonLoopbackContext()` (via `loopbackContextFromTransport`), so interpretation happens on the frontend serving the reverse RPC as documented.
- **§5.9 / §5.14 consistency** — no contradictions with the surrounding REV-1 and thin-proxy paragraphs; cross-references (§5.9 ↔ §5.14) and section numbers are correct; style matches the surrounding contract-block prose. PR title is a valid conventional `docs:` commit and the `pr-title` check passes.

Two non-blocking observations (fine to leave as-is for a wire-contract doc):

1. The FE also runs a **reachability probe** before navigating to a URL rewritten to the *remote* daemon host (`probeRewrittenRemoteOrigin`, 1500 ms): an unreachable origin fails the action with an explanatory error (bind-0.0.0.0 hint) instead of navigating. That is agent-visible behavior in the result envelope, but it's an FE implementation detail rather than wire shape — documenting it is optional.
2. Degenerate case: in remote mode with an **unknown daemon host**, `daemon.localhost` / bare-loopback URLs are left unchanged (non-rewritten, no echo fields). The doc's unconditional "rewritten to the daemon host on a remote one" doesn't cover this, but the byte-identical non-rewritten contract still holds, so no contradiction.

panghy:
Addressed the review:

1. **Reachability probe** — intentionally left out of this PR: the probe (and the upcoming tunnel fallback) are being documented alongside their own changes (separate probe/tunnel doc tasks under monorepo#2323), so this PR stays scoped to the hostname-interpretation contract.
2. **Unknown-daemon-host degenerate case** — folded in as c6e77e2: §5.9 now notes that when the remote daemon host cannot be determined, daemon-targeting URLs are left unchanged (non-rewritten, no echo fields).

PR #2376, opened 2026-08-14
docs: describe event-chained release pipeline with cron backstops

panghy:
Updates the release-process docs to describe the event-chained release pipeline that replaced the pure :00/:15/:30 cron-offset cadence, landed in [intent-hq/intentd#1213](https://github.com/intent-hq/intentd/pull/1213) and [intent-hq/cloudlands-fe#1191](https://github.com/intent-hq/cloudlands-fe/pull/1191).

## What changed

**`AGENTS.md` (Release Process):**
- **intentd**: documents the fail-soft `intentd-alpha-published` `repository_dispatch` sent by `publish-channel-manifest.yml` to cloudlands-fe after the alpha manifest publishes (source + mirror), including the `FE_DISPATCH_TOKEN` prerequisite (fine-grained PAT, contents:write on cloudlands-fe) and its fail-soft semantics (missing secret / failed dispatch → warning, never fails the publish).
- **cloudlands-fe**: `auto-pin-intentd.yml` is now event-driven on that dispatch with the :15 cron as backstop; `auto-cut-beta.yml` is event-chained — a pin-bump push to `main` touching `intentd.version` chains into a cut run that polls (30s interval, up to 15 min) for a refreshed, green Release PR, while the :30 cron stays the backstop and the fe-only path (check-once-and-exit). Documents the open pin-PR defer and the in-flight intentd guardrail (newer intentd tag than the alpha manifest, younger than 90 min → defer; fails open on any lookup error).
- **Coordinated Release Ordering**: rewritten as the full chain — alpha publish → dispatch → pin bump → pin push → chained fe cut — with crons as fail-soft backstops.
- **Tracking shipped work**: intentd changes now ride a same-cycle fe alpha when the chain is live (next hourly cron cut when it falls back).

**`docs/fe/DEPLOYING.md`:** step 2 of the release-please flow updated from "auto-merged hourly (cron at :30)" to the event-chained description with the cron backstop and in-flight defer.

`docs/fe/RELEASING.md` mentions the pin but describes no cadence, so it is untouched.

## Verification

Cross-read against the merged workflow files on each repo's `main`:
- `intent-hq/intentd` `.github/workflows/publish-channel-manifest.yml` (dispatch step + header)
- `intent-hq/cloudlands-fe` `.github/workflows/auto-pin-intentd.yml` and `auto-cut-beta.yml` (triggers, polling bounds, guardrails, token semantics)

Docs-only change; no workflow files touched.

### DIFF
--- a/docs/fe/DEPLOYING.md
+++ b/docs/fe/DEPLOYING.md
@@ -42,8 +42,8 @@ version number by hand.
    (30s interval, up to 15 min) for release-please to refresh the Release PR
    and for CI Gate to go green, then merges — so a new intentd ships in the
    same fe alpha cycle. The hourly cron at :30 is the backstop and the normal
-   path for fe-only changes (check-once-and-exit, no polling); the cron cut
-   defers while an intentd release build is in flight (a semver tag on
+   path for fe-only changes (check-once-and-exit, no polling); every run type
+   defers the cut while an intentd release build is in flight (a semver tag on
    `intent-hq/intentd` newer than the published alpha manifest and younger
    than 90 minutes; fails open on any lookup error). The `hold-release`
    label on the Release PR pauses the auto-cut; a human can still merge
--- a/docs/fe/DEPLOYING.md
+++ b/docs/fe/DEPLOYING.md
@@ -35,8 +35,17 @@ version number by hand.
 1. On every push to `main`, release-please opens (or updates) a **Release PR**
    that bumps `package.json` and regenerates `CHANGELOG.md` from the
    conventional commits since the last `v*` tag.
-2. The Release PR is auto-merged hourly by `auto-cut-beta.yml` (cron at :30)
-   when it is green — that is the release timing gate. The `hold-release`
+2. The Release PR is auto-merged by `auto-cut-beta.yml` when it is green —
+   that is the release timing gate. The workflow is event-chained with an
+   hourly cron backstop: a sidecar pin-bump squash merge (a push to `main`
+   touching `intentd.version`) chains straight into a cut run that polls
+   (30s interval, up to 15 min) for release-please to refresh the Release PR
+   and for CI Gate to go green, then merges — so a new intentd ships in the
+   same fe alpha cycle. The hourly cron at :30 is the backstop and the normal
+   path for fe-only changes (check-once-and-exit, no polling); the cron cut
+   defers while an intentd release build is in flight (a semver tag on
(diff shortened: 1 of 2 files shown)


docs/fe/DEPLOYING.md
```diff
@@ -35,8 +35,17 @@ version number by hand.
...
+   that is the release timing gate. The workflow is event-chained with an
+   hourly cron backstop: a sidecar pin-bump squash merge (a push to `main`
+   touching `intentd.version`) chains straight into a cut run that polls
+   (30s interval, up to 15 min) for release-please to refresh the Release PR
+   and for CI Gate to go green, then merges — so a new intentd ships in the
+   same fe alpha cycle. The hourly cron at :30 is the backstop and the normal
+   path for fe-only changes (check-once-and-exit, no polling); the cron cut
+   defers while an intentd release build is in flight (a semver tag on
+   `intent-hq/intentd` newer than the published alpha manifest and younger
+   than 90 minutes; fails open on any lookup error). The `hold-release`
```
panghy: Minor: this phrasing attributes the in-flight defer to the cron cut only, but in `auto-cut-beta.yml` the "Check for an in-flight intentd release build" step runs on **all** run types (cron, push-chained, and manual dispatch) — the workflow comment reads "In-flight intentd release guardrail (all run types)". In practice it mostly fires on cron runs (a push-chained run follows the pin bump, whose alpha already published), so the statement is true as written, just narrower than the actual behavior; AGENTS.md's version doesn't scope it to cron. Non-blocking — consider "the cut defers" instead of "the cron cut defers" if this file is touched again.

panghy (commented):
Verified the doc text against the merged workflows on each repo's `main` (`intent-hq/intentd` `publish-channel-manifest.yml`; `intent-hq/cloudlands-fe` `auto-pin-intentd.yml` + `auto-cut-beta.yml`). Dispatch type/payload, FE_DISPATCH_TOKEN fail-soft semantics, :15/:30 cron backstops, push trigger on `intentd.version`, 30s/15-min bounded polling on push runs only, open pin-PR defer, and the 90-min fail-open in-flight guardrail all match the doc text. One minor wording nit below; non-blocking.

docs/fe/DEPLOYING.md
panghy: Fixed in ba389cb — reworded to "every run type defers the cut" so the guardrail is no longer described as cron-only, matching the workflow's "all run types" behavior.

Issue #2275, opened 2026-08-13
intentd: attachment upload sessions have no cap/TTL; pipelined chunk+commit misclassifies as -32603


panghy:
Follow-up from the review of intent-hq/intentd#1187 (the `file.attachmentUpload.*` staged upload surface, protocol v6.16) — two explicit-decision items deferred from that PR:

### 1. No cap or TTL on concurrent upload sessions

`workspace.import.begin` serializes per workspace; `file.attachmentUpload.begin` has no per-workspace or global session cap and no idle expiry. Any authenticated client can open unbounded sessions, each reserving up to 1 GiB of staging disk, and an abandoned session holds memory + disk until daemon restart — the lazy orphan sweep only removes dirs *without* live sessions, so a live-but-stalled session is never reclaimed.

Probably acceptable under the trusted-client threat model, but it deserves an explicit decision. Options: a small per-workspace concurrent-session cap, and/or an idle TTL after which a session is dropped and its staging dir swept.

### 2. Pipelined final-chunk + commit can fail `-32603`

`chunk` reserves its bytes under the registry lock but writes the chunk file after releasing it. A client that pipelines the final chunk and the commit (without awaiting the chunk response) can have the commit pass the completeness/gap checks and start assembly before the file write lands → `-32603 "read staged chunk failed"` for what is really a caller pipelining race. Recoverable via commit retry, and sequential clients never hit it — but the error is classified Internal rather than a caller error. If session TTLs are added, consider also tightening this classification (e.g. detect a reserved-but-unwritten seq at commit and return `-32602` advising a retry).

Ref: review findings 4–5 on intent-hq/intentd#1187.

panghy:
Fixed in intentd v0.6.19 (alpha).

Issue #1217, opened 2026-07-31
Sender-attribution chip missing on agent messages delivered via the wakeOrCreate path


panghy:
## Symptom

In the chat transcript, the sender-attribution chip ("<Agent> sent a message") is missing on agent-to-agent messages that were delivered through the `agent.wakeOrCreate` wake path, while messages sent via `agent.send` on the same agent render the chip correctly.

Observed while dogfooding (`make dev`): a coordinator sent a HALT message via `agent.send` (chip shows) and later a RESUME message via `wakeOrCreate` to the same, now-idle agent (chip missing).

## Root cause

The FE renders the chip from **row-level** `message.metadata` (`ChatMessage.svelte` → `getAgentMessageAttribution(message?.metadata)`). The daemon has three delivery paths for agent-originated messages, and one drops the row-level metadata:

| Path | Persist call | Row `metadata` | Chip |
|---|---|---|---|
| `send_message` direct (idle / interrupt) | `append_agent_message_with_id(..., message_metadata, ...)` | ✅ persisted | ✅ |
| Queue drain (`persist_user`) | `append_agent_message_with_metadata(..., message_metadata, ...)` | ✅ persisted | ✅ |
| **`deliver_wake_message`** (wakeOrCreate → live agent, idle branch) | `append_agent_message(...)` — **no metadata param** | ❌ NULL | ❌ |

`deliver_wake_message` (`crates/intent-services/src/agent_ops.rs`, ~L6185) folds the attribution **into the text block** as `messageMetadata` (`build_block`) but persists the row via `Store::append_agent_message`, which writes `metadata = NULL`. The store-only fallback `deliver_wake_message_store_only` (~L6305) has the same defect. The FE transcript path only reads row-level metadata, so no chip.

## Evidence (live dev DB)

For the affected agent (`agent-62420512…`):
- HALT (seq 2, via `agent.send`): row metadata = `{"type": "agent_message", "fromAgentId": ..., "fromAgentName": "Coordinator"}` → chip shows.
- RESUME (seq 4, via `wakeOrCreate`): row metadata = **NULL**; attribution survives only inside the text block's `messageMetadata`, which the chip path never reads → no chip.

## Fix

[...]

PR #2397, opened 2026-08-14
docs: document attachment upload session cap, idle TTL, and -32602 race reclassification

panghy:
Documents the [intent-hq/intentd#1217](https://github.com/intent-hq/intentd/pull/1217) hardening of the `file.attachmentUpload.*` staged-upload surface (#2275) in `docs/PROTOCOL.md`. The daemon side is already merged, so intentd-first ordering is satisfied.

## What's documented

- **Per-workspace cap (4 live sessions)** — a `begin` at the cap is `-32602` naming the live count ("commit or abort one before beginning another"). The expired-session drain, per-workspace count, and insertion happen under one registry lock hold, so concurrent begins cannot overshoot the cap.
- **Idle TTL (15 minutes), lazy expiry** — the next `begin` drains expired sessions and reclaims their staging dirs (so expired sessions never hold cap slots); a late `chunk`/`commit` on an expired id gets `-32602` ("expired after Ns of inactivity — begin a new upload"). Sessions never expire while a commit is in flight, and a *failed* commit refreshes the idle clock so the retry-after-more-chunks window is a fresh 15 minutes.
- **Pipelined-race reclassification** — a commit catching a reserved-but-unwritten or partially-written chunk (a `chunk` call still in flight) is now `-32602` advising to wait for the chunk call to return and retry the commit, formerly `-32603` Internal; the session stays alive.

## Where

- §2 version history: "Also within 6.17" entry (behavior only, no method-catalog or wire-shape change, so no version bump)
- §5.9 method-catalog rows for `begin` / `chunk` / `commit`
- §5.9 new "Session bounds" prose block after the existing `file.attachmentUpload.*` block

Refs #2275 (closed by the intentd PR merge).


docs/PROTOCOL.md
```diff
@@ -23,7 +23,7 @@ This document is the canonical wire contract between Intent clients (desktop, iO
...
...ipped, and the whole path is best-effort — a workspace-read or unarchive failure logs a warning and the turn proceeds. Also within 6.17 (behavior only, no method-catalog or wire-shape change, so no version bump — [intent-hq/intentd#1217](https://github.com/intent-hq/intentd/pull/1217), [monorepo#2275](https://github.com/intent-hq/monorepo/issues/2275)): **attachment upload session bounds** (§5.9) — the `file.attachmentUpload.*` staged sessions gain a per-workspace cap of **4** live sessions (a `begin` at the cap is -32602 naming the live count and advising commit/abort) and a **15-minute idle TTL** expired lazily (the next `begin` drains expired sessions and reclaims their staging dirs so they never hold cap slots; a late `chunk`/`commit` on an expired id is -32602 "expired after Ns of inactivity — begin a new upload"); a session never expires while a commit is in flight, and a failed commit refreshes the idle clock so the retry-after-more-chunks window is fresh; and the pipelined chunk+commit race (a commit catching a reserved-but-unwritten or partially-written chunk) is reclassified from -32603 Internal to -32602 advising to wait for the chunk call to return and retry the commit.
```
panghy: "reclassified from -32603 Internal to -32602" overstates the change: in the pre-#1217 code (intentd ebdad751) only the reserved-but-unwritten guise was `Error::Internal("read staged chunk {seq} failed: …")` → -32603. The partial-write/size-mismatch guise was already `Error::InvalidParams("assembled attachment is {total} bytes, expected {declared_size}")` → -32602 — #1217 only added the race explanation and retry advice to its message. Suggest something like: "…the pipelined chunk+commit race is now uniformly -32602 with wait-and-retry advice — the reserved-but-unwritten guise reclassified from -32603 Internal, the partial-write guise (already -32602) gaining the same advisory wording."

docs/PROTOCOL.md
```diff
@@ -2828,9 +2828,9 @@ The namespace holds the **two** workspace/active-PR-scoped methods that survived
...
+| file.attachmentUpload.commit *(v6.16)* | uploadId (req) | { ok, path, fileName, size, attachmentId, mimeType?, uploadedAt } — byte-shape-identical to a successful file.placeAttachment result: verifies staged bytes = sizeBytes with gap-free seqs from 0 and a matching SHA-256, then places through the same collision-safe placement + attachment-registry path. A failed commit leaves the session alive for retry or abort (and refreshes the idle clock, monorepo#2275); incomplete/gapped/mismatched payloads are -32602. A commit on an idle-expired session is -32602 ("expired … — begin a new upload"), and a commit racing an in-flight chunk (the pipelined chunk+commit race) is -32602 advising to wait for the chunk call to return and retry — formerly -32603 Internal (monorepo#2275) |
```
panghy: Same issue as the version-history entry: "formerly -32603 Internal" only holds for the reserved-but-unwritten guise. The partial-write (size-mismatch) guise of the race was already -32602 before #1217; only its message changed to carry the retry advice.

docs/PROTOCOL.md
```diff
@@ -2940,6 +2940,33 @@ The namespace holds the **two** workspace/active-PR-scoped methods that survived
...
+> **never expired while a commit is in flight**, and a *failed* commit (checksum
+> mismatch, incomplete staging) refreshes the clock as it releases the commit
+> claim, so the documented retry-after-more-chunks window is a fresh 15 minutes
+> even when the commit itself outlived the TTL. **Pipelined-race
+> reclassification:** a commit that catches a reserved-but-unwritten or
+> partially-written chunk (a `chunk` call still in flight when `commit` fires) is
+> a caller-sequencing error, not a daemon fault — both guises are now -32602
+> (`chunk N is still being written — wait for the chunk call to return, then
+> retry the commit` / `assembled attachment is N bytes, expected M — a chunk may
+> still be being written; …`), formerly -32603 Internal; the session stays alive
```
panghy: "both guises are now -32602 (…), formerly -32603 Internal" — the "formerly -32603" attaches to both guises but is only true for the first. Pre-#1217 (ebdad751): missing chunk file → `Error::Internal` (-32603); short assembly → `Error::InvalidParams("assembled attachment is {total} bytes, expected {declared_size}")` (-32602, no advice). Suggest: "…both guises now carry the wait-and-retry advice — the reserved-but-unwritten guise reclassified from -32603 Internal, the partial-write guise already -32602 but formerly a bare size mismatch with no advice; the session stays alive…"

docs/PROTOCOL.md
```diff
@@ -2940,6 +2940,33 @@ The namespace holds the **two** workspace/active-PR-scoped methods that survived
...
+> new session's insertion happen under one registry lock hold, so concurrent
+> begins serialize and cannot overshoot the cap. **Idle TTL:** a session with no
+> begin/chunk/commit activity for **15 minutes** expires lazily, mirroring the
+> orphaned-staging sweep — the next `begin` (any workspace) drains expired
+> sessions and reclaims their staging dirs (outside the lock; expired sessions
+> never hold cap slots), while a late `chunk`/`commit` on an expired id gets
+> -32602 — `attachment upload <id> expired after Ns of inactivity — begin a new
+> upload`. Each successful `begin`/`chunk` refreshes the idle clock; a session is
+> **never expired while a commit is in flight**, and a *failed* commit (checksum
+> mismatch, incomplete staging) refreshes the clock as it releases the commit
```
panghy: Minor, non-blocking: "a *failed* commit (checksum mismatch, incomplete staging) refreshes the clock as it releases the commit claim" — the incomplete-staging (and gapped-seqs) failures are rejected in phase-1 validation *before* the `committing` claim is taken, so there is no claim release on that path; their refresh comes from the `last_activity` touch at validation time. Only phase-2 failures (checksum mismatch, placement, the race guises) refresh via `release_failed_commit_claim`. The observable contract (any failed commit leaves a fresh 15-minute window) is correct either way, so feel free to leave as-is or drop "as it releases the commit claim".

panghy (commented):
Verified against the merged implementation (intent-hq/intentd `origin/main`, `crates/intent-services/src/attachment_upload.rs`, landed in intentd@fa4aece4 / #1217; pre-state ebdad751 / #1187).

**Accurate (checked against source):**
- Cap = 4 (`ATTACHMENT_UPLOAD_MAX_SESSIONS_PER_WORKSPACE`); begin-at-cap error wording matches the code byte-for-byte (`workspace <id> already has N attachment uploads in progress (max 4) — commit or abort one before beginning another`), and the drain → per-workspace count → insert really do happen under one registry lock hold, with expired staging dirs deleted outside the lock (even when the begin was rejected at the cap).
- Idle TTL = 15 min (`ATTACHMENT_UPLOAD_IDLE_TTL`); lazy expiry via the next `begin` (drain is registry-wide, so "any workspace" is right); late `chunk`/`commit` error wording matches exactly; `expired()` requires `!committing`, so committing sessions never expire; `release_failed_commit_claim` refreshes `last_activity` (regression-tested in `failed_commit_outliving_ttl_still_gets_fresh_retry_window`).
- Both race-guise messages quoted in the prose block match the code exactly.
- Conventions: the "Also within 6.17" entry, the method-table row annotations, and the prose-block heading style are all consistent with their neighbors.

**One inaccuracy** (three occurrences — version history, commit table row, prose block): "formerly -32603 Internal" is only true for the reserved-but-unwritten guise. See inline comments. Plus one minor non-blocking mechanism nit.

docs/PROTOCOL.md
panghy: Fixed in 71efeec — the three spots now distinguish the guises: reserved-but-unwritten reclassified from -32603 Internal to -32602; partially-written was already -32602 and gains the retry advice.

docs/PROTOCOL.md
panghy: Fixed in 71efeec — the three spots now distinguish the guises: reserved-but-unwritten reclassified from -32603 Internal to -32602; partially-written was already -32602 and gains the retry advice.

docs/PROTOCOL.md
panghy: Fixed in 71efeec — the three spots now distinguish the guises: reserved-but-unwritten reclassified from -32603 Internal to -32602; partially-written was already -32602 and gains the retry advice.

docs/PROTOCOL.md
panghy: Fixed in 71efeec — dropped the "as it releases the commit claim" mechanism claim; the prose now just says a failed commit refreshes the clock.

panghy:
Re-verified at 71efeec against intentd origin/main (fa4aece4) and the pre-#1217 state (ebdad751): all three spots (version-history entry, `commit` table row, prose block) now correctly distinguish the guises — reserved-but-unwritten reclassified from -32603 Internal to -32602, partially-written already -32602 and gaining only the retry advice — and the prose block no longer claims the claim-release refresh mechanism. All other semantics (cap 4 + one-lock-hold atomicity, 15-min lazy idle TTL, committing-never-expires, failed-commit clock refresh, exact error wordings) were verified accurate in the previous pass and are unchanged. No remaining inaccuracies from my review.

PR #2432, opened 2026-08-14
docs: task.subscribe specLinked refreshes on spec-body edits (#2407)

### LINKED ISSUE #2407: task.subscribe: spec-body edits do not refresh specLinked on already-emitted task rows
Follow-up from the review of intent-hq/intentd#1218 (which carries `specLinked` on §6.9 task-channel snapshot/delta payloads).

## Gap

Spec-body edits do not refresh already-emitted `specLinked` flags on the task channel. When the spec note is edited to add/remove an `intent://local/task/{id}` link, the task channel sees `note:updated` for `"spec"`, which `task_delta` maps to `removedIds: ["spec"]` (the spec has no task metadata) — no `updated` rows are emitted for the tasks whose linkage just flipped. A subscriber holds the stale `specLinked` until the next event touches each affected task, or it resubscribes.

This is parity-consistent with `task.list` (read-time semantics), but a live channel implies freshness.

## Suggested fix

On a spec `note:updated`, diff the extracted link set against the previous one and emit `updated` rows for the tasks whose linkage flipped — bounded by one `list_notes` read per spec edit, keeping the O(rows) cost contract.

## Interim state

The behavior is documented as a freshness caveat in PROTOCOL.md §6.9 (monorepo#2402).

panghy:
Documents the daemon-side fix from [intent-hq/intentd#1224](https://github.com/intent-hq/intentd/pull/1224) (merged; fixes intent-hq/monorepo#2407): spec-body edits now refresh flipped `specLinked` flags on the `task.subscribe` channel.

Changes to `docs/PROTOCOL.md`:

- **§6.9 `task.subscribe` row**: replaces the freshness caveat (spec `note:updated` → `removedIds: ["spec"]`, stale flags until the next per-task event or resubscribe) with the new contract — the forwarder tracks the spec's `intent://local/task/{id}` link set, spec `note:updated`/`note:deleted` diff against it and emit `updated` rows for exactly the flipped tasks, neutral spec edits emit nothing, dangling links flip silently, spec deletion re-emits every tracked task `specLinked: false`, and per-task events stamp only their own id (race-safe). Cost contract extended: one additional bounded spec read per spec event (O(links in spec)).
- **§1 version history (within 6.17)**: extends the intentd#1218 note to record the intentd#1224 refresh behavior.

Docs-only; no daemon or FE code changes ride this PR.


docs/PROTOCOL.md
```diff
@@ -6623,7 +6623,7 @@ The `events.subscribe` firehose (§6.1) carries every bus event; a thin client t
...
...intent-hq/intentd/pull/1224), fixes [monorepo#2407](https://github.com/intent-hq/monorepo/issues/2407))*: the forwarder tracks the spec's `intent://local/task/{id}` link set for the life of the subscription, and a spec `note:updated` / `note:deleted` diffs the current link set against the tracked one and emits `updated` rows (fresh row + current `specLinked`) for exactly the tasks whose linkage flipped — the spec's own `note:updated` no longer maps to the junk `removedIds: ["spec"]`, a spec edit that leaves the link set unchanged emits no delta (unrelated tasks are never re-emitted on spec keystrokes), a flipped link to a nonexistent task note flips silently (no row to refresh, but the tracked set still advances), and spec **deletion** re-emits every previously linked task with `specLinked: false`. Per-task events stamp only their own note id into the tracked set, so a task event whose fresh spec read races ahead of the spec's own `note:updated` cannot swallow another task's flip. Cost contract: the flag is derived from ONE spec-note read per snapshot / per delta batch (O(rows) overall), and a spec event costs one additional bounded spec read for the flip diff (O(links in spec)). |
```
panghy: The new cost-contract clause — "a spec event costs one additional bounded spec read for the flip diff (O(links in spec))" — mischaracterizes the merged implementation on three counts:

1. **Read kind**: `spec_link_flip_delta` performs one `list_notes` read over the whole workspace, not a bounded spec-note read (`get_note("spec")` is what the *per-task* `stamp_spec_linked` path does). intentd#1224's own PR body says "a spec keystroke costs one bounded `list_notes` read".
2. **Complexity**: that single read yields both the new link set and the task rows to re-emit, so the spec-event cost is O(rows) — the code comment says "keeping the O(rows) cost contract — no per-task scans" — not O(links in spec); only the diff itself is O(links).
3. **"one additional"**: on the spec-event path the `list_notes` read is the *only* read — the per-task spec-stamp read does not also run — so nothing is "additional".

Suggested wording: "…and a spec event costs one `list_notes` read that yields both the new link set and the rows to re-emit (O(rows), same as the snapshot — no per-task scans)."

Non-blocking side note: the pre-existing "a transient spec-read failure likewise degrades rather than dropping the delta" sentence describes the per-task stamping path only — on the new spec-event path a `list_notes` failure *drops* the delta and keeps the tracked set (reconverging on the next spec read). Worth a clarifying clause if you're touching this sentence anyway, since the row now covers both paths.

panghy (commented):
Verified the two hunks against the merged implementation on intentd `main` (17e7324c, [intent-hq/intentd#1224](https://github.com/intent-hq/intentd/pull/1224); `crates/intent-transport/src/subscriptions.rs` — `task_snapshot` / `task_delta` / `stamp_spec_linked` / `spec_link_flip_delta` — and `conn.rs` `forward_channel_subscription`).

Accurate against the code:

- **Tracked link set for the life of the subscription** — `forward_channel_subscription` holds `spec_links`, seeded from `task_snapshot`'s own `list_notes` read.
- **Flip diff on spec `note:updated` / `note:deleted`** — both event types route to `spec_link_flip_delta`, which takes the symmetric difference against the tracked set and emits `updated` rows (fresh row + current `specLinked`) for exactly the flipped task notes.
- **Neutral edits emit nothing** — empty `updated` → `None` (no delta push, `seq` untouched).
- **Dangling links flip silently** — flipped ids without task metadata are filtered out of `updated`, and `*linked = new_linked` advances the tracked set unconditionally.
- **Spec deletion** — the spec row is gone from `list_notes`, so `new_linked` is empty and every previously linked task re-emits with `specLinked: false`.
- **Race-safety of per-task stamps** — `stamp_spec_linked` inserts/removes ONLY the event's own note id, matching the "cannot swallow another task's flip" claim.
- **§1 version-history extension** — accurate, follows the within-6.17 convention with full cross-repo links and the `see §6.9` pointer.
- **No leftovers** — the old "Freshness caveat" text is fully removed; the only remaining `removedIds: ["spec"]` occurrences are the two intentional historical references (§1, §6.9), and the §5.4 rows (task.list/task.get) needed no change.

One wording inaccuracy in the new cost-contract sentence, flagged inline — approvable once that line is fixed.

docs/PROTOCOL.md
panghy: Fixed in 9176fc3 — adopted your suggested wording for the cost clause (one `list_notes` read yielding both the new link set and the rows to re-emit, O(rows), no per-task scans), and also took the non-blocking side note: the degradation sentence now distinguishes the per-task stamping path (degrades to `false`) from the spec-event path (drops that delta, keeps the tracked set, reconverges on the next spec read).

PR #2601, opened 2026-08-16
docs: document system.status workspaces-root disk fields

panghy:
Documents the additive `workspacesDiskAvailableBytes` / `workspacesDiskTotalBytes` fields on the `system.status` result, added by intent-hq/intentd#1261: the available/total bytes of the mounted volume containing the daemon's resolved workspaces root, served from a background sampler (~30s cadence, synchronous first sample).

Both fields follow the presence-detection convention — absent (never `null`/`0`) until the first sample lands or when no mounted volume matches the root. Additive response fields on an existing method, so no protocol version bump per the documented convention.

### DIFF
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -358,6 +358,22 @@ When the aggregate agent memory budget is on (`agents.memoryBudgetMb` > 0, §5.1
 - `queuedSpawns` counts the agent spawns currently parked in the admission wait queue — whether they queued on the concurrency slot cap or on the memory budget. `0` means nothing is waiting.
 - Unlike the child-tree fields above (always present, `null` until sampled), all three follow the **presence-detection convention**: when the budget is off (`agents.memoryBudgetMb = 0`, the default) they are **absent** — never `null`. Reading them never perturbs admission state. Additive response fields shipped without a version bump (the method surface is unchanged); clients must detect them by **presence**, not by protocol version.
 
+#### `system.status` — workspaces-root disk fields (additive)
+
+The `system.status` result additionally reports the disk space of the **volume containing the daemon's resolved workspaces root** — the parent directory workspace checkouts are provisioned under (startup-pinned `workspaces.root` / `INTENTD_WORKSPACES_DIR`, else the `~/intent/workspaces` default; fixed for the daemon's lifetime) — so a client can warn when workspace provisioning is about to run out of space:
+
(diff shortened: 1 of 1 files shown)


panghy:
Verifier cross-check against the implementation on [intent-hq/intentd#1261](https://github.com/intent-hq/intentd/pull/1261): field names, jsonc example shape, sampler cadence (~30s, synchronous first sample), longest-mount-point matching against the canonicalized root, and the presence-detection convention all match the code. Two wording accuracy nits:

1. **"or when no mounted volume matches the root (e.g. the configured directory does not exist yet)"** — on Unix the sampler falls back to prefix-matching the *raw* path when canonicalization fails, and any absolute path still matches the `/` mount, so a not-yet-created root typically still yields **present** fields (describing the root volume). Absence in practice means an empty disks list (locked-down container) or an unmounted Windows drive letter. Suggest swapping the example so clients don't infer "absent ⇒ root missing". (Same example appears in the `SystemStatus` doc comment — flagged inline on intentd#1261.)

2. **"the parent directory workspace checkouts are provisioned under"** — slightly overstated: when `workspace.worktreesLocation` is set (and `workspaces.root` isn't startup-pinned), `resolve_workspaces_parent` provisions new checkouts under that location instead, which can be a different volume than the one these fields describe. There's an open review thread about this on intentd#1261; whichever way it resolves, this sentence should match (e.g. "the daemon's configured workspaces root" with a caveat, or updated behavior).

Neither nit blocks the fields/convention documentation itself, but since PROTOCOL.md is the canonical wire contract they're worth fixing before merge (and this PR should land only after intentd#1261, per the intentd-first rule).

panghy:
Both nits addressed in e3402e3:

1. The absence example is now an empty disks list (locked-down container) or an unmounted Windows drive letter, with an explicit note that a not-yet-created root still matches its would-be volume — clients cannot infer "absent ⇒ root missing".
2. The resolution wording now describes the provisioning-parent precedence `workspace.create` actually uses (startup-pinned `workspaces.root` / `INTENTD_WORKSPACES_DIR` > non-empty absolute `workspace.worktreesLocation` > `~/intent/workspaces` default, resolved once at boot). The open thread on intentd#1261 resolved in favor of matching the create path — the sampler was updated accordingly (intentd 9777de36), so the doc and behavior now agree.

Merge order noted: this PR lands only after intentd#1261.

PR #2645, opened 2026-08-16
docs: harness versioning documentation

panghy:
Adds `docs/HARNESS.md`, a durable engineering doc for the harness versioning system, and indexes it in `docs/README.md`.

## What it covers

- The **doctrine vs. reference** split: versioned instruction/prompt text vs. the never-versioned wire protocol / MCP tool schemas / runtime semantics.
- The permanent creation-time stamp: `harnessVersion` (currently `"1.0"`, immutable, creation-time-only — mixed-version agent trees are expected) and `harnessFeatures` (the effective `agentFeatures` snapshot the session actually runs with, including the two documented live exceptions and the legacy NULL-snapshot materialization path).
- Backend implementation in `intentd`: the `Harness` trait (one method per text surface), one module per version, per-version `Doctrine` bundles under `resources/agent-instructions/<ver>/` + `resources/specialists/<ver>/`, the `REGISTRY` / `HarnessEntry` mapping with fail-open unknown-version resolution, and per-session resolution in prompt assembly.
- Golden-test byte pinning (`v1_goldens.rs`, `v1_turn_envelope_goldens`) and the mechanical steps for minting a new harness version.
- The read-only frontend surface in `cloudlands-fe` (agent tab / AgentCard menu item → `HarnessFeaturesModal`, snapshot-wins union with the feature catalog, legacy/absent-field behavior).
- Design invariants (permanent stamps, all versions bundled, byte changes are version decisions, fail-open resolution).

Wire shapes are referenced to `docs/PROTOCOL.md` §5.5 rather than duplicated. Every claim was verified against the current `intent-hq/intentd` and `intent-hq/cloudlands-fe` main branches.

Part of intent-hq/monorepo#2459.

### DIFF
--- a/docs/HARNESS.md
+++ b/docs/HARNESS.md
@@ -13,6 +13,10 @@ envelope, wake/queue system messages, notices, and the bundled instruction
 and specialist markdown ("doctrine"). Versioning it gives every agent session
 a permanent, creation-time pin to the exact doctrine it was created under, so
 a daemon upgrade never silently changes what an existing session runs with.
+Today the session pin governs system-prompt/doctrine assembly (and the
+specialist bundle); the turn-envelope and wake/notice surfaces are versioned
+behind the same trait but their call sites still resolve the latest harness —
+routing them through the session's stamp is the intended end state.
 
 ## Doctrine vs. reference
 
@@ -55,7 +59,9 @@ served on the `AgentSession` and `AgentLite` projections (shapes in
 Legacy sessions persisted before the feature snapshot existed (NULL in the
 store) follow the live effective settings on read until their first
 post-launch activation, which materializes the snapshot once from the
-resolved live values and persists it (idempotent: the store write is guarded
+resolved live values — with one exception: the legacy per-session `taskGraph`
+pin, where present, wins over the live setting for that key, matching the
+pre-freeze read — and persists it (idempotent: the store write is guarded
 on `harness_features IS NULL`). From then on the row reads its frozen
 snapshot like any new session; `harnessVersion` stays `"1.0"` — only the
 flags freeze.
--- a/docs/HARNESS.md
+++ b/docs/HARNESS.md
@@ -104,9 +104,11 @@ The doctrine text is **byte-pinned**: `intent-services/src/v1_goldens.rs`
 asserts full literal bytes (or a SHA-256 pin for the large bundled layers)
 of every system-generated string that reaches an agent conversation, and
 `agent_manager::v1_turn_envelope_goldens` pins the composed turn envelope.
-Any wording or whitespace change fails the goldens and forces a deliberate
-decision: update the golden (byte-neutral refactor) or mint a new harness
-version (material doctrine change).
+Any wording or whitespace change to a versioned surface fails that
+version's goldens and must be answered by minting a new harness version —
+a version's goldens are never updated to absorb a doctrine change. Code
+moves that keep the output byte-identical (refactors) pass the goldens
+unchanged; that is what they exist to prove.
 
 ### Minting a new harness version
 
@@ -133,10 +135,14 @@ The stamp is read-only in the UI. The agent tab `⋯` menu and the AgentCard
 context menu show a "Harness v{version}" item; selecting it opens
 `HarnessFeaturesModal.svelte`, a read-only list of the session's
 `harnessFeatures` snapshot (name + description per row, On/Off state).
-Legacy sessions without a snapshot open the modal too — every catalog
-feature renders Off — while sessions from daemons that predate the field
-omit the menu item entirely. Unknown snapshot keys (from a newer daemon)
-are rendered with humanized labels rather than dropped.
+The daemon always serves a `harnessFeatures` value (legacy pre-snapshot
+rows follow the live effective settings until activation freezes them —
+see above), so the modal normally renders a real snapshot; as a defensive
+fallback, an absent snapshot renders every catalog feature Off. Sessions
+from daemons that predate the field omit the menu item entirely. Snapshot
+keys unknown to the catalog (from a newer daemon) are rendered with
+humanized labels rather than dropped, and catalog keys absent from the
+snapshot render Off (an older harness never had the newer features).
 
 ## Design invariants
 
--- a/docs/HARNESS.md
+++ b/docs/HARNESS.md
@@ -1,0 +1,150 @@
+# Harness Versioning
+
+> Audience: engineers working on `intentd` (`packages/intentd`) and
+> `cloudlands-fe` (`packages/cloudlands-fe`). Companion documents:
+> [ARCHITECTURE.md](./ARCHITECTURE.md) (backend architecture) and
+> [PROTOCOL.md](./PROTOCOL.md) §5.5 (the wire contract for
+> `harnessVersion` / `harnessFeatures` — wire shapes are specified there,
+> not here).
+
+The **harness** is the versioned owner of every system-generated string that
+shapes an agent's behavior: the assembled system prompt, the per-turn prompt
+envelope, wake/queue system messages, notices, and the bundled instruction
+and specialist markdown ("doctrine"). Versioning it gives every agent session
+a permanent, creation-time pin to the exact doctrine it was created under, so
+a daemon upgrade never silently changes what an existing session runs with.
+
+## Doctrine vs. reference
+
+Two layers with different versioning rules:
+
+- **Doctrine** — the instruction/prompt *text* and the `agentFeatures` values
+  a session was created under. Versioned by the harness; pinned per session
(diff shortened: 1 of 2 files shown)


docs/HARNESS.md
```diff
@@ -0,0 +1,150 @@
...
+bundled markdown is used untouched.
+
+### Golden tests
+
+The doctrine text is **byte-pinned**: `intent-services/src/v1_goldens.rs`
+asserts full literal bytes (or a SHA-256 pin for the large bundled layers)
+of every system-generated string that reaches an agent conversation, and
+`agent_manager::v1_turn_envelope_goldens` pins the composed turn envelope.
+Any wording or whitespace change fails the goldens and forces a deliberate
+decision: update the golden (byte-neutral refactor) or mint a new harness
```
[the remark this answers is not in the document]
panghy: Fixed in 16a2733 — reworded: goldens are never updated to absorb a doctrine change; any wording/whitespace change to a versioned surface must mint a new version, and byte-identical refactors pass the goldens unchanged.

docs/HARNESS.md
```diff
@@ -0,0 +1,150 @@
...
+are coherent (unique versions, wired doctrine, non-empty labels), and that
+unknown versions fall back to the latest.
+
+## Frontend surface (`cloudlands-fe`)
+
+The stamp is read-only in the UI. The agent tab `⋯` menu and the AgentCard
+context menu show a "Harness v{version}" item; selecting it opens
+`HarnessFeaturesModal.svelte`, a read-only list of the session's
+`harnessFeatures` snapshot (name + description per row, On/Off state).
+Legacy sessions without a snapshot open the modal too — every catalog
```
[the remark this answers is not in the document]
panghy: Fixed in 16a2733 — clarified that the daemon always serves a harnessFeatures value (legacy rows read live until activation freezes them), the all-Off rendering is a defensive fallback, and daemons predating the field omit the menu item.

docs/HARNESS.md
```diff
@@ -0,0 +1,156 @@
...
+> [PROTOCOL.md](./PROTOCOL.md) §5.5 (the wire contract for
+> `harnessVersion` / `harnessFeatures` — wire shapes are specified there,
+> not here).
+
+The **harness** is the versioned owner of every system-generated string that
+shapes an agent's behavior: the assembled system prompt, the per-turn prompt
+envelope, wake/queue system messages, notices, and the bundled instruction
+and specialist markdown ("doctrine"). Versioning it gives every agent session
+a permanent, creation-time pin to the exact doctrine it was created under, so
+a daemon upgrade never silently changes what an existing session runs with.
```
panghy, line 15: Accuracy nuance: as of intentd main, only **system-prompt assembly** (and the specialist/doctrine bundle) resolves through the session's pin (`rules.rs::session_harness_entry`). The turn-envelope composition and wake/notice surfaces are versioned behind the `Harness` trait but their call sites currently resolve `crate::harness::latest()` — `harness/mod.rs` says so itself: "Call sites not yet routed through a session's stamp (turn envelope, wakes) use this." So a daemon upgrade *does* change envelope/wake wording for existing sessions today; only the doctrine is pinned. Since the trait bullet below lists turn-envelope composition, completion wakes, and PR-monitor notices as harness surfaces, a reader will infer those are session-pinned too. Suggest one clarifying sentence (e.g. in the "Prompt assembly resolves the session's pinned entry" paragraph): today the session pin governs doctrine/system-prompt assembly; envelope and wake surfaces route through the trait but are latest-bound pending per-session routing.

docs/HARNESS.md
```diff
@@ -0,0 +1,156 @@
...
+  from the persisted snapshot, not the live settings, so the wire report
+  never disagrees with the runtime surface. Two documented exceptions stay
+  live (`stateSnapshot` is read live each turn; `backgroundHooks` is
+  re-checked live on every `hook.schedule`) — for those, the captured value
+  records the creation-time setting without freezing the behavior.
+
+Legacy sessions persisted before the feature snapshot existed (NULL in the
+store) follow the live effective settings on read until their first
+post-launch activation, which materializes the snapshot once from the
+resolved live values and persists it (idempotent: the store write is guarded
```
panghy: Small omission: the materialization isn't purely "from the resolved live values" — `materialize_legacy_harness_features` also folds in the legacy per-session taskGraph pin (the 0095 `task_graph_enabled` column), which **wins over the live setting** for that key (matching the pre-freeze COALESCE read). Worth a parenthetical, since it's the one key where the frozen snapshot can differ from the live settings at freeze time.

docs/HARNESS.md
```diff
@@ -0,0 +1,156 @@
...
+bundled markdown is used untouched.
+
+### Golden tests
+
+The doctrine text is **byte-pinned**: `intent-services/src/v1_goldens.rs`
+asserts full literal bytes (or a SHA-256 pin for the large bundled layers)
+of every system-generated string that reaches an agent conversation, and
+`agent_manager::v1_turn_envelope_goldens` pins the composed turn envelope.
+Any wording or whitespace change to a versioned surface fails that
+version's goldens and must be answered by minting a new harness version —
```
panghy, line 114: Wording nit: "update the golden (byte-neutral refactor)" is self-contradictory — a byte-neutral change wouldn't fail the goldens. What's meant is a deliberate byte change that is *not doctrine-material* (e.g. fixing a typo without minting a version). Suggest something like "update the golden (deliberate non-material byte change)".

panghy (commented):
Reviewed for technical accuracy against `intent-hq/intentd` origin/main (d28ab37) and `intent-hq/cloudlands-fe` origin/main (c3b87f5), wire-shape duplication vs `PROTOCOL.md` §5.5, banned vocabulary, and style consistency with `ARCHITECTURE.md`.

**Accuracy — verified.** Every checkable claim matches the implementation: `CURRENT_HARNESS_VERSION = "1.0"` (`intent-core/src/model.rs:2261`); `Harness` trait / `HarnessEntry` / `Doctrine` / `REGISTRY` / `LATEST_VERSION` and the fail-open WARN fallback (`harness/mod.rs`); `rules.rs::session_harness_entry`; the feature gates named (taskGraph, backgroundHooks, scripts, terminalAccess, richChatBlocks in `instructions.rs`; structuredQuestions, attentionRequests in `rules.rs`); `v1_goldens.rs` literal-bytes + SHA-256 pinning and `agent_manager::v1_turn_envelope_goldens`; registry coherence tests (`registry_rows_are_coherent`, unknown-version fallback, `latest_is_current_harness_version`); migration `0096` backfill to `'1.0'` with NULL features; `legacy_import.rs:1688` stamping the literal `"1.0"`; `materialize_legacy_harness_features` guarded on `harness_features IS NULL` at the `ensure_started` choke point; `session_agent_features` snapshot-wins respawn behavior; the two live exceptions (stateSnapshot read live per turn in `agent_ops.rs:7453`, backgroundHooks re-checked live in `hook_manager.rs:440`). FE claims match `AgentTabType.svelte`, `AgentCard.svelte`, `HarnessFeaturesModal.svelte` + `harness-feature-catalog` exactly (menu item gated on `harnessVersion` presence, snapshot-wins union, all-Off for null snapshots, humanized unknown keys appended after the catalog).

**No wire-shape duplication.** The JSON shapes live only in PROTOCOL §5.5; this doc summarizes semantics in prose and points there. Good.

**Vocabulary.** No "wave" or H-number terms anywhere in the doc or the README addition (note the *source code* doc comments in `harness/mod.rs` do use H-numbers, but this doc correctly does not import them).

**Style.** Consistent with `ARCHITECTURE.md` (audience header, component-scoped sections, invariants list, cross-links).

Three inline comments below — one accuracy nuance about the scope of the session pin, one omitted detail in the legacy-freeze paragraph, one wording nit. None are blocking; I am deliberately not approving/merging per review scope.

docs/HARNESS.md
panghy: Fixed in 9167afb — the intro now states that today only system-prompt/doctrine assembly (and the specialist bundle) resolves through the session pin, while turn-envelope and wake/notice call sites still resolve the latest harness, with session-stamp routing as the intended end state.

docs/HARNESS.md
panghy: Fixed in 9167afb — added the exception: the legacy per-session taskGraph pin, where present, wins over the live setting for that key during materialization, matching the pre-freeze read.

docs/HARNESS.md
panghy: This anchor is stale — the quoted wording was already removed in 16a2733 (pushed before this review landed). The section now says goldens are never updated to absorb a doctrine change: any wording/whitespace change to a versioned surface mints a new version, and byte-identical refactors pass the goldens unchanged. No 'update the golden' route remains.

PR #2743, opened 2026-08-17
docs: update A2A interrupt defaults

panghy:
Documents the landed semantics of [intent-hq/intentd#1292](https://github.com/intent-hq/intentd/pull/1292) (squash commit `79e5600`): the agent-facing `ws.agent.send` / `ws.agent.sendToTask` MCP bindings now deliver with **interrupt priority by default**.

## Semantics documented

- Omitted (or `null`) `priority` resolves to `"interrupt"` at the MCP binding layer.
- Explicit `priority: "queue"` is the opt-out, mapping to `"normal"` (queue-if-busy).
- Every other explicit value passes through unchanged.
- **Wire defaults unchanged**: `agent.sendMessage` / `agent.sendToTask` RPC defaults for an omitted `priority` (FE front door, internal wakes, automated deliveries) remain queue-if-busy.
- Existing interrupt constraints unaffected: single-pending-message guard (no interrupt bypass), question hold still parks default-interrupt A2A sends front-of-queue with `interruptPriority: true`, and the turn-startup fallback still queues keep-alive instead of preempting.

## Changes

- `docs/protocol/methods/agents.md` — new "Interrupt-by-default A2A sends" block in §5.5; binding-default notes on the `agent.sendMessage` / `agent.sendToTask` catalog rows; Question-hold "What is held" paragraph updated.
- `docs/protocol/07-agent-streaming.md` — "Sending input" note distinguishing the wire default from the MCP binding default.
- `docs/protocol/versioning.md` — behavior-only entry within 7.0 (MCP-binding-local, no method-catalog or wire-shape change, no version bump).

No submodule pin changes.


docs/protocol/methods/agents.md
```diff
@@ -567,7 +581,11 @@ wakes, `agent.reportToParent` immediate wakes), event-subscription batch deliver
 `agent.sendToTask`, and `agent.wakeOrCreate` context messages. Automatic **interrupt-priority**
 sends are held too — no exceptions — parking front-of-queue with `interruptPriority: true` (behind
 existing interrupt entries, ahead of normal ones; the preemption is skipped entirely — there is
-nothing to preempt, since a hold implies the question-asking turn already ended). A held park
+nothing to preempt, since a hold implies the question-asking turn already ended). Since
+[intentd#1292](https://github.com/intent-hq/intentd/pull/1292) the `ws.agent.send` /
+`ws.agent.sendToTask` bindings default an omitted `priority` to `"interrupt"`, so a default A2A
+send under hold parks front-of-queue this way — only the explicit `priority: "queue"` opt-out
```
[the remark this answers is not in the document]
panghy: Fixed in 0bbe67c — the sentence now says an explicit non-interrupt priority (the "queue" opt-out, or any other value, since only "interrupt" is interrupt at the service layer) parks as a normal entry.

docs/protocol/methods/agents.md
```diff
@@ -567,7 +581,11 @@ wakes, `agent.reportToParent` immediate wakes), event-subscription batch deliver
...
+parks as a normal entry. A held park
```
panghy: "only the explicit `priority: "queue"` opt-out parks as a normal entry" — the "only" is too strong and contradicts the §5.5 block added in this same PR ("any non-`"interrupt"` value is non-interrupt at the service layer, and every other explicit value passes through unchanged"). The binding maps `"queue"` → `"normal"`, but any other explicit non-interrupt value (e.g. `priority: "normal"`) passes through unchanged and equally parks as a normal entry under the hold.

Suggested minimal fix: "— an explicit non-interrupt `priority` (e.g. the `"queue"` opt-out) parks as a normal entry."

panghy (commented):
Verified this docs-only PR against the merged intent-hq/intentd#1292 implementation (the `effective_priority` helper in `crates/intent-acp/src/mcp_server/bindings/agent.rs` on main):

- ✅ §5.5 "Interrupt-by-default A2A sends" block matches the helper exactly: omitted/`null` → `"interrupt"`, explicit `"queue"` → `"normal"`, every other explicit value passes through unchanged; wire-level `agent.sendMessage` / `agent.sendToTask` RPC defaults untouched; existing interrupt constraints (single-pending-message guard, question-hold front-of-queue park, turn-startup keep-alive fallback) correctly stated as unchanged.
- ✅ `07-agent-streaming.md` §7 note correctly distinguishes the wire default (queue-if-busy) from the MCP binding default (interrupt), with the right §5.5 cross-reference.
- ✅ `versioning.md` 7.0 entry is consistent with both (behavior only, MCP-binding-local, no version bump).
- ✅ Only the three `docs/protocol/` files are touched — no submodule pins.

One wording inaccuracy in the question-hold paragraph flagged inline; everything else is accurate.

docs/protocol/methods/agents.md
panghy: Already fixed in 0bbe67c (pushed just before this review landed) — the sentence now reads: "an explicit non-interrupt `priority` (the `"queue"` opt-out, or any other value, since only `"interrupt"` is interrupt at the service layer) parks as a normal entry."

panghy (commented):
Re-verified at head `0bbe67c` — **approved from my side**. ✅

The question-hold paragraph now reads "an explicit non-interrupt `priority` (the `"queue"` opt-out, or any other value, since only `"interrupt"` is interrupt at the service layer) parks as a normal entry", which resolves the wording inaccuracy I flagged and is now fully consistent with the §5.5 "Interrupt-by-default A2A sends" block and the merged intentd#1292 `effective_priority` semantics. The fix commit touches only `docs/protocol/methods/agents.md`; the PR still changes exactly the three `docs/protocol/` files with no submodule pins, and both review threads are resolved.

Everything in this PR is accurate against the landed implementation. Not merging — leaving that to a human.

PR #2860, opened 2026-08-19
docs: document slim page byte budget on slim conversation reads

panghy:
Documents the slim page byte budget shipped in [intent-hq/intentd#1314](https://github.com/intent-hq/intentd/pull/1314).

## What

Under `projection: "slim"`, `agent.getConversation` pages (and the seq-0 `chat.subscribe` snapshot, which reuses the same read) are now additionally bounded by `SLIM_PAGE_BUDGET_BYTES` (512 KiB) of total serialized message bytes. The per-block 2 KiB budget caps each body, but `limit` counts messages and a message can carry hundreds of capped blocks — a real-world slim page carried 3,672 blocks and serialized to 2.87 MB. With the page budget, `limit` becomes a maximum: the read stops early, keeps the page's anchor end, always serves at least one message, and re-mints the continuation cursors (`nextToken` / `prevToken`) at the first excluded row so token walks resume seamlessly.

## Docs updated

- `docs/protocol/methods/agents.md` — `agent.getConversation` row: the slim page byte budget contract (anchor-end semantics, ≥1 message floor, cursor re-mint, frame ceiling, unchanged `totalMessages`/`truncated` semantics, full reads never budgeted).
- `docs/protocol/versioning.md` — "Also within 7.2 (behavior only, no method-catalog or wire-shape change …)" entry per the established convention.
- `docs/protocol/07-agent-streaming.md` — §7.1 slim projection paragraph: slim snapshots inherit the page budget via the shared conversation read.

Behavior only — no new wire fields, no method-catalog change, no version bump.


docs/protocol/methods/agents.md
```diff
@@ -9,7 +9,7 @@ The largest namespace. Every `agent.*` method is served daemon-primary by `inten
...
...ssage bytes — the per-block budget caps each body, but `limit` counts messages and a message can carry hundreds of capped blocks, so a message-counted slim page could still serialize to multiple MB. `limit` is therefore a **maximum** under `"slim"`: the read stops early once the page would exceed the budget, keeping the page's anchor end (newest for legacy backward pages, oldest for `prevToken` forward continuations, the target message — grown outward within budget — for seek pages) and always serving at least one message (a single over-budget message serves alone; never an empty page or a token loop). The continuation cursor(s) — `nextToken`, and `prevToken` on the seek path — are re-minted at the first excluded row, so an existing token walk resumes seamlessly and reconstructs the identical sequence as unbudgeted paging, just in more round-trips. Frame ceiling = budget + one message's slim size (well under the transport's 1 MiB large-frame warn). `totalMessages` / `truncated` semantics are unchanged (transcript-wide, not page-length), the seq-0 `chat.subscribe` snapshot inherits the bound automatically (it reuses this read), and full (absent-projection) reads are never budgeted |
```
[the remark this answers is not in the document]
panghy: Fixed in 6f31e1a. The "well under 1 MiB" claim was indeed wrong: the >=1-message floor means a single message with enough capped blocks can alone exceed the budget (and the transport's 1 MiB large-frame warn). The row now states the ceiling is budget + the slim size of the last admitted message, explicitly NOT a hard bound, with the worst case degrading to one whole message per page rather than a guaranteed frame size. This matches the implementation in intent-hq/intentd#1314, where the floor takes precedence to prevent empty pages and token loops.

docs/protocol/07-agent-streaming.md
```diff
@@ -102,7 +102,13 @@ observe the same bus, and `events.subscribe(["agent:stream:*"])` is unchanged.
   tool-block deltas, and the seq-0 live-turn merge — so slim snapshots and deltas agree on the
   served block shape. Absent / `null` keeps the subscription byte-identical to before (the
   additive convention); any other value is `-32602`, never coerced. A client holding a truncated
-  slim block fetches the full body on demand via `agent.getMessageBlock` (§5.5, v7.2).
+  slim block fetches the full body on demand via `agent.getMessageBlock` (§5.5, v7.2). Slim
+  snapshots additionally inherit the **slim page byte budget** (within v7.2 —
+  [intent-hq/intentd#1314](https://github.com/intent-hq/intentd/pull/1314)): the seq-0 and
+  lag-recovery snapshots reuse the `agent.getConversation` read, so a snapshot page is bounded
```
panghy, line 108: "so a snapshot page is bounded at `SLIM_PAGE_BUDGET_BYTES` (512 KiB)" holds for the persisted page, but the seq-0 snapshot then merges the in-flight live turn on top of the budgeted read (`merge_live_turn`), and `slim_message_blocks` caps each block (~2 KiB) without capping block count — so the emitted snapshot frame itself is not bounded at 512 KiB when a tool-heavy turn is in flight. This is flagged on intent-hq/intentd#1314 (existing review thread); depending on how that resolves (bounding the merged live blocks vs. scoping it out with a follow-up), this sentence should either stand as-is or gain a short qualifier like "the persisted page is bounded; an in-flight live turn merged on top is block-capped but not page-budgeted". Suggest holding this PR until that thread settles so the doc matches the landed behavior.

panghy (commented):
Verifier review of the docs against the intentd#1314 implementation (no merge action taken).

**Contract match — verified:** the documented semantics all hold against the implementation at intentd@2f87e85: `limit` becomes a maximum under slim; the anchor-end table (newest for legacy backward pages, oldest for forward continuations, target for seek pages) matches `BudgetAnchor` in `pagination.rs`; the at-least-one-message floor, cursor re-mint at the first excluded row, unchanged full (absent-projection) reads, and transcript-wide `totalMessages`/`truncated` semantics are all implemented as documented and covered by the pagination/agent_ops/wss tests (ran locally, all pass).

One wording caveat, tracking an open thread on the implementation PR — see the line comment below.

docs/protocol/07-agent-streaming.md
panghy: The intentd thread settled on bounding the merge: `merge_live_turn` now re-budgets the merged slim page (intentd commit 01accb7d on #1314) — the live turn anchors as the newest row (always served, per the §5.5 one-message floor), oldest persisted rows are evicted until the merged page fits `SLIM_PAGE_BUDGET_BYTES`, and `truncated`/`nextToken` are re-minted at the eviction boundary so evicted rows stay reachable via `agent.getConversation`.

The doc now documents exactly that: the sentence gained a follow-on stating the budget covers the live-turn merge too, including the anchor/eviction/re-mint semantics. Pushed as a new commit on this PR.

PR #2894, opened 2026-08-19
fix(make): rename run-fe to dev-fe paired with dev-daemon, wire SEED_SOURCE_DB

panghy:
## Summary

Audit of the monorepo Makefile found two real discrepancies (all other targets verified valid against current submodule sources); this PR fixes both plus a requested rename:

- **`run-fe` → `dev-fe`**: the FE's zero-config default changed from `ws://127.0.0.1:5181/ws` (`DEFAULT_DEV_WS_URL`, since removed) to UDS at the installed daemon's socket, so a bare `make run-fe` silently stopped pairing with `make dev-daemon`. The new `dev-fe` pre-checks `$(DEV_DATA_DIR)/intentd.sock` (with guidance to start `dev-daemon`) and launches the FE with `INTENTD_SOCKET` pinned to the dev seat socket, via a new internal `fe-launch` helper shared with `dev-prod` (whose behavior is unchanged). `make dev` remains the one-command isolated mode; `dev-fe` + `dev-daemon` are the two-terminal counterparts.
- **`SEED_SOURCE_DB` wiring**: the `seed-dev-workspaces` comment advertised `SEED_SOURCE_DB=/path/to/intentd.db` but the recipe never forwarded it; it now passes `--source` via `$(if ...)` conditionals (no flag when unset).
- **Stale comment refresh**: no `DEFAULT_DEV_WS_URL` / ws-default claims remain; `DEV_TCP_PORT` is documented as serving iOS (and explicit `INTENTD_WS_URL` clients); README's two-terminal mode synced to `make dev-daemon` + `make dev-fe`.

## Verification

- `make -n dev-fe` / `make -n dev-prod` / `make -n seed-dev-workspaces` (all `SEED_SOURCE_DB` × `SEED_INCLUDE_ARCHIVED` combinations) parse and show the expected recipes
- `grep` clean: no `DEFAULT_DEV_WS_URL`, no bare `run-fe` (only `run-fe-local`) in Makefile or README
- `make help` lists `dev-fe` accurately; `fe-launch` stays internal/unlisted
- Comments cross-checked against `resolveBackendConfig` in `packages/cloudlands-fe/src/features/backend/main/backend-connection.ts` (`INTENTD_SOCKET` > `INTENTD_WS_URL` > `INTENTD_TCP` > UDS default honoring `INTENTD_DATA_DIR`)
- Two independent verification passes approved the changes; only `Makefile` + `README.md` modified (no submodule edits)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@
 # Four postures for local work:
 #   1. `make dev-daemon` — default dev seat. intentd on an isolated data dir,
 #      serving UDS + insecure TCP (`--insecure`) bound on 0.0.0.0:$(DEV_TCP_PORT)
-#      (reachable from loopback and LAN). Pairs with `make run-fe` (whose dev default
-#      is `ws://127.0.0.1:5181/ws`) and with the iOS app (`make ios-info`).
+#      (reachable from loopback and LAN). Pairs with `make dev-fe` (which pins
+#      the FE to the dev seat's UDS socket) and with the iOS app (`make ios-info`).
 #   2. `make release-daemon` — occasional "debug the release app with its own
 #      state" seat. intentd on the real data dir, UDS-always and no
 #      `--insecure`. No TCP port is bound unless the persisted
@@ -12,7 +12,7 @@
 #      listener binds `server.wsApi.port` (default 5181 — the same as
 #      $(DEV_TCP_PORT)); if the dev seat already holds it, the bind failure
 #      is non-fatal and UDS keeps serving.
-#   3. `make run-fe` / `make ios-open` / `make ios-info` — clients pointed at
+#   3. `make dev-fe` / `make ios-open` / `make ios-info` — clients pointed at
 #      the dev daemon.
 #   4. `make dev-prod` — FE dev build connected to the already-running packaged
 #      app daemon, showing the same workspaces and agents without starting a
@@ -39,10 +39,13 @@ SUBMODULES = $(INTENTD_DIR) $(FE_DIR) $(IOS_DIR)
 # seat never touches the packaged app's real data dir (Config::resolve in
 # packages/intentd/crates/intent-core/src/config.rs honours INTENTD_DATA_DIR).
 #
-# DEV_TCP_PORT is the intentd TCP/WebSocket base port. 5181 matches the FE's
-# built-in dev default (`DEFAULT_DEV_WS_URL = ws://127.0.0.1:5181/ws` in
-# packages/cloudlands-fe/src/features/backend/main/backend-connection.ts), so
-# `make dev-daemon` + `make run-fe` connect with no env overrides. The daemon
+# DEV_TCP_PORT is the intentd TCP/WebSocket base port for the dev seat's
+# insecure plain-ws:// listener. It serves the iOS app (`make ios-info`) and
+# any client pointed at it explicitly via INTENTD_WS_URL — not the FE's
+# default: the FE's zero-config default is UDS at the platform data-dir
+# socket, honoring INTENTD_DATA_DIR (resolveBackendConfig in
+# packages/cloudlands-fe/src/features/backend/main/backend-connection.ts),
+# and `make dev-fe` pins the FE to the dev seat's UDS socket. The daemon
 # reads it via the INTENTD_TCP_PORT env seam (ws_options_from_env in
 # packages/intentd/crates/intentd/src/main.rs).
 #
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -356,23 +358,36 @@ run-intentd: ## DEPRECATED alias for release-daemon
...
+	# The pin is needed because the FE's zero-config default is UDS at the
+	# PLATFORM data-dir socket (honoring INTENTD_DATA_DIR) — i.e. the installed
+	# daemon, not the dev seat. The FE does NOT spawn intentd; start
+	# `make dev-daemon` in another terminal first, or use `make dev` for the
+	# one-command sidecar mode. Other clients:
+	#   `make run-fe-local` → dev FE against the installed daemon's UDS socket.
+	#   `make dev-prod`     → same, with the packaged app's live daemon.
 	# Long-running; does not exit until you stop it (Ctrl-C).
-	@[ -d $(FE_DIR)/node_modules ] || (echo "[run-fe] installing FE deps (pnpm install)" && cd $(FE_DIR) && pnpm install)
+	@if [ ! -S "$(DEV_DATA_DIR)/intentd.sock" ]; then \
```
panghy: The new pre-check is a Windows (MSYS/MINGW) regression relative to old `run-fe`, which had no pre-check: `run-fe-local` below deliberately skips its `[ ! -S ... ]` test on `MINGW*|MSYS*|CYGWIN*` because the daemon socket is "not a filesystem entry" there (lines 394, 410-426), so on Windows this check will fail even with `dev-daemon` running, making `dev-fe` unusable. Suggest mirroring `run-fe-local`'s `is_windows` guard (skip the `-S` test on those platforms), or — if the dev seat is considered macOS/Linux-only — saying so in the recipe comment so the hard failure is at least documented.

panghy (commented):
Reviewed the full diff against the current Makefile, README, and the FE source the comments cite. Verification performed:

- `make -n seed-dev-workspaces` in all four `SEED_SOURCE_DB` x `SEED_INCLUDE_ARCHIVED` combinations — the `$(if ...)` conditionals emit exactly the right flags, and `--source "$(SEED_SOURCE_DB)"` stays shell-quoted (verified with a space-containing path).
- `make -n dev-fe` / `make -n dev-prod` — `INTENTD_SOCKET` propagates correctly through the `$(MAKE) fe-launch` sub-make (command-scope env inherited by the child's recipe shells; `export DEV_PORT` still applies), and `fe-launch`'s body is byte-for-byte the old `run-fe` recipe, so `dev-prod` is behaviorally unchanged.
- `.PHONY` updated correctly (`run-fe` removed; `dev-fe` and `fe-launch` added).
- `grep` clean: no residual `run-fe` (other than `run-fe-local`) or `DEFAULT_DEV_WS_URL` in Makefile/README; `make help` lists `dev-fe` with an accurate description and `fe-launch` stays unlisted.
- Comment accuracy cross-checked against `resolveBackendConfig` in `packages/cloudlands-fe/src/features/backend/main/backend-connection.ts`: `INTENTD_SOCKET` is indeed the highest-precedence override and the zero-config default is UDS at the platform data dir honoring `INTENTD_DATA_DIR`. README's two-terminal line matches the new behavior.

One actionable issue posted inline (Windows portability of the new `dev-fe` socket pre-check). One non-blocking observation, no change requested: `[ -S ]` passes on a stale socket file left by a dead daemon, so the pre-check can green-light a connection that then fails in the FE — same property as `run-fe-local`'s existing check, so consistent with the file's conventions.

Makefile
panghy: Fixed in f5b390f: `dev-fe` now mirrors `run-fe-local`'s `is_windows` guard — `uname -s` matching `MINGW*|MSYS*|CYGWIN*` skips the `-S` pre-check (the daemon socket path is not a filesystem entry on Windows; the win32 FE derives a named pipe from it and reports the connection failure itself). The macOS/Linux pre-check with the `make dev-daemon` guidance is unchanged, and a comment references the same rationale as `run-fe-local`.

Makefile
```diff
@@ -356,23 +358,36 @@ run-intentd: ## DEPRECATED alias for release-daemon
...
-	# intentd; pair this with `make dev-daemon` (default) — the FE's dev build
-	# defaults to `ws://127.0.0.1:5181/ws` (DEFAULT_DEV_WS_URL in
-	# packages/cloudlands-fe/src/features/backend/main/backend-connection.ts),
-	# which matches dev-daemon's DEV_TCP_PORT. Overrides:
-	#   INTENTD_SOCKET=/path/to.sock  → UDS (highest precedence; e.g. point at
-	#                                    release-daemon's default socket).
-	#   INTENTD_WS_URL=ws://host:port → plain WebSocket to a specific URL.
-	#   `make run-fe-local`           → shorthand that resolves INTENTD_SOCKET to
-	#                                    the installed daemon's default UDS path.
+dev-fe: ensure-fe-submodule ## Run the FE dev stack against dev-daemon's UDS socket (two-terminal pair)
```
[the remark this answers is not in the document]
panghy: Fixed in 1e4cad2: the `.intent/config.json` dev service entry now points at the renamed target (name `dev-fe`, command `make dev-fe`). JSON validity checked with `python3 -m json.tool`, and `git grep -n "run-fe" -- ':!packages'` confirms no other bare `run-fe` references remain (only `run-fe-local`, which is a distinct, still-existing target). The deprecated `run-intentd` entry is left as-is since that target still exists as an alias.

PR #2943, opened 2026-08-20
docs: document queueInfo.batchId batch-flush stamp in protocol contract

panghy:
Documents the new `queueInfo.batchId` field in the canonical wire contract (docs-only; no code).

## Changes

`docs/protocol/methods/agents.md`, dequeue-wait annotation section (the `queueInfo { queuedAt, waitedMs }` stamp):

- Adds `batchId` to the `queueInfo` shape as an **optional string**: present on every user row drained together in one multi-message batch flush ("Queued-message flush" section), with all rows of one flush sharing the same value so clients can group the stacked rows; absent on single-entry drains and on rows persisted by older daemons.
- Notes that entries whose wait fell below the 5-second annotation threshold still carry a `queueInfo` containing only `batchId` (no `queuedAt`/`waitedMs`, no content note) when they are part of a batch flush.
- Updates the "Per-entry transcript rows" bullet in the queued-message-flush section to match: each row's `queueInfo` stamp carries the shared `batchId`, plus `queuedAt`/`waitedMs` when the wait met the threshold.

### DIFF
--- a/docs/protocol/methods/agents.md
+++ b/docs/protocol/methods/agents.md
@@ -385,12 +385,21 @@ sends. **MCP-only surface changes** (§6.8 principle) — no new wire methods; t
   below the **5-second annotation threshold** (monorepo#2353): a sub-threshold hop —
   e.g. a question-wizard answer converted into an enqueue + immediate drain by the #1791
   FIFO-restore branch — is treated like an immediate delivery (no note, no `queueInfo`
-  stamp), so instant queue hops never render a "waited 0s" chip. Alongside the content
+  stamp — except that a sub-threshold entry drained as part of a multi-message batch flush
+  still carries a `queueInfo` holding only `batchId`, below), so instant queue hops never
+  render a "waited 0s" chip. Alongside the content
   note, the drained entry's `messageMetadata` is stamped with structured queue info —
-  `queueInfo: { "queuedAt": "<ISO enqueue timestamp>", "waitedMs": <non-negative millis> }`
+  `queueInfo: { "queuedAt": "<ISO enqueue timestamp>", "waitedMs": <non-negative millis>,
+  "batchId": "<opaque id>"? }`
   — persisted on the user transcript row and round-tripping on chat reads
   (`agent.getConversation` / `chat.subscribe`) like the A2A sender-attribution metadata, so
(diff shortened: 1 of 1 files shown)


panghy:
**Verifier review — one stale sentence now contradicts the new stamp**

The added §5.5 dequeue-wait text and the "Per-entry transcript rows" bullet accurately match the intent-hq/intentd#1335 implementation (`stamp_flush_batch_id`): shared value per >=2-entry flush, absent on single-entry drains, batchId-only `queueInfo` for sub-threshold entries, existing `queueInfo` fields never overwritten, `persisted: true` never stamped.

However, the **Turn correlation (monorepo#1022)** bullet in the same "Queued-message flush" section still says:

> The queue entries themselves keep their own `turnId`s (entry ids / `messageMetadata` / `queueInfo` are untouched).

That parenthetical is now contradicted by the batch stamp, which mutates each drained entry's `messageMetadata.queueInfo` (adds `batchId`) before the rows persist. intent-hq/intentd#1335 updated the equivalent Rust doc comment on `prepare_flush_turn` for exactly this reason:

> Queue entries keep their own `turn_id`s and ids; the only metadata mutation beyond the single-drain annotations is the shared `queueInfo.batchId` grouping stamp (`stamp_flush_batch_id`) on every entry of the batch.

Suggest mirroring that wording, e.g.: "The queue entries themselves keep their own `turnId`s and ids; the only metadata mutation beyond the single-drain annotations is the shared `queueInfo.batchId` grouping stamp on every entry of the batch."

panghy:
Re: https://github.com/intent-hq/monorepo/pull/2943#issuecomment-5352499189 — agreed, the "Turn correlation (monorepo#1022)" parenthetical ("entry ids / `messageMetadata` / `queueInfo` are untouched") contradicted the `batchId` stamp documented two bullets earlier.

This PR was already merged and its branch deleted, so the fix is in a follow-up PR: #2949. It rewords the parenthetical to: "keep their own `turnId`s and ids; the only metadata mutation beyond the single-drain annotations is the shared `queueInfo.batchId` grouping stamp on every entry of the batch" — matching the intent-hq/intentd#1335 doc-comment update on `prepare_flush_turn`.

PR #2966, opened 2026-08-20
docs: document agent-flipped completion wake triggers

panghy:
# Document agent-flipped completion wake triggers

Protocol docs for [intent-hq/intentd#1340](https://github.com/intent-hq/intentd/pull/1340) — agent-flipped completions (typically a verifier flipping other task notes to `complete`) now join the wake's `unblockedTriggerTasks` trigger set, so the delivery-time "tasks now unblocked" hint fires for verifier-verified work.

## Changes

- `docs/protocol/methods/agent-aux.md` — the "Delivery-time tasks now unblocked hints" section's enqueue-time bullet now covers the flipped-completion record/stamp/consume contract (persisted `agent_flipped_completion` table, cap/dedup/exclusion rules, consume-on-stamp at all three wake-composition sites, and the multi-pair group-event stamp with legacy single-object back-compat), and the citation line references intentd#1340.
- `docs/protocol/versioning.md` — adds the "Also within 7.2 (behavior + persisted-store change only, no method-catalog or wire-shape change; so no version bump)" entry for the same change, per the file's existing convention.

Behavior + persisted-store change only — the `unblockedTriggerTasks` metadata shape is unchanged and there is no method-catalog change, so no protocol version bump.

Docs-only monorepo PR; the implementation lands via [intent-hq/intentd#1340](https://github.com/intent-hq/intentd/pull/1340) (intentd merges first per the protocol-change ordering rule).


panghy (commented):
Reviewed `docs/protocol/methods/agent-aux.md` and `docs/protocol/versioning.md` against the intent-hq/intentd#1340 implementation.

**Verdict: accurate, no issues found.**

Checked point by point:

- The described trigger sources match the implementation: the completing child's own linked task plus its recorded agent-flipped completions, joined and deduped at stamp time.
- Recording semantics match: only status writes crossing INTO `complete` on a task note OTHER than the caller agent's own linked task are recorded; user-initiated (caller-less) writes record nothing; a transition back out of `complete` removes the pair for every recording agent.
- Consume-on-stamp is correctly described — a flip is attributed to at most one wake (ungrouped watch wake, `agent.reportToParent` wake, or captured at group record time for `after_all` aggregated wakes) and never re-attributed by a later completion cycle.
- Best-effort/advisory framing matches the fail-open implementation (recording and stamping never fail the underlying status write or wake delivery).
- Cap (50, oldest-first eviction) and persistence-across-restart behavior match migration 0100 + the repo.
- `versioning.md`: correct that no protocol version bump is needed — the trigger metadata extends an existing advisory stamp shape (array-of-pairs with a legacy single-object reader fallback in the daemon); no wire-shape break.

CI green (bridge-test, pr-title). Note the usual ordering rule applies: this docs PR describes behavior landing in intentd#1340, so the daemon PR should merge first.

Issue #2857, opened 2026-08-19
Design: agent-scoped browser tabs — per-agent ownership, claiming, and scoped listTabs


panghy:
## Problem

Multiple agents (and the user) share a single pool of browser tabs per workspace. Dogfooding on 2.6x surfaced the failure class repeatedly:

- Agent validation runs open tabs into the user's live session (tab pollution).
- Hidden offscreen webviews stole the app-level DevTools toggle (#2844) because background guests are invisible and unattributed.
- URL-based dedupe has to guess whether an existing tab "belongs" to the caller; scoping was ambiguous by construction.

Fundamentally: tabs have no owner, so every browser op is a potential cross-agent conflict.

### Model
- Every tab carries a nullable `ownerAgentId`. User-opened tabs start **unowned**.
- Agents can only manipulate (navigate/close/evaluate/screenshot/...) tabs they own. Other agents' tabs are read-visible but not manipulable; an agent wanting the same page opens its own tab with the same URL.
- **Claiming**: an agent may claim an unowned tab and gains exclusive access. Claims are atomic — first claim wins, the loser receives a structured `already-claimed` error naming the owner. No stealing. Unowned tabs can only originate from users.
- **Lifecycle**: owned tabs' webviews are always active for the agent's lifetime. Tabs keep ownership when the agent completes; they are discarded only when the agent is deleted (or the workspace is archived/deleted).
- **Dedupe** becomes per-agent: same `requestedUrl` within an agent reuses; across agents, each agent gets its own tab.
- **listTabs scoping**: `scope: 'mine' | 'unclaimed' | 'all'` on the request; every returned tab carries `ownerAgentId` + owner display info. (Supersedes an earlier two-boolean sketch: `currentAgent` / `userTabs`.)

### Protocol (intentd first)
- `browser.exec` requests carry the initiating `agentId` (the daemon knows the caller; the FE currently does not).
- New/extended ops: `claimTab`; `listTabs(scope)`; structured ownership errors (`not-owner`, `already-claimed`).
- `docs/protocol/` is the canonical contract; the daemon PR merges before the FE PR per the cross-component rule.

### UI
- Browser panel header: `Browser — [agent chip]`; clicking the chip navigates to the owning agent.
- Sidebar: tabs grouped by owner agent (collapsible, expanded by default), plus an "Unclaimed" group for user-opened tabs; per-tab liveness dot + title + URL as today.
[...]

PR #2995, opened 2026-08-20
docs(protocol): state §5.9 viewport sizing bounds 320–3840 px

panghy:
Aligns the §5.9 viewport sizing prose with the implemented FE validation bounds (refs #2857; flagged by the fe#1458 verifier).

## What changed

`docs/protocol/methods/files-terminal-browser.md`, §5.9 **Viewport sizing invariant** block only:

- `width` / `height` (everywhere sizes are accepted: `claimTab` / `openTab` / `resizeTab`) are now stated as **integers within 320–3840 CSS px, inclusive** — previously the prose said only "positive integers", while the merged FE `ViewportDimensionSchema` (fe#1458) enforces `z.number().int().min(320).max(3840)`.
- Out-of-range behavior is stated explicitly: fractional, non-finite, or out-of-range values are validation errors reported inside the per-action `{ action, success: false, error }` result envelope.

No other contract semantics changed.

## Verification

- Cross-checked against the merged FE schema: `src/features/browser/main/browser-action-executor.ts` — `const ViewportDimensionSchema = z.number().int().min(320).max(3840);`, applied to `openTab` (both optional), `claimTab` (`width` required), and `resizeTab` (`width` required).
- Docs-only change; diff is strictly the §5.9 sizing prose.

### DIFF
--- a/docs/protocol/methods/files-terminal-browser.md
+++ b/docs/protocol/methods/files-terminal-browser.md
@@ -262,10 +262,13 @@
 > emulation, so owned tabs render deterministically offscreen without disturbing the
 > user's panel layout. There is no opt-in and no clear/reset op. Everywhere they appear
 > (`claimTab` / `openTab` / `resizeTab`), `width` and `height` are **integers within
-> 320–3840** (CSS px, inclusive): fractional, non-finite, or out-of-range values
-> (below 320 — which covers zero and negatives — or above 3840) are validation
-> errors, reported inside the per-action `{ action, success: false, error }` result
-> envelope — an emulated tab can never carry a disabling zero size. Agent-issued
+> 320–3840** (CSS px, inclusive), enforced up-front by action-sequence schema
+> validation: fractional, non-finite, or out-of-range values (below 320 — which
+> covers zero and negatives — or above 3840) reject the **whole batch before any
+> action executes**, surfacing through the FE top-level failure envelope
+> (`success: false` + top-level `error`, `Invalid action sequence: …`) that the
+> daemon maps to `-32603` (see above) — an emulated tab can never carry a
+> disabling zero size. Agent-issued
 > `openTab` accepts optional `width` / `height` and the tab is emulated from creation;
 > omitted `width` defaults to **1280** and omitted `height` to **800** (the standard
 > desktop viewport 1280×800 when both are omitted), and `claimTab` with omitted
--- a/docs/protocol/methods/files-terminal-browser.md
+++ b/docs/protocol/methods/files-terminal-browser.md
@@ -261,9 +261,11 @@
 > tabs are **always emulated** — the FE applies the size as CDP device-metrics viewport
 > emulation, so owned tabs render deterministically offscreen without disturbing the
 > user's panel layout. There is no opt-in and no clear/reset op. Everywhere they appear
-> (`claimTab` / `openTab` / `resizeTab`), `width` and `height` are **positive
-> integers** (CSS px): zero, negative, fractional, or non-finite values are validation
-> errors — an emulated tab can never carry a disabling zero size. Agent-issued
+> (`claimTab` / `openTab` / `resizeTab`), `width` and `height` are **integers within
+> 320–3840** (CSS px, inclusive): fractional, non-finite, or out-of-range values
+> (below 320 — which covers zero and negatives — or above 3840) are validation
+> errors, reported inside the per-action `{ action, success: false, error }` result
+> envelope — an emulated tab can never carry a disabling zero size. Agent-issued
 > `openTab` accepts optional `width` / `height` and the tab is emulated from creation;
 > omitted `width` defaults to **1280** and omitted `height` to **800** (the standard
 > desktop viewport 1280×800 when both are omitted), and `claimTab` with omitted


docs/protocol/methods/files-terminal-browser.md
```diff
@@ -261,9 +261,11 @@
...
 > emulation, so owned tabs render deterministically offscreen without disturbing the
 > user's panel layout. There is no opt-in and no clear/reset op. Everywhere they appear
-> (`claimTab` / `openTab` / `resizeTab`), `width` and `height` are **positive
-> integers** (CSS px): zero, negative, fractional, or non-finite values are validation
-> errors — an emulated tab can never carry a disabling zero size. Agent-issued
+> (`claimTab` / `openTab` / `resizeTab`), `width` and `height` are **integers within
+> 320–3840** (CSS px, inclusive): fractional, non-finite, or out-of-range values
+> (below 320 — which covers zero and negatives — or above 3840) are validation
+> errors, reported inside the per-action `{ action, success: false, error }` result
+> envelope — an emulated tab can never carry a disabling zero size. Agent-issued
```
panghy: The bounds themselves are correct, but the delivery claim does not match the merged implementation. `ViewportDimensionSchema` is enforced only through the sequence-level `ActionSequenceSchema.safeParse` at the top of `executeActions` (`browser-action-executor.ts:1166`), before any action runs. On a violation the FE returns the **top-level** failure envelope `{ success: false, results: [], error: "Invalid action sequence: actions.N.width: …" }` — `results` is empty, so no per-action `{ action, success: false, error }` envelope ever exists for these errors. The fe#1458 tests confirm this (`browser-action-executor.test.ts:2591` — `rejects a non-positive-integer width via schema validation` for `0 / -100 / 1.5 / NaN` asserts top-level `result.success === false` with the CDP call never made).

Daemon-side, `shape_result` (intentd `crates/intent-services/src/browser_ops.rs:132-141`) maps any top-level `success: false` envelope to JSON-RPC `-32603` (`browser.exec failed: …`) before it ever looks at `results` — consistent with the earlier §5.9 prose "an FE-reported failure envelope all surface as `-32603`". So out-of-range / fractional / non-finite sizes actually reach the caller as a `-32603` JSON-RPC error carrying `Invalid action sequence: …`, not inside a per-action result envelope.

Suggested rewording: state that sizes are validated up-front as part of action-sequence schema validation, and that violations reject the whole batch before execution, surfacing through the FE top-level failure envelope → `-32603` path already described earlier in this section. (If per-action delivery is the *intended* contract, that is an FE behavior change and shouldn't be documented as current behavior by a PR whose stated purpose is aligning prose with the implemented validation.)

Non-blocking, pre-existing observation while tracing this: the ownership-errors bullet above claims ownership failures are "never" reported as the FE's top-level failure envelope, but `executeActions` wraps *any* failing action in a top-level `success: false` envelope (`browser-action-executor.ts:1192-1198`), which `shape_result` likewise maps to `-32603`. That prose predates this PR and is out of its scope, but it may deserve its own look — flagging since this PR's wording leans on the same delivery description.

panghy (commented):
Reviewed against the merged FE `ViewportDimensionSchema` (cloudlands-fe#1458, merged 2026-08-19, merge commit `d436ba6c`) as the source of truth.

**Verified**
- **Bounds match the merged schema**: `z.number().int().min(320).max(3840)` (`src/features/browser/main/browser-action-executor.ts:130`), applied at `openTab` (both optional, l.148-149), `claimTab` (`width` required, `height` optional, l.161-162), `resizeTab` (`width` required, `height` optional, l.173-174). The new "integers within 320–3840 (CSS px, inclusive)" prose is exact, and the required/optional split stated elsewhere in §5.9 (claimTab bullet, resizeTab shape, openTab optional sizes) is unchanged and consistent.
- **Internal consistency**: defaults 1280×800 are within bounds; "below 320 — which covers zero and negatives" correctly preserves the old zero/negative rejection and the "never a disabling zero size" invariant.
- **Scope**: the diff is strictly the §5.9 sizing prose — one hunk, one file, no other contract semantics touched.

**One substantive issue** (inline thread): the claim that these validation errors are "reported inside the per-action `{ action, success: false, error }` result envelope" does not match the merged FE + daemon behavior — they surface as a JSON-RPC `-32603` via the FE top-level failure envelope. Details and suggested rewording in the inline comment.

Not approving/merging per instructions — leaving as comments only.

docs/protocol/methods/files-terminal-browser.md
panghy: Confirmed against the source — `ViewportDimensionSchema` is only reached via `ActionSequenceSchema.safeParse` at the top of `executeActions` (`browser-action-executor.ts:1166`), which returns the top-level `{ success: false, results: [], error: "Invalid action sequence: …" }` envelope before any action runs, and `shape_result` maps that to `-32603`. Fixed in 6fae84b: the prose now states the bounds are enforced up-front by action-sequence schema validation, violations reject the whole batch before any action executes, and they surface through the FE top-level failure envelope → `-32603` path already described earlier in this section.

On the non-blocking observation about the ownership-errors bullet's "never top-level envelope" claim: agreed it deserves a look, but it's pre-existing prose outside this PR's scope (strictly the sizing prose), so leaving it unchanged here.

panghy (commented):
Verified against the merged FE implementation on cloudlands-fe `main` — approving.

**Bounds** — `src/features/browser/main/browser-action-executor.ts:130` on `main`: `const ViewportDimensionSchema = z.number().int().min(320).max(3840);`, applied to `openTab` (`width`/`height` optional, L148-149), `claimTab` (`width` required, L161-162), and `resizeTab` (`width` required, L173-174). The prose "integers within 320–3840 (CSS px, inclusive)" everywhere sizes are accepted matches exactly.

**Error delivery** — matches the merged behavior after the round-1 fix (6fae84b): `ViewportDimensionSchema` is reached only via `ActionSequenceSchema.safeParse` at the top of `executeActions` (L1166), which returns the top-level `{ success: false, results: [], error: "Invalid action sequence: …" }` envelope before any action runs; intentd `shape_result` (`crates/intent-services/src/browser_ops.rs`) maps any top-level `success: false` envelope to `-32603`. The prose's up-front-validation / whole-batch-rejection / top-level-envelope → `-32603` wording is accurate, and the "(see above)" reference resolves to the existing failure-envelope → `-32603` prose earlier in the section.

**Scope** — diff is a single hunk in `docs/protocol/methods/files-terminal-browser.md` (where README §5.9 delegates), touching only the Viewport sizing invariant block. No other contract semantics changed; the pre-existing ownership-errors prose flagged in review round 1 is correctly left out of scope.

CI green (pr-title + bridge-test) on head 6fae84b; the one review thread is resolved.

PR #3335, opened 2026-08-24
docs: specialist replacement mode and specialists.default Run behavior

panghy:
Documents the specialist base-tier replacement directory and the `specialists.default`-driven Run behavior in the canonical wire contract. Docs only — no code.

## Changes

**§5.11 (`docs/protocol/methods/misc-namespaces.md`)** — new "Base-tier replacement mode" paragraph:
- The effective `specialists.dir` setting (the `INTENTD_SPECIALISTS_DIR` startup pin, else a hand-written `[specialists] dir` in config.toml; the `intentd serve --specialists-dir` flag folds into the env var pre-runtime, so the flag wins; empty counts as unset) wholesale-replaces the base tier.
- The embedded bundle and the on-disk bundled `resources/specialists/` directory are both excluded; the named directory becomes the sole base (`bundled`, read-only) tier, and the user/project tiers fold on top unchanged.
- A missing or empty replacement directory yields an empty base tier; session bundle pins never resurrect excluded shipped bundles.

**§5.12 (`docs/protocol/methods/settings.md`)** — settings reference:
- Added `specialists.dir` to the Providers/agents key list.
- `specialists.default` documented as a client-consumed setting (the daemon stores and serves it but never consults it) read by the desktop FE's task-note Run path: Run spawns the configured specialist when it resolves to a pickable (visible, not `hidden`) roster entry, with the `implementor` fallback when unset or unresolvable.
- `specialists.dir` documented as read-only on the wire with its pin/file value sources.

The dropped `default:` frontmatter key is deliberately not documented.

## Verification

No docs build/lint tooling exists in the repo; reviewed against the intentd implementation (`specialists.rs` replacement-dir handling, `main.rs` startup-pin/flag fold, settings schema/catalog) and the FE `runAgentForNote` saga for accuracy, and against existing §5.11/§5.12 prose style.


docs/protocol/methods/settings.md
```diff
@@ -50,7 +50,7 @@ the overriding flag ("overridden by startup flag …").
...
...d at boot, and a legacy `config.toml` key is still tolerated + stripped on boot with its value discarded. `specialists.default` *(optional string, TOML-backed under `[specialists]` as `default`; default: unset)* is the id of the specialist applied when none is chosen — a **client-consumed** setting: the daemon stores and serves it but never consults it when creating agents. The desktop FE's task-note **Run** path reads it: Run spawns the note's agent as the configured specialist when the id resolves to a pickable entry in the resolved specialist roster (visible to the client and not `hidden`, §5.11), and falls back to `implementor` when the key is unset or does not resolve. `specialists.dir` *(optional string, **read-only on the wire** — `settings.update` / `settings.reset` yield `-32602`)* reports the base-tier replacement directory for specialist resolution (§5.11 "Base-tier replacement mode"): it takes a value via the `INTENTD_SPECIALISTS_DIR` startup pin (or the `intentd serve --specialists-dir` flag, which folds into the env var pre-runtime — the flag wins over an inherited env value), else a hand-written `[specialists] dir` in config.toml; unset or empty means no replacement.
```
panghy: **[medium]** The claim that `settings.reset` on `specialists.dir` yields `-32602` doesn't match the implementation: only `settings.update` has a read-only guard (`settings.rs` `update()` rejects `def.read_only` with `-32602`), while `reset()` has no such check — it rejects `-32602` only when the key is startup-pinned (registry pin check), and a file-written `[specialists] dir` (or an unset key) resets successfully, removing the key from config.toml. Suggest rewording to: `settings.update` yields `-32602`; `settings.reset` yields `-32602` only while the value is startup-pinned, else it clears a file-written value (or file an intentd fix to guard `reset` and keep this wording).

panghy (commented):
Reviewed against the merged implementations (intent-hq/intentd#1435, intent-hq/cloudlands-fe#1640). The §5.11 replacement-mode paragraph and the `specialists.default` Run-path description are accurate, internally consistent with the existing §5.11 watching/tier prose, and correctly omit the dropped `default:` frontmatter key. One accuracy issue on the `specialists.dir` read-only claim, noted inline.

docs/protocol/methods/settings.md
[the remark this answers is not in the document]
panghy: This is stale against intentd main: [intent-hq/intentd#1435](https://github.com/intent-hq/intentd/pull/1435) (merged) adds `specialists.dir` to `KNOWN_PATHS` and the settings definitions, and introduces the `INTENTD_SPECIALISTS_DIR` replacement seam (`REPLACEMENT_DIR_ENV` in `specialists.rs`) alongside the pre-existing `INTENTD_BUNDLED_SPECIALISTS_DIR` override. The protocol docs document intentd main as merged; the monorepo pin advances automatically via the auto-bump-submodules workflow.

panghy: Fixed in 3ef9d0e: reworded the parenthetical so `settings.update` on `specialists.dir` always yields `-32602` (read-only), while `settings.reset` yields `-32602` only when the value is startup-pinned and otherwise clears a file-written value (removing the key from config.toml) — matching `settings.rs` where only `update()` checks `def.read_only` and `reset()` rejects solely via the pin guard.

Issue #1445, opened 2026-08-04
fe: AgentCard still shows previous-turn completion report during a new turn's tool-only (no-text) window


panghy:
## Summary

Follow-up to #1327 (fixed by cloudlands-fe#636, shipped v2.13.0, plus cloudlands-fe#682, v2.15.0). Dogfooding on 2.15.0 still reproduces the sticky completion message: a delegated agent completes with `reportToParent`, the parent sends a follow-up, and while the child works on the new turn the conversation-footer AgentCard keeps showing the previous turn's completion report.

## Residual gap

The #636 fix clears the push-applied `session.digest` at the first activity ping of a new turn and suppresses summaries while `liveResponseLine` is non-empty. Three stale sources remain un-gated in `AgentCard.svelte`'s `effectiveCompletionReport`:

1. **Transcript-derived digest** (`agentData?.digest` from `getAgentPeekData`) — extracted from the previous turn's last assistant message; never cleared at a turn boundary.
2. **`metadata.completionReport`** (`agentData?.completionReport` and the `completionReport`/`lastResponseSummary` props) — persists from the last `reportToParent`; not gated per-turn.
3. **The no-text window**: `liveResponseLine` requires `receivedFirstChunk` + `lastAgentResponse` (text). An agent doing tool-only work for minutes emits no text, so the suppression condition (`liveResponseLine` non-empty) never triggers and `effectiveCompletionReport` falls through to the stale chain the whole time.

## Expected

While the agent is responding (`$agentIsResponding$`), previous-turn summaries (transcript digest, completionReport props/metadata, lastResponseSummary) are never the preview. Precedence while responding: live digest (this turn) → live text → newest user message → tool-use preview. Stale summaries remain the fallback only for idle agents.

## Actual

During a tool-only new turn the footer row shows the previous turn's completion report until the turn ends or text streams.

## References

- #1327 (original report; fix shape)
- intent-hq/cloudlands-fe#636, intent-hq/cloudlands-fe#682 (prior fixes)
- intent-hq/cloudlands-fe#594 / #1231 (no-op guard snapshot)

panghy:
Fixed in cloudlands-fe v2.16.0 (beta), bundles intentd v0.4.2.

panghy:
cloudlands-fe v2.16.0 promoted to stable, bundles intentd v0.4.2.

PR #3354, opened 2026-08-24
docs: document A2A send refusal discriminator, delivery outcome, and replacePending

panghy:
Documents the final landed state of three intentd PRs in the §5.5 agent-facing queue visibility & sender hygiene block of `docs/protocol/methods/agents.md`:

- [intent-hq/intentd#1439](https://github.com/intent-hq/intentd/pull/1439) — the `refused: true` discriminator on the single-pending-message guard refusal (present only on guard refusals, distinguishing them from other `ok: false` shapes), plus the canonical `delivery: "delivered" | "queued" | "held"` outcome field on every successful `ws.agent.send` / `ws.agent.sendToTask` result (with the persist-only-success no-claim rule and the interrupt-dedup "delivered" carve-out).
- [intent-hq/intentd#1440](https://github.com/intent-hq/intentd/pull/1440) — refusal-shape clarifications: `pendingMessageId`, the presented queue echo, and the caveat that manual `ws.agent.removeQueuedMessage` + re-send is NOT atomic (the pending entry is gone even if the re-send then fails).
- [intent-hq/intentd#1445](https://github.com/intent-hq/intentd/pull/1445) — the atomic `replacePending: true` option (options-object third argument): send-first, retract-after (lossless — a failed send never discards the pending entry), reporting `replaced` / `replacedMessageId` on success and `replaceOutcome: "drained" | "none" | "reassigned" | "error"` otherwise; the refusal `instruction` now recommends it as the atomic remediation.

All three are **MCP-only surface changes** (§6.8 principle) — the wire `agent.sendMessage` / `agent.sendToTask` RPCs and their result shapes are unchanged, which the doc states explicitly.

Docs-only PR: no submodule pin change is included (the auto-bump workflow owns pin advancement). Shapes verified against the landed code in `packages/intentd/crates/intent-acp/src/mcp_server/bindings/agent.rs` at intentd main (`e1532b4a`).

### DIFF
--- a/docs/protocol/methods/agents.md
+++ b/docs/protocol/methods/agents.md
@@ -343,11 +343,12 @@ sends. **MCP-only surface changes** (§6.8 principle) — no new wire methods; t
   fromAgentName? }` — sender attribution is **lifted to top level** from the entry's
   `agent_message` auto-tag (`messageMetadata.fromAgentId`/`fromAgentName`, §5.5
   `agent.sendMessage`) and is absent for user-sent entries. Note the MCP presentation differs
-  cosmetically from the services-layer snapshot the guard refusal and `agent.diagnostics`
-  embed: both truncate `content` to 200 chars (with a `…` ellipsis) and drop the bulky
-  `imageBlocks`/`fileBlocks` payloads, but the MCP view lifts attribution top-level while
-  the internal snapshot leaves it inside `messageMetadata` — same entries, same order,
-  different attribution placement.
+  cosmetically from the services-layer snapshot `agent.diagnostics` embeds (the guard
+  refusal embeds this presented view, not the internal snapshot): both truncate `content`
+  to 200 chars (with a `…` ellipsis) and drop the bulky `imageBlocks`/`fileBlocks`
+  payloads, but the MCP view lifts attribution top-level while the internal snapshot
+  leaves it inside `messageMetadata` — same entries, same order, different attribution
+  placement.
 - **Queue merged into `ws.agent.status`** — the target's pending queue rides the status
   result inline as `queue` + `queueLength` (same entry shape and drain-order sorting as
   `ws.agent.getQueue`, `content` truncated to 200 chars), so one status call shows both
--- a/docs/protocol/methods/agents.md
+++ b/docs/protocol/methods/agents.md
@@ -368,12 +368,53 @@ sends. **MCP-only surface changes** (§6.8 principle) — no new wire methods; t
   agent-origin send while the caller already has a pending entry on the target's queue is
   **refused** (agent-origin sends only: FE/user sends and internal wakes are unaffected;
   `priority: "interrupt"` is included — no bypass; `editing: true` entries don't count,
-  the drain skips them). The refusal echoes the target's full queue plus remediation:
-  keep the pending message, or retract it via `ws.agent.removeQueuedMessage` and resend —
-  noting a retract-and-resend lands at the END of the queue. Different senders may still
+  the drain skips them). The refusal is a successful tool result with `ok: false` and —
+  since [intentd#1439](https://github.com/intent-hq/intentd/pull/1439) — the unmissable
+  `refused: true` discriminator (present only on guard refusals, so a naive sender can
+  tell "refused, act on the instruction" apart from other `ok: false` shapes), echoing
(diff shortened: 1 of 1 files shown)


panghy:
## Verification review (agent-verified against intentd main `e1532b4a`)

Verified the documented shapes against `crates/intent-acp/src/mcp_server/bindings/agent.rs` at intentd `origin/main` (`e1532b4a`, the same commit the PR body cites) plus the MCP tool schema in `tools.rs` and the file lists of intentd #1439/#1440/#1445.

### ✅ Shapes match the merged code

- **`refused: true` refusal** — matches `pending_send_refusal()` exactly: successful tool result with `ok: false`, `refused: true` (the only producer of the field in the MCP surface), `error` naming the rule, `pendingMessageId`, `queueLength` + presented `queue` (drain order, 200-char truncated entries), and the `instruction` recommending `replacePending` over manual remove + re-send. `sendToTask` refusals additionally echo `taskNoteId` as documented elsewhere in §5.5.
- **`delivery` outcome** — matches `delivery_outcome()`: `"held"` = `heldForQuestions`; `"queued"` = `queued: true` including the `quarantined`/`archivedParked` indefinite parks (flags kept in the result); `"delivered"` claimed only on a `turnId`-bearing result plus the `deduplicated: true` interrupt-dedup replay; persist-only store-fallback success (`queued: false`, no `turnId`) gets no claim; non-success shapes return `None` (unit tests `delivery_outcome_*` cover all four assertions). Inserted on both `send` and `send_to_task` results.
- **`replacePending`** — matches: options-object third argument (`{ priority?, replacePending? }` per the tool schema), send-first/retract-after (a failed send returns `Err` before any retraction, so the pending entry survives), `replaced: true` + `replacedMessageId` on success, and **all four** `replaceOutcome` values exist in code: `"drained"` (`Err(NotFound)` — entry delivered between guard check and retraction), `"none"` (both fall-through arms: no pending entry / no guard target), `"reassigned"` (`send_to_task` only — the op's re-resolved assignee differs from the guard target, nothing retracted), `"error"` (any other removal error, logged via `tracing::warn!`). The new message is sent in every arm, and an agent caller always gets a replace report.

### ✅ MCP-only framing

intentd #1439 and #1445 touch only `crates/intent-acp/src/mcp_server/{bindings/agent.rs,tools.rs}` + tests; #1440 touches only `tools.rs`. The wire `agent.sendMessage` / `agent.sendToTask` handlers are untouched, so the "no wire-shape change" statement is accurate and consistent with the §6.8 principle framing already used in this block.

### ✅ Diff scope

Single file changed: `docs/protocol/methods/agents.md` (+45/−4). No submodule pin or `.gitmodules` changes.

### ⚠️ One internal inconsistency now exposed (one-line fix)

The new text correctly says the refusal echoes the target's **presented** queue — `pending_send_refusal()` has used `fetch_presented_queue()` → `present_queue()` (attribution lifted top-level, `messageMetadata` dropped) since #816, with `truncate_entry_content()` on top. But the pre-existing "cosmetic difference" note in the `ws.agent.getQueue` bullet a few lines above still claims the guard refusal embeds the **services-layer** snapshot that "leaves it \[attribution\] inside `messageMetadata`". That was already inaccurate for the refusal (it is accurate for `agent.diagnostics`), and the new wording now directly contradicts it within the same §5.5 block.

Suggested minimal fix: in that note, drop "the guard refusal" from the services-layer-snapshot claim (or state that the refusal embeds the presented view and only `agent.diagnostics` embeds the internal snapshot).

### Non-blocking observation

No `docs/protocol/versioning.md` entry — consistent with precedent (#816's guard itself has none; MCP-only surface changes have been recorded there only sporadically, e.g. intentd#1292/#1293), so not treated as a gap.

panghy:
Addressed the ⚠️ internal inconsistency in `89c0541`: the stale "cosmetic difference" note in the `ws.agent.getQueue` bullet now reads "the services-layer snapshot `agent.diagnostics` embeds (the guard refusal embeds this presented view, not the internal snapshot)" — dropping the incorrect claim that the guard refusal embeds the internal snapshot with attribution left inside `messageMetadata`. §5.5 is now internally consistent with the new refusal documentation (presented queue, lifted attribution, per intent-hq/intentd#816).

Replying to the verification review in https://github.com/intent-hq/monorepo/pull/3354#issuecomment-5397135023.

Issue #3403, opened 2026-08-25
MCP OAuth tokens are never refreshed — no expiry check or refresh_token grant in intentd or cloudlands-fe


panghy:
## Summary

MCP OAuth access tokens are never refreshed anywhere in the stack. Once a token expires, requests to the MCP server start failing (typically 401) until the user manually re-runs the OAuth flow. There is no expiry check, no `refresh_token` grant, and no OAuth client library in either component.

## Current behavior

**intentd** (owns the persisted bag):
- The daemon persists whatever JSON the FE hands it via `mcp.oauth.set` into the `mcp_oauth_tokens` SQLite table (migration `0027_mcp_oauth_tokens.sql`) and treats it as an opaque secret.
- `McpOauthService::authorization_header` (`crates/intent-services/src/mcp_oauth.rs`) builds `Bearer <access_token>` straight from the stored bag — it never inspects `expires_at` and never runs a refresh grant.
- No `oauth2` / `openidconnect` crate is in the dependency tree. The only OAuth machinery in intentd is the hand-rolled GitHub device flow (`intent-sourcecontrol/device_flow.rs`), which is unrelated and also refresh-free.

**cloudlands-fe**:
- `src/features/mcp/main/mcp-oauth.ts` defines `refresh_token?` and `expires_at?` on its `OAuthTokens` interface, but nothing ever reads them — `getMcpAuthHeaderAsync` returns the cached `access_token` unconditionally.
- The FE cache is in-memory only (the daemon deliberately never echoes bags back per PROTOCOL.md §5.22.1), so an app restart drops it entirely and FE-side consumers require re-auth even before expiry.
- No `@modelcontextprotocol/sdk` (whose `OAuthClientProvider` would provide refresh) and no OAuth client dependency in `package.json`.

**Related observation**: the interactive flow (`initiateMcpOAuth` — discovery, callback server, code exchange) referenced by comments in `mcp-oauth.ts` no longer exists in the FE source; the write path that populates fresh bags appears to have moved or been dropped during the P3 canonical-persistence migration. Worth confirming while fixing this.

## Expected behavior

Expired MCP OAuth access tokens are transparently refreshed using the stored `refresh_token` (when the server issued one), so authenticated MCP servers keep working without manual re-auth.

## Suggested fix-site

The daemon is the natural place, since it owns the persisted bag and builds the outbound Authorization header:

1. At `authorization_header` build time, check `expires_at` with a skew margin.
[...]

Issue #3402, opened 2026-08-25
Surface a stalled-stream indicator with cancel/retry while a turn has no model activity


panghy:
## Problem

While dogfooding on 2026-08-25 (Camiel's "Fable 5 stalls" report, debug export `intent-debug-2026-08-25-2021.zip`), a turn hung with zero model-stream activity for 14+ minutes while the UI showed only "Thinking · Sent prompt…". The user had no way to tell whether the agent was working, stuck, or dead, and no affordance to act on it short of switching models and re-sending.

Evidence from the export (all UTC, app v2.86.4):

- `00:37:27` — turn failed terminally only after Augment's edge cut the request: `HTTP 408 {"apiStatus":"augmentClientTimeout","httpUrl":"https://e2.api.augmentcode.com/chat-stream"}`. The daemon sat silently waiting the whole time.
- `02:14:40` — `turn resolved normally after a sustained silent tail — suspected truncation (monorepo#2669) silent_tail_ms=923245` (~15.4 min). The detector fired **after** the turn ended.
- `03:07:08` — `agent.sendMessage` dispatched (1.4 MB frame); then complete silence in the intentd log until the debug export at `03:21:39`. No error, no deltas, no state change surfaced anywhere.

Root cause of these specific hangs was upstream (Fable 5 serving path via Augment `chat-stream`; GPT 5.6 streamed fine in the same session), but the client experience is the gap: a dead stream is indistinguishable from a long-thinking model.

## Prior art

- #2669 / #2863 (closed): daemon-side silent-tail/suspected-truncation detection — logs a WARN when the turn *ends*, does nothing while the stall is in progress.
- #977, #1599 (closed): session/prompt idle-timeout handling (1800s) — far too slow to help a user watching a spinner, and terminal rather than recoverable.

## Proposed behavior

1. **intentd**: while a turn is in flight, track time since the last model-stream event (the silent-tail machinery from #2669 already measures this). After a threshold (e.g. 60–120s of zero stream activity mid-turn), emit a non-terminal stall notification on the chat/status stream (e.g. `streamStalled: { silentMs }`), and clear it when activity resumes.
2. **cloudlands-fe**: render the stall state in the transcript/status area — e.g. "No model activity for Nm" alongside the existing Thinking indicator — with two affordances:
   - **Cancel** (existing stop path), and
   - **Retry** — cancel the hung `session/prompt` and re-dispatch the same turn input without the user re-typing/re-attaching.
[...]

PR #3428, opened 2026-08-25
docs(protocol): tokenImpact annotation on SettingDefinition (§5.12)

panghy:
## Summary

Documents the additive `tokenImpact` annotation on the §5.12 `SettingDefinition` wire shape:

- `docs/protocol/methods/settings.md`: adds the optional, presence-detected `tokenImpact?: string` field to the `SettingDefinition` interface and describes the per-toggle `agentFeatures.*` annotations (approximate prompt-token cost, chars/4 measurement; stateSnapshot per turn, taskGraph with a variable per-completion-wake term).
- `docs/protocol/versioning.md`: within-7.4 note — additive response field, presence-detected, no method-catalog change, so no version bump.

Companion implementation PRs: intent-hq/intentd#1467 (daemon field + annotations) and intent-hq/cloudlands-fe#1683 (dimmed FE render).


docs/protocol/versioning.md
```diff
@@ -4,7 +4,7 @@
...
...calls stay workspace-silent — superseding its pre-existing "does NOT touch the workspace-level `unread` flag" contract, and `workspace.markSeen` becomes the mark-ALL composite: it advances every top-level session's seen marker through the same op (each emitting its own `agent:updated` marker event) before the legacy guarded clear, so workspace-seen and conversation-seen can no longer disagree. The turn-end unread raise still writes the stored flag (the none→unread transition emit is unchanged). Also within 7.4 (additive response field, presence-detected per the convention below; no method-catalog change, so no version bump): the `SettingDefinition` **`tokenImpact` annotation** (§5.12) — each `agentFeatures.*` catalog entry carries the optional `tokenImpact` string (an approximate, human-readable prompt-token cost of the gated surface, e.g. `"~620 tokens/session"`; `stateSnapshot` is per turn, `taskGraph` adds a variable per-completion-wake term), served on `settings.list` / `settings.get` definitions and **omitted** on unannotated entries (absent, never `null`), so clients render the dimmed cost hint next to the toggle when present and older daemons simply serve the previous shape.
```
panghy: **Blocking — this rewrite drops two existing within-7.4 notes.**

The new version of this line is ~1.9 KB shorter than the base version, and two notes that exist on current `main` are gone from it:

1. the **mid-turn stream-stall advisories** note (`agent:stream:status` gains `stalled` / `resumed`, `silentMs`; intentd#1462, monorepo#3402), and
2. the **daemon-side MCP OAuth token refresh** note (§5.22.1; monorepo#3403).

It looks like the line was rewritten from a stale base (pre-#3402/#3403) with the `tokenImpact` note appended, which truncates everything `main` added after that point. Verified with:

- base (`main`) contains both phrases: `grep -c 'mid-turn stream-stall advisories' docs/protocol/versioning.md` → 1, `grep -c 'daemon-side MCP OAuth token refresh'` → 1
- PR head contains neither.

Please rebase on current `main` and re-apply the `tokenImpact` note as a pure append so the two existing notes are preserved. (The `settings.md` change in this PR is a clean append and is fine; the new `tokenImpact` note text itself also reads correctly.)

panghy: Confirmed — the line was rewritten from a stale base and dropped the #3402 (stream-stall advisories) and #3403 (MCP OAuth refresh) notes. Fixed in d10d16e: restored `versioning.md` line 7 from current `main` and re-applied the `tokenImpact` note as a pure append. Both phrases now grep at 1 on the PR head, and the diff is a single-line append-only change.

PR #1489, opened 2026-08-05
feat: agent.markSeen seen marker — PROTOCOL v4.5 + intentd/cloudlands-fe submodule bumps

panghy:
Phase-2 monorepo update for the per-conversation **seen marker** feature. Both component PRs are already merged, with the intentd-first merge ordering followed (daemon landed and was confirmed merged before the FE PR merged):

- [intent-hq/intentd#907](https://github.com/intent-hq/intentd/pull/907) — `agent.markSeen` router method (protocol 4.5), squash commit `9edeea0`
- [intent-hq/cloudlands-fe#748](https://github.com/intent-hq/cloudlands-fe/pull/748) — markSeen trigger + "New messages" divider, squash commit `037442ab`

## What changed

**`docs/PROTOCOL.md`** — Version 4.4 → 4.5:

- Title, `**Protocol Version:**` header, `**Version:**` line, the two advertised-version bullets, and the §5.17 `client.hello` examples/prose now read `4.5`; a Version 4.5 entry is appended to the version-history preamble.
- §5 method counts: 315 → 316 dispatchable names, 278 → 279 router methods, agent namespace 40 → 41 (adds `markSeen`).
- §5.5: new `agent.markSeen` catalog row (monotonic persist of `lastSeenMessageId` in session metadata, `agent:updated` emit, dangling-id tolerance, `AgentLite`/`agent.getSession` serving) and `lastSeenMessageId?` added to the `agent.list` `AgentLite` metadata block.
- §6.5: the `agent:updated` event row documents the `agent.markSeen` marker persist (`data { agentId, lastSeenMessageId }`; not emitted on the monotonic no-op or idempotent re-mark).

**Submodule refs**:

- `packages/intentd`: `5b9e4b4` → `218d30f` (latest main; includes intentd#907 `9edeea0` plus intentd#908/#911, both already documented on main via #1487)
- `packages/cloudlands-fe`: `12dd2584` → `037442ab` (latest main = cloudlands-fe#748 squash commit)

## Overlap with #1480 (resolved)

This branch originally carried a 4.4 backfill of the version-history entry for the voice no-API-key error. [#1480](https://github.com/intent-hq/monorepo/pull/1480) merged first with a fuller 4.4 entry, so this branch was rebased onto main: the merged #1480 wording is kept verbatim (the 4.4 entry appears exactly once) and the 4.5 entry is appended after it.

## Verification

- `grep -c 'Version 4.4 is an' docs/PROTOCOL.md` → 1 (no duplicate history entry)
[...]

PR #3520, opened 2026-08-26
docs(protocol): document turn-boundary lastActivity emission cadence

panghy:
## Summary

Documents the new emission cadence of the debounced `workspace:updated { lastActivity }` event in the canonical protocol docs, matching the implementation in [intent-hq/intentd#1489](https://github.com/intent-hq/intentd/pull/1489): the event is now scheduled **only at turn boundaries** instead of on every message/status/agent-op mutation.

## Changes

- **`docs/protocol/methods/workspace.md`** — new "Push cadence — turn boundaries only" block under the `lastActivity` field description: the 3-second debounce window, the three scheduling boundary classes (agent turn end, user-origin message sends, attention raises), the removed mid-turn scheduling sites (agent-origin appends, wake deliveries, agent-op mutations, token-usage recomputes), the scoped monotonic column persist, and the note that the read-path derivation stays fresh regardless.
- **`docs/protocol/06-events.md`** — the `workspace:updated` row now names the debounced `lastActivity`-only delta and its turn-boundary scheduling, cross-referencing §5.1.

Behavior-only change on the implementation side — the derivation formula, the delta wire shape, and the debounce window are unchanged, so no version bump.


panghy (commented):
Verified the prose against the intentd implementation on the `feat/last-activity-turn-boundaries` branch (intent-hq/intentd#1489, submodule checkout at `packages/intentd`). Grepped all non-test `schedule_last_activity_event` call sites; the documented cadence is accurate:

- **Turn end** — `persist_status` / `persist_status_with_stop_reason` (`agent_manager.rs`) gate on `!is_active` (plus the `turn_boundary` opt-out for the no-turn `agent.retry` Error-clear, which the doc's "transitioning the session OUT of an active state" phrasing correctly excludes). ✅
- **User-origin sends** — direct send gated on `origin.is_user()` (`agent_manager.rs` sendMessage), queue-drain persist gated on `user_origin` (`agent_manager.rs`), user-role `agent.appendMessage` gated on `role == "user"` (`agent_ops.rs`). ✅
- **Attention raises** — `agent_request_attention_op` (`agent_ops.rs`) plus `raise_attention` (`lib.rs`; its sole caller is the turn-end unread raise, so it coincides with a turn boundary). ✅
- **Unchanged mechanics** — `LAST_ACTIVITY_DEBOUNCE_MS = 3000`, timer reset per trigger, emit-only-on-change (`old_activity != new_activity`), scoped monotonic `bump_workspace_last_activity` column write, and the `list_workspaces_lite` / seq-0-snapshot restart-freshness rationale all match the doc text verbatim. ✅
- No forbidden terminology; intentd#1489 linked with the full cross-repo form; both checks (bridge-test, pr-title) green.

Two minor, non-blocking notes:

1. **Send-now path omitted from the user-origin enumeration.** Both files enumerate the user-origin class as "the FE `agent.sendMessage` direct send, a user-origin queue-drain delivery, and a user-role `agent.appendMessage`" — but the implementation also schedules on the user-origin-gated `agent.sendQueuedMessageNow` store-only path (`agent_send_queued_message_now_op`, gated on `entry.user_origin`; the manager-backed send-now routes through the direct-send gate). If the list is meant to be exhaustive, consider appending "and the user-origin `agent.sendQueuedMessageNow` force-send"; otherwise it reads as if that path does not schedule.

2. **Merge ordering.** intentd#1489 is still open (auto-merge armed). Per the protocol-docs daemon-first rule, this docs PR should land only after #1489 is confirmed merged, so the canonical contract never describes behavior that isn't on intentd main.

panghy:
Thanks for the thorough verification — both notes addressed:

1. **Send-now path appended** (c66b7db): the user-origin enumeration in both files now includes the user-origin `agent.sendQueuedMessageNow` store-only force-send (`agent_send_queued_message_now_op`, gated on `entry.user_origin`), with `docs/protocol/methods/workspace.md` also noting that the manager-backed send-now routes through the direct-send gate. The list is now exhaustive over all four user-origin scheduling sites.

2. **Merge ordering** — agreed and already the plan: this PR will not be merged (and no auto-merge armed) until [intent-hq/intentd#1489](https://github.com/intent-hq/intentd/pull/1489) is confirmed merged, per the daemon-first rule for protocol docs.

PR #3622, opened 2026-08-27
docs(protocol): document agent retire cascade, cleanup, and watch settlement

panghy:
Follow-up protocol documentation for intent-hq/intentd#1517 (agent retire cascade + cleanup), which merged to intentd main as `cbc4ff8c`.

## What is documented

**`docs/protocol/methods/agents.md`** — new "Retire cascade & cleanup" section under §5.5:

- **Active-descendant guard**: retire fails with `InvalidParams` (`-32602`, nothing mutated — no partial cascade) while any descendant — transitive over `parent_agent_id`, grandchildren included — is still running a turn; the error names each active child. Retired descendants never count; idle/waiting children pass.
- **Child cascade**: every non-terminal, not-already-retired descendant is cascade-retired through the same shared per-session cleanup, each emitting `agent:retired` with `reason: "parent <name> retired"`. A child that started a turn after the guard passed is skipped, not stopped. Restoring the parent does not restore cascaded children.
- **Hook / PR-monitor sweep**: the retiring session's active background hooks and PR monitors are cancelled through the shared cancel transitions (`hook:cancelled` / `prMonitor:cancelled` emitted) with **no wake notice parked** — the owner retired itself and is inert (mirrors the `workspace.archive` sweeps). Event subscriptions are dropped; the pending message queue is kept for a possible restore.
- **Incoming-watch settlement**: `agent:retired` is a terminal completion signal — watchers holding a completion watch get one wake with the terminal retired notice (no re-arm pointer, since `agent.watch` rejects retired targets), and a retired child records in its parent's `after_all` group as terminal non-completing, so groups never hang on it.
- **Restore resurrects nothing**: `agent.restore` clears the mark only — cancelled hooks/PR monitors stay `cancelled` and consumed watches stay consumed.

**`docs/protocol/06-events.md`** — the `agent:retired` event row now covers the per-cascaded-child emits, the no-notice hook/PR-monitor sweep events, and the terminal-completion-signal semantics for watches and `after_all` groups.

## References

- intent-hq/intentd#1517 — the implementing PR (merged)


panghy (commented):
Verified the documented contract against the intentd implementation (squash `cbc4ff8c`, intent-hq/intentd#1517; the retire-relevant sources are byte-identical between that squash and the feature-branch checkout used for review). No inaccuracies found.

**Accuracy — every claim traced to code:**

- **Active-descendant guard**: `agent_retire_op` (`intent-services/src/agent_ops.rs`) rejects with `Error::InvalidParams` — mapped to `-32602` by the router (`intent-transport/src/router.rs`, `INVALID_PARAMS = -32602`) — with the exact message shape `cannot retire: {n} active child agent(s) still running a turn: {name} ({id}), …`, nothing mutated; `collect_retire_descendants` is transitive over `parent_agent_id` (grandchildren included, traversal continues through deleted intermediaries), and the guard filters on `retired_at.is_none() && is_running_turn(status)`, so retired descendants never count and idle/waiting children pass.
- **Cascade**: cascade reason is `format!("parent {} retired", session.name)`; the loop skips already-retired, terminal, and running-turn children (the post-guard race → skipped, not stopped), and a per-child failure logs and continues. Each child goes through the same `retire_one_session` cleanup with its own `agent:retired` emit.
- **Hook / PR-monitor sweep**: `cancel_agent_hooks` / `cancel_agent_pr_monitors` cancel via the shared transitions with notice `None` (no wake parked; the archive sweeps pass a notice — the asymmetry the doc calls out is real), emitting `hook:cancelled` / `prMonitor:cancelled`. `remove_event_subscriptions_for_agent` drops subscriptions; the queue entry is explicitly kept ("restore may drain it later"), and `agent_restore_op` kicks `try_drain_queue`, matching "restore may drain it later".
- **Watch settlement**: the AS-3 delivery loop subscribes to `AGENT_RETIRED` alongside idle/failed/deleted (`spawn_completion_delivery_loop`); the wake tail text matches verbatim ("The agent retired and cannot be re-watched or woken again", no re-arm pointer — `harness/v1.rs` `completion_wake`); `agent.watch` → `agent_watch_op` → `require_agent_session` rejects retired targets; grouped watches record `deleted = event_type == AGENT_DELETED || AGENT_RETIRED` — the terminal, non-completing bucket, exactly as documented.
- **Restore resurrects nothing**: `agent_restore_op` clears the mark and emits `agent:restored` only — no hook/monitor/watch resurrection anywhere on the path.

**Style/terminology**: §5.40 (hooks), §5.42 (PR monitors), and the §5.5 cross-references match established usage; `AS-3` is a first appearance in the docs but consistent with the existing internal-codename convention (STAB-28, SUB-1, P3-1.2b).

**Table integrity**: the edited `agent (lifecycle)` row in `06-events.md` remains a single table row with the correct cell count (the only added pipes are the row delimiters).

No changes requested.

Issue #3682, opened 2026-08-27
fe: provider model pickers query the app-primary backend, not the window's backend — wrong daemon's model catalog in multi-backend setups


panghy:
All nine `<provider>:get-models` IPC handlers discard the invoke `_event` and `daemon-model-catalog.ts` calls the **unrouted** `getBackendClient()`. With per-window backend connections (#1930), the model picker in a window bound to backend A therefore fetches `models.list` from whichever backend is app-primary/active (B).

## Repro (observed)

Three connected backends: `local` (intentd 0.8.3), `intent1:5181` (0.8.4, app-active after `reconcileActiveConnectionOnBoot`), `mac-mini:5181` (0.8.5).

Working in a workspace that lives on **local**, provider = claude-code:

- Model picker shows 3 models — Opus (1M context), Sonnet, Haiku — with `… · $3/$15 per Mtok`-style descriptions.
- The **local** daemon actually serves 4:

```
$ intentd call models.list --params {"providerId":"claude-code"}
opus[1m]            | Opus (1M context)
claude-fable-5[1m]  | Fable
sonnet              | Sonnet
haiku               | Haiku
```

Neither cloudlands-fe nor intentd generates the `per Mtok` pricing text (`grep -r Mtok` is empty in both) — it comes verbatim from a different host adapter/CLI, confirming the rows are intent1's, not local's.

- The picker's per-group ↻ refresh does not help: `forceRefresh: true` also goes to the wrong daemon.
- Same window logged `BackendError: Workspace not found` / `not found: workspace fable-model` from the active connection while displaying that local workspace — the same misrouting surfacing elsewhere.

## Root cause

`src/main/utils/daemon-model-catalog.ts`:

```ts
const result = await getBackendClient().request<WireModelsListResult>(models.list, params);
```

`getProviderModelsEnvelope(providerId, params)` takes no event/backend id, and every handler drops it:

```ts
ipcMain.handle(CLAUDE_CODE_CHANNELS.GET_MODELS, async (_event, params) =>
  getProviderModelsEnvelope(claude-code, params));
```

Affected: auggie, claude-code, codex, cortex, droid, grok, opencode, pi, unsloth — plus the main-side model-override validator in `model-pool.ts`, which shares the helper.

The correct routed seam already exists and is used elsewhere: `getBackendClientForIpcEvent(event)` (backend.ipc.ts) and `workspace-backend-client.ts` (#3501).

## Impact

[...]

PR #3684, opened 2026-08-27
docs: document the 24h model-catalog staleness threshold

panghy:
Documents the model-catalog cache staleness threshold introduced by [intent-hq/intentd#1538](https://github.com/intent-hq/intentd/pull/1538) (`MODELS_STALE_AFTER`, 24 hours), superseding the v6.0 "cached entries are served indefinitely" policy (intentd#987).

Refs intent-hq/intent#3682 (docs follow-up; the issue itself is fixed by the intentd PR).

## Changes

- **`docs/protocol/methods/models-providers.md`** (§5.30 semantics):
  - "One generic per-provider cache" bullet: replaces "no TTL / served indefinitely" with the 24-hour freshness contract (`MODELS_STALE_AFTER`), including the saturating-subtraction age computation (a future-stamped entry reads as age zero — fresh).
  - `forceRefresh` bullet: softened from "the only client-driven way to re-probe" to "the only client-driven way to re-probe a *still-fresh* cached provider on demand" (aged entries re-probe on their own).
  - "Non-forced reads" bullet: a probe now runs on a true miss **or an aged entry** (at or past the 24-hour threshold); a failed re-probe of an aged entry keeps serving the last-good list labeled `stale: true` + `warning`.
  - "Probe guards" bullet: clarifies the negative-window behavior for aged entries (stale-labeled last-good, no re-probe) vs. true misses, and that a success clears the negative entry / empty successes are never cached.
  - "Legacy path" bullet: "served indefinitely" → "served while fresh and re-probed once aged (the same 24-hour staleness threshold)".
- **`docs/protocol/versioning.md`**:
  - v6.0 intentd#987 entry: adds a supersession pointer to the v8.4 threshold.
  - New "Also within 8.4" entry (behavior only, no method-catalog or wire-shape change, so no version bump) documenting the threshold, referencing intentd#1538 and intent#3682.

## Consistency sweep

`grep -rniE "indefinitely|no TTL|never expires|regardless of age" docs/` — all remaining matches are unrelated surfaces (PR monitors' deliberate no-TTL contract, background-hook TTL docs, the `ProcessRegistry::evict_idle` LRU note in ARCHITECTURE.md §process pool) and are correct as written. `docs/ARCHITECTURE.md`'s "cached per provider" mention (§default-model resolution) stays accurate without change.

**Merge gate:** do not merge until intentd#1538 (currently in the intentd merge queue) is confirmed landed on `main`.


panghy (commented):
**Verification review** (verifier agent) — checked against the authoritative source `crates/intent-services/src/model_catalog.rs` at intentd#1538 head `e56b9b13`.

## 1. Technical accuracy vs. source — all claims verified ✅

| Documented claim | Source evidence (e56b9b13) |
|---|---|
| 24-hour threshold `MODELS_STALE_AFTER` | `const MODELS_STALE_AFTER = Duration::from_secs(24 * 60 * 60)` |
| Fresh (< 24h) served, aged (at or past threshold) re-probed | `fresh_hit()` gates on `age < MODELS_STALE_AFTER` (strict `<`, so "at or past" = aged is exact); `resolve_with_cache` probes on miss/aged/forced |
| Failed re-probe of aged entry serves last-good labeled `stale: true` + `warning`; never degrades below last-good | `failure_fallback()` → `last_good()` (age-agnostic, version-key-checked) with `stale: true` + "; serving last known model list" warning |
| Negative window (60s): true miss → degradation (static); aged entry → stale-labeled last-good | `negative_reason()` short-circuits to `failure_fallback()`: with a last-good it serves it stale-labeled; without one it returns `models: None` (caller degrades). Fresh entries never reach the negative check (`fresh_hit` is consulted first) |
| Saturating-subtraction age; future-stamped entry reads as age zero (fresh) | `now_ms.saturating_sub(entry.fetched_at_ms)` in `fresh_hit`, with the matching doc comment |
| Single-flight per (provider, version key); `forceRefresh` skips both cache reads but is still single-flighted | `join_inflight()` runs unconditionally; the `!force_refresh` guard only skips `fresh_hit` + `negative_reason` |
| Empty successes served but never cached; any success clears the negative entry | initializer: `clear_negative()` on any success, `store()` only when `!models.is_empty()` |

Dropping the old "the only automatic invalidation" qualifier on the pin-bump sentence is correct — aging is now a second automatic re-probe trigger.

## 2. Consistency sweep ✅

`git grep -niE "indefinitely|no TTL|never expires|regardless of age" docs/` at the PR head: every remaining match is an unrelated surface (PR monitors' deliberate no-TTL contract in agent-aux.md / pr.md, ProcessRegistry LRU note in ARCHITECTURE.md:734, FE troubleshooting guide, workspace.md `waiting`-flag note). No lingering model-catalog "served indefinitely" / "no TTL" claims survive outside the intentionally historical v6.0 entry, which now carries the supersession pointer.

## 3. versioning.md conventions ✅

- New entry uses the established "Also within 8.4 (behavior only, no method-catalog or wire-shape change, so no version bump — …)" framing, referencing intentd#1538 and intent#3682.
- v6.0 intentd#987 entry keeps its historical wording and gains the parenthetical supersession pointer ("later superseded within v8.4 … see below"), matching prior supersession precedents.

## 4. Footer-token check ✅

No literal breaking-change footer token (or Release-As form) anywhere in the diff, branch commit messages, PR title, or body. Single branch commit is a valid conventional commit.

**Verdict: APPROVED.** Per the PR's own merge gate: do not merge until intentd#1538 is confirmed landed on intentd `main`.

PR #3720, opened 2026-08-28
chore: reconcile issue-intake labels, severity taxonomy, and contact links

panghy:
Intake-surface groundwork for the automatic issue triage pipeline (part of the triage effort; no workflow code in this PR).

## Changes

**Repo labels** (already created via `gh label create`, listed here for the record):
- `needs-triage` — awaiting triage
- `needs-info` — waiting on the reporter; removed on their next comment
- `possible-duplicate` — suspected duplicate, needs confirmation before closing
- `priority:P0` — critical: crash, data loss, or corruption (color `#B60205`, consistent with the existing `priority:*` family)

**`bug_report.yml`**
- Severity dropdown expanded to P0–P3 so options map 1:1 onto `priority:P0|P1|P2|P3` labels (previously P0/P1/P2 vs labels P1/P2/P3, with "papercut" ambiguously on P2). The mapping is documented in a YAML comment; each option is keyed on its leading `Pn` token for mechanical extraction.
- Component and agent-filed disclaimers now say labels are applied automatically from the checked boxes (they will be, by the triage workflow) instead of promising manual labeling.

**`feature_request.yml`**
- Same disclaimer update for the component checkboxes.

**`config.yml`**
- `contact_links` added: private security advisories (private vulnerability reporting is enabled) and repo discussions for questions. `blank_issues_enabled: true` kept.

## Verification
- `gh label list --repo intent-hq/intent` shows all four new labels with descriptions.
- All three template YAML files parse cleanly (`yaml.safe_load`).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -30,23 +30,32 @@ body:
     attributes:
       label: Component
       description: >
-        Check all that apply. Checkboxes do not apply labels automatically —
-        a triager will add the matching label (`component:intentd`,
-        `component:fe`, `component:ios`, `docs`).
+        Check all that apply. Labels are applied automatically from the
+        checked boxes (`component:intentd`, `component:fe`, `component:ios`,
+        `docs`).
       options:
         - label: intentd (Rust backend daemon)
         - label: cloudlands-fe (Electron + SvelteKit desktop frontend)
         - label: ios (SwiftUI companion app)
         - label: docs / tooling
+  # Severity → label mapping (each option maps 1:1 onto a priority:* label,
+  # keyed on the leading "Pn" token):
+  #   P0 — … → priority:P0
+  #   P1 — … → priority:P1
+  #   P2 — … → priority:P2
+  #   P3 — … → priority:P3
   - type: dropdown
     id: severity
     attributes:
       label: Severity
-      description: Per the taxonomy in docs/README.md
+      description: >
+        The matching `priority:*` label is applied automatically from your
+        selection.
       options:
-        - P0 — crash, data-loss, or corruption; blocks shipping to external users
+        - P0 — crash, data loss, or corruption; blocks shipping to external users
         - P1 — broken feature; app still usable but with significant workaround required
-        - P2 — papercut; annoying but does not block workflows
+        - P2 — degraded behavior; should be fixed, but impact is limited
+        - P3 — papercut; annoying but does not block workflows
     validations:
       required: true
   - type: checkboxes
@@ -54,4 +63,4 @@ body:
     attributes:
       label: Agent-filed
       options:
-        - label: This issue was filed by an AI agent (a triager will add the `agent-filed` label)
+        - label: This issue was filed by an AI agent (the `agent-filed` label is applied automatically)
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -30,28 +30,37 @@ body:
...
+      description: >
+        The matching `priority:*` label is applied automatically from your
+        selection.
       options:
-        - P0 — crash, data-loss, or corruption; blocks shipping to external users
+        - P0 — crash, data loss, or corruption; blocks shipping to external users
         - P1 — broken feature; app still usable but with significant workaround required
-        - P2 — papercut; annoying but does not block workflows
+        - P2 — degraded behavior; should be fixed, but impact is limited
+        - P3 — papercut; annoying but does not block workflows
```
panghy, line 58: The old description pointed at the taxonomy in `docs/README.md`, which still documents the previous three-tier scheme (lines 95–98: P0/P1, and **P2 — papercut**). This PR removes the pointer but leaves that doc stale, so the repo now carries two conflicting definitions of P2. Since `docs/README.md` lives in this same repo, please update its severity taxonomy to the new P0–P3 wording in this PR (P2 = degraded behavior, P3 = papercut) so the canonical doc and the intake form agree.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1 +1,8 @@
 blank_issues_enabled: true
+contact_links:
+  - name: Report a security vulnerability
+    url: https://github.com/intent-hq/intent/security/advisories
```
panghy: `/security/advisories` is the list of published advisories. Since private vulnerability reporting is enabled on this repo, consider linking to `https://github.com/intent-hq/intent/security/advisories/new`, which takes the reporter straight to the private report form (this is the URL GitHub's docs recommend for contact links). The list page does have a "Report a vulnerability" button, so this is a nit, but the direct form removes one step and avoids showing published advisories to someone trying to file privately.

panghy (commented):
Reviewed for YAML validity, issue-forms schema correctness, label mapping, and wording. All three files parse cleanly, the forms schema is valid, the P0–P3 options map unambiguously onto the existing `priority:P0..P3` labels (all four exist, as do `needs-triage`/`needs-info`/`possible-duplicate`/`agent-filed`), and both contact-link targets are live (discussions and private vulnerability reporting are enabled on the repo). Two inline comments below.

One non-blocking sequencing note: the templates now state labels "are applied automatically", but the triage workflow that does this is deliberately not in this PR. Until it lands, issues filed from these forms will promise automation that doesn't run yet. Fine if the workflow follows shortly; just flagging the gap.

.github/ISSUE_TEMPLATE/bug_report.yml
panghy: Done in 190778c — docs/README.md now documents the four-tier P0–P3 taxonomy (P3 = papercut) and notes the 1:1 mapping onto the priority:* labels.

.github/ISSUE_TEMPLATE/config.yml
panghy: Done in 190778c — link now points at /security/advisories/new (direct private-report form).

panghy:
Re the sequencing note: acknowledged — the deterministic triage workflow that applies these labels is being built in parallel (same effort, separate PR). The disclaimer wording is intentionally forward-looking so the templates need not change again when it lands.

PR #3722, opened 2026-08-28
ci: add deterministic issue-triage workflow (pass 1)

panghy:
Pass 1 of the issue-triage pipeline (spec: "Automatic issue triage for intent-hq/intent"): deterministic normalization of every issue — template fields become labels instantly, and `needs-triage` marks the untriaged queue.

## What's in here

- **`.github/workflows/issue-triage.yml`** — runs on `issues: [opened, edited, reopened]`, plus `workflow_dispatch` (issue number + `dry_run` defaulting to true) as the dry-run mode that prints intended label ops without writing.
  - Least privilege: `contents: read`, `issues: write`, default `GITHUB_TOKEN`.
  - Concurrency group per issue number.
  - `needs-triage` is added on `opened` only — never re-added on `edited`/`reopened` once removed.
  - Add-only and idempotent: the job computes labels from the current body, diffs against current labels, and only adds what's missing. It never removes labels, so human corrections stick and re-runs never flip anything.
- **`.github/scripts/issue-triage/parse-issue.js`** — zero-dependency CommonJS parser for rendered issue-form bodies:
  - Component checkboxes → `component:intentd` / `component:fe` / `component:ios` / `docs` (matched on the option's leading token, tolerant of wording tweaks).
  - Severity dropdown → `priority:P0..P3`, keyed on the leading `Pn` token per the mapping comment in `bug_report.yml` (landed in #3720).
  - Agent-filed checkbox → `agent-filed`.
  - Blank / free-form bodies yield no labels (so blank issues get only `needs-triage`).
- **Fixture-driven tests** — `parse-issue.test.js` (node:test, no deps) over five checked-in sample bodies plus inline edge cases (CRLF, `[X]`, `_No response_`, duplicate sections, unknown checkbox text). Wired into CI as a new `triage-parser-test` job mirroring `bridge-test`.

## Verification

- `node --test .github/scripts/issue-triage/parse-issue.test.js` — 10/10 pass locally.
- `actionlint` clean on `issue-triage.yml` and `ci.yml`.
- Post-merge smoke: `gh workflow run issue-triage.yml -f issue_number=<n> -f dry_run=true` prints intended ops without writing.

## Not in scope (later spec tasks)

Needs-info completeness loop; agentic pass 2 (dedup / inference / priority suggestion / `needs-triage` removal).

### DIFF
--- a/.github/scripts/issue-triage/parse-issue.js
+++ b/.github/scripts/issue-triage/parse-issue.js
@@ -1,0 +1,96 @@
+// Deterministic issue-form parser for the triage workflow (pass 1).
+//
+// Reads the rendered markdown body of an issue filed through the
+// .github/ISSUE_TEMPLATE forms and derives the labels that the template
+// fields map onto:
+//   - "Component" checkboxes -> component:intentd / component:fe /
+//     component:ios / docs
+//   - "Severity" dropdown    -> priority:P0..P3 (keyed on the leading
+//     "Pn" token, per the mapping comment in bug_report.yml)
+//   - "Agent-filed" checkbox -> agent-filed
+//
+// Pure and dependency-free so it can be unit-tested with `node --test`
+// and required from actions/github-script. It only ever *derives* labels
+// to add; the workflow never removes labels based on its output.
+
+'use strict';
+
+// Checkbox option text -> label, matched on the option's leading token so
+// wording tweaks after the token don't break extraction.
+const COMPONENT_PREFIXES = [
+  { prefix: 'intentd', label: 'component:intentd' },
+  { prefix: 'cloudlands-fe', label: 'component:fe' },
+  { prefix: 'ios', label: 'component:ios' },
+  { prefix: 'docs', label: 'docs' },
+];
+
+// Split a rendered issue-form body into { sectionLabel: contentLines }.
+// Issue forms render each field as a `### <label>` heading.
+function splitSections(body) {
+  const sections = Object.create(null);
+  let current = null;
+  for (const line of String(body).split(/\r?\n/)) {
+    const heading = line.match(/^###\s+(.+?)\s*$/);
+    if (heading) {
+      current = heading[1];
+      if (!sections[current]) sections[current] = [];
+      continue;
+    }
+    if (current) sections[current].push(line);
+  }
+  return sections;
+}
+
+// Checked checkbox items ("- [x] <text>") within a section.
+function checkedItems(lines) {
+  const items = [];
+  for (const line of lines) {
+    const m = line.match(/^\s*-\s*\[[xX]\]\s*(.+?)\s*$/);
+    if (m) items.push(m[1]);
+  }
+  return items;
+}
+
+// First non-empty line of a section (the dropdown selection, or
+// "_No response_" when unanswered).
+function firstContentLine(lines) {
+  for (const line of lines) {
+    const trimmed = line.trim();
+    if (trimmed) return trimmed;
+  }
+  return '';
+}
+
+// Derive the labels a template-filed issue body maps onto. Returns a
+// deduplicated array; an empty/blank/free-form body yields [].
+function labelsForIssueBody(body) {
+  const labels = [];
+  if (!body || typeof body !== 'string') return labels;
+  const sections = splitSections(body);
+
+  if (sections['Component']) {
+    for (const item of checkedItems(sections['Component'])) {
+      const lower = item.toLowerCase();
+      for (const { prefix, label } of COMPONENT_PREFIXES) {
+        if (lower.startsWith(prefix)) {
+          labels.push(label);
+          break;
+        }
+      }
+    }
+  }
+
+  if (sections['Severity']) {
+    const selection = firstContentLine(sections['Severity']);
+    const m = selection.match(/^P([0-3])\b/);
+    if (m) labels.push(`priority:P${m[1]}`);
+  }
+
+  if (sections['Agent-filed'] && checkedItems(sections['Agent-filed']).length > 0) {
+    labels.push('agent-filed');
+  }
+
+  return [...new Set(labels)];
+}
+
+module.exports = { labelsForIssueBody };
--- a/.github/scripts/issue-triage/parse-issue.js
+++ b/.github/scripts/issue-triage/parse-issue.js
@@ -25,11 +25,20 @@ const COMPONENT_PREFIXES = [
 ];
 
 // Split a rendered issue-form body into { sectionLabel: contentLines }.
-// Issue forms render each field as a `### <label>` heading.
+// Issue forms render each field as a `### <label>` heading. Lines inside
+// fenced code blocks (``` / ~~~) are skipped entirely so template markdown
+// pasted into a fence (logs, "here's my template output") cannot derive
+// spurious headings or checkboxes.
 function splitSections(body) {
   const sections = Object.create(null);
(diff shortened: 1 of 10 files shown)


.github/scripts/issue-triage/parse-issue.js
```diff
@@ -0,0 +1,96 @@
...
+    const heading = line.match(/^###\s+(.+?)\s*$/);
+    if (heading) {
+      current = heading[1];
+      if (!sections[current]) sections[current] = [];
+      continue;
+    }
+    if (current) sections[current].push(line);
+  }
+  return sections;
+}
```
panghy, line 51: **Code-fence blindness** (non-blocking): `splitSections` doesn't track fenced code blocks, so a body that pastes template markdown inside a ``` fence (e.g. "here's my template output:" followed by a fenced copy) derives spurious labels. Verified:

```
fenced-block body -> ["component:intentd","priority:P0"]
```

Impact is limited (add-only, wrong labels are human-removable), but combined with the re-add-on-edited behavior a fence-injected label is sticky. Minimal fix: toggle an `inFence` flag on lines matching `/^\s*(```|~~~)/` and skip heading/checkbox matching while inside. Worth a fixture + test when addressed.

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,93 @@
+name: Issue triage
+
+# Pass 1 of the triage pipeline: deterministic normalization. Template
+# fields become labels (component checkboxes, severity dropdown, agent-filed
+# checkbox) and `needs-triage` marks the untriaged queue on open. The parser
+# only ever ADDS labels — it never removes any, so human corrections stick
+# and re-runs are idempotent. `needs-triage` is added only on `opened`,
+# never re-added on `edited`/`reopened` once removed.
```
panghy: **"Human corrections stick" is slightly overstated** (non-blocking, matches the acceptance criteria's add-only definition): corrections stick only when the corrected label is no longer derivable from the current body. If a human removes a label the body still derives (wrong checkbox, or the fence-injection case), the next `edited` event re-adds it. Two related consequences worth a sentence in this comment block:

1. Priority accumulation: user files with P2, later edits the dropdown to P0 -> both `priority:P2` and `priority:P0` end up applied (add-only never retires the stale one). Pass 2 / a human has to resolve the contradiction.
2. The only way to make a correction stick against a template-derived label is to edit the body itself.

Both are acceptable pass-1 tradeoffs per the task note; suggest documenting them here so the next person doesn't read "corrections stick" as unconditional.

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,93 @@
...
+            // payload can be stale on rapid edits, and workflow_dispatch has
+            // no issue payload at all).
+            const { data: issue } = await github.rest.issues.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issueNumber,
+            });
+
+            const derived = labelsForIssueBody(issue.body || '');
+            if (action === 'opened') derived.push('needs-triage');
```
panghy: **Manual dispatch can never apply `needs-triage`** (non-blocking): `action` is `'manual'` on `workflow_dispatch`, so a dispatch with `dry_run=false` — the natural recovery path when the `opened` run was missed or failed — cannot add `needs-triage`. If that recovery scenario matters, consider an explicit `apply_needs_triage` boolean input (default false) so the operator opts in; otherwise a one-line comment noting the limitation would do.

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,93 @@
...
+          ISSUE_NUMBER: ${{ github.event.issue.number || inputs.issue_number }}
+          DRY_RUN: ${{ inputs.dry_run == true }}
+        with:
+          script: |
+            const path = require('path');
+            const { labelsForIssueBody } = require(
+              path.join(process.env.GITHUB_WORKSPACE, '.github/scripts/issue-triage/parse-issue.js')
+            );
+
+            const issueNumber = Number(process.env.ISSUE_NUMBER);
```
panghy, line 69: Nit: `Number(process.env.ISSUE_NUMBER)` on a garbage dispatch input yields `NaN` and fails later inside `issues.get` with a confusing API error. A cheap early guard (`if (!Number.isInteger(issueNumber)) core.setFailed(...)`) makes operator mistakes self-explanatory.

panghy (commented):
Verifier review (pass 1 triage workflow). Independently verified: `node --test` 10/10 pass locally, `actionlint` v1.7.12 clean on both workflows, parser output matches the exact template option strings from #3720 (including the em-dash severity options, `docs / tooling`, and `_No response_` handling), least-privilege permissions look right (`contents: read` / `issues: write`, no body interpolation into shell or expression contexts, numeric concurrency key). CI green including the new `triage-parser-test` job.

Four non-blocking findings below as inline comments: code-fence blindness in the parser, two documented-tradeoff clarifications on the add-only semantics, and a small dispatch-input validation nit. None block approval against the task's acceptance criteria. Not merging — merge requires explicit human permission.

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,93 @@
...
+              issue_number: issueNumber,
+            });
+
+            const derived = labelsForIssueBody(issue.body || '');
+            if (action === 'opened') derived.push('needs-triage');
+
+            const current = new Set(
+              issue.labels.map((l) => (typeof l === 'string' ? l : l.name))
+            );
+            const toAdd = [...new Set(derived)].filter((l) => !current.has(l));
```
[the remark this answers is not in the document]
panghy: Addressed in 109ac84 by documenting the behavior precisely in the workflow header: add-only means a label the CURRENT body still derives is re-added on the next edited event — the way to make a correction stick is to fix the body's checkbox/dropdown itself (the authoritative source pass 1 normalizes from). Removing labels based on prior-run state is out of scope for pass 1 per the task note ("only ADD labels derived from the body; never remove"); pass 2 / humans own re-classification.

.github/scripts/issue-triage/parse-issue.js
panghy: Fixed in 109ac84: `splitSections` now toggles an `inFence` flag on lines matching `/^\s*(```|~~~)/` and skips everything inside, so fenced template copies derive nothing. Added the `fenced-template.md` fixture plus three tests (fixture, fence-only body, tilde fences with a real section after the fence) — 13/13 pass.

.github/workflows/issue-triage.yml
panghy: Added in 109ac84: new apply_needs_triage boolean dispatch input (default false); a dispatch with dry_run=false and apply_needs_triage=true is now the recovery path for a missed opened run.

.github/workflows/issue-triage.yml
panghy: Added in 109ac84: early Number.isInteger guard with core.setFailed on a garbage issue_number input, so operator mistakes fail fast with a clear message.

.github/workflows/issue-triage.yml
panghy: Documented in 109ac84: the workflow header now spells out both tradeoffs — re-add-on-edited while the body still derives the label (a correction must edit the body itself), and priority-label accumulation across dropdown edits (stale label retired by pass 2 or a human) — instead of the unconditional "corrections stick" phrasing.

PR #3726, opened 2026-08-28
ci: add needs-info completeness loop to issue triage

panghy:
Extends the pass-1 issue-triage workflow (#3722) with the needs-info completeness loop from the triage-pipeline spec.

## What

- **Completeness check** (new step in the `triage` job): after label derivation, the issue body is assessed —
  - bug-template bodies must have real content in **Reproduction steps** and **Expected vs actual** (empty, `_No response_`, the untouched `1. ...` placeholder, or filler like `n/a`/`tbd` are flagged);
  - blank issues (no template sections) must clear a minimal meaningful-content bar (markdown image/link URLs don't count);
  - feature requests and other sectioned bodies are never flagged.

  Incomplete issues get **one** polite templated comment + the `needs-info` label. The comment embeds a hidden HTML marker (`<!-- issue-triage: needs-info -->`, same idempotency precedent as `packages/intentd/scripts/notify-fixed-issues.sh`), so edits/re-runs never double-post or re-apply the label once the nudge exists.

- **Author-reply handler** (new `needs-info-reply` job on `issue_comment: created`): when the **issue author** comments on a `needs-info` issue, the label is removed and the pass-1 label derivation re-runs on the current body (add-only, never re-adds `needs-triage`/`needs-info`). Non-author comments — and the marker-carrying nudge itself — change nothing. PR comments are excluded.

- No stale-close of any kind (per spec).

## Files

- `.github/workflows/issue-triage.yml` — completeness step, `issue_comment` trigger + reply job, header docs
- `.github/scripts/issue-triage/needs-info.js` — pure decision logic: `assessCompleteness`, `buildNudgeComment`, `shouldNudge` (marker idempotency), `shouldClearNeedsInfo` (author check)
- `.github/scripts/issue-triage/needs-info.test.js` + 2 new fixtures — 18 tests covering incomplete-body detection, marker idempotency, author-vs-non-author
- `.github/scripts/issue-triage/parse-issue.js` — export `splitSections` (no behavior change)
- `.github/workflows/ci.yml` — `triage-parser-test` also runs the new suite

## Verification

- `node --test .github/scripts/issue-triage/parse-issue.test.js .github/scripts/issue-triage/needs-info.test.js` → 31/31 pass
- `actionlint` on both workflows → clean
- Dry-run path: `gh workflow run issue-triage.yml -f issue_number=<n> -f dry_run=true` now also prints the intended nudge comment/label ops without writing.

### DIFF
--- a/.github/scripts/issue-triage/needs-info.js
+++ b/.github/scripts/issue-triage/needs-info.js
@@ -1,0 +1,128 @@
+// Needs-info completeness loop for the triage workflow (pass 1).
+//
+// Detects incomplete bug reports (empty/placeholder "Reproduction steps" or
+// "Expected vs actual" sections) and near-empty blank-issue bodies, and
+// carries the pure decision logic for the one-nudge loop:
+//   - shouldNudge:          post the templated nudge + `needs-info` at most
+//                           once per issue (hidden HTML marker comment is the
+//                           idempotency guard, same precedent as
+//                           packages/intentd/scripts/notify-fixed-issues.sh)
+//   - shouldClearNeedsInfo: only the issue author's reply clears the label
+//
+// Pure and dependency-free so it can be unit-tested with `node --test`
+// and required from actions/github-script.
+
+'use strict';
+
+const { splitSections } = require('./parse-issue.js');
+
+const NEEDS_INFO_LABEL = 'needs-info';
+
+// Hidden marker embedded in the nudge comment; its presence anywhere in the
+// issue's comments means the nudge was already posted (re-runs never
+// double-post, even after the label was removed).
+const NEEDS_INFO_MARKER = '<!-- issue-triage: needs-info -->';
+
+// Bug-template sections that must carry real content.
+const REQUIRED_BUG_SECTIONS = ['Reproduction steps', 'Expected vs actual'];
+
+// Words that alone convey no information ("tbd", "n/a", ...).
+const PLACEHOLDER_WORDS = new Set(['tbd', 'todo', 'none', 'nil', 'na', 'idk', 'unknown']);
+
+// Minimum meaningful characters for a free-form (non-template) body.
+const NEAR_EMPTY_MIN_CHARS = 30;
+
+// True when a section's content is empty, "_No response_", or contains no
+// meaningful words (only punctuation, list numbering, or placeholder words —
+// e.g. the template's untouched "1. ...\n2. ..." placeholder).
+function isPlaceholderSection(lines) {
+  const text = (lines || []).join('\n').trim();
+  if (!text || /^_no response_$/i.test(text)) return true;
+  const tokens = (text.match(/[A-Za-z]{2,}/g) || [])
+    .map((t) => t.toLowerCase())
+    .filter((t) => !PLACEHOLDER_WORDS.has(t));
+  return tokens.length === 0;
+}
+
+// Meaningful character count of a free-form body: markdown images/links and
+// bare URLs are stripped first so a body that is only a screenshot or a link
+// still counts as near-empty.
+function meaningfulLength(body) {
+  const stripped = String(body || '')
+    .replace(/!?\[[^\]]*\]\([^)]*\)/g, ' ')
+    .replace(/https?:\/\/\S+/g, ' ');
+  const tokens = stripped.match(/[A-Za-z0-9]{2,}/g) || [];
+  return tokens.join('').length;
+}
+
+// Assess an issue body's completeness. Returns { incomplete, reasons }.
+//   - Bodies with the bug template's repro/expected sections must have real
+//     content in both.
+//   - Bodies with no template sections at all (blank issues) must clear a
+//     minimal meaningful-content bar.
+//   - Everything else (e.g. feature requests) is never flagged.
+function assessCompleteness(body) {
+  const reasons = [];
+  const sections = splitSections(body || '');
+  const isBugTemplate = REQUIRED_BUG_SECTIONS.some((name) => sections[name]);
+  if (isBugTemplate) {
+    for (const name of REQUIRED_BUG_SECTIONS) {
+      if (!sections[name] || isPlaceholderSection(sections[name])) {
+        reasons.push(`the "${name}" section is empty or placeholder-only`);
+      }
+    }
+  } else if (Object.keys(sections).length === 0) {
+    if (meaningfulLength(body) < NEAR_EMPTY_MIN_CHARS) {
+      reasons.push(
+        'the issue body has almost no content — please describe what happened, ' +
+          'how to reproduce it, and what you expected'
+      );
+    }
+  }
+  return { incomplete: reasons.length > 0, reasons };
+}
+
+// The one templated nudge comment, marker included.
+function buildNudgeComment(reasons) {
+  const bullets = reasons.map((r) => `- ${r}`).join('\n');
+  return [
+    'Thanks for the report! To triage it we need a bit more detail — right now:',
+    '',
+    bullets,
+    '',
+    'Could you reply to this issue with the missing details? The `needs-info`',
+    'label is removed automatically on your next comment and triage re-runs.',
+    '',
+    NEEDS_INFO_MARKER,
+  ].join('\n');
+}
+
+// Nudge at most once: never when the body is complete, the label is already
+// present, or any existing comment carries the marker (i.e. we nudged before,
+// even if the label was since removed).
+function shouldNudge({ assessment, labels, commentBodies }) {
+  if (!assessment || !assessment.incomplete) return false;
+  if ((labels || []).includes(NEEDS_INFO_LABEL)) return false;
+  const bodies = commentBodies || [];
+  if (bodies.some((b) => typeof b === 'string' && b.includes(NEEDS_INFO_MARKER))) return false;
+  return true;
+}
+
+// Clear `needs-info` only when the ISSUE AUTHOR comments on a labeled issue.
+// Comments from anyone else — including our own marker-carrying nudge — do
+// not clear it.
+function shouldClearNeedsInfo({ labels, issueAuthor, commentAuthor, commentBody }) {
+  if (!(labels || []).includes(NEEDS_INFO_LABEL)) return false;
+  if (!issueAuthor || !commentAuthor || commentAuthor !== issueAuthor) return false;
+  if (typeof commentBody === 'string' && commentBody.includes(NEEDS_INFO_MARKER)) return false;
+  return true;
+}
+
+module.exports = {
+  NEEDS_INFO_LABEL,
+  NEEDS_INFO_MARKER,
+  assessCompleteness,
+  buildNudgeComment,
+  shouldNudge,
+  shouldClearNeedsInfo,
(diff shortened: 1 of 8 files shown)


panghy (commented):
**Verdict: changes requested** (submitted as a comment review — GitHub rejects a request-changes review from the PR author's own account). Reviewed with focus on nudge idempotency, author-vs-non-author handling, PR-comment exclusion, Actions security, placeholder detection, and workflow/pure-logic parity. Ran the suite on this branch: `node --test parse-issue.test.js needs-info.test.js` → 31/31 pass. (Could not run actionlint locally — not installed.)

**What checks out**

- **Idempotency**: the marker-comment guard is checked against the *full paginated* comment list, the label check is a secondary guard, and the nudge posts the comment *before* adding the label — so a failure between the two calls still leaves the marker as the authoritative suppressor. The per-issue concurrency group covers both `issues` and `issue_comment` events (same `github.event.issue.number` expression), so the triage and reply jobs serialize per issue. Re-runs/edits cannot double-post.
- **Author handling / PR exclusion**: `shouldClearNeedsInfo` requires exact author match and a `needs-info`-labeled issue; `needs-info-reply` is gated on `!github.event.issue.pull_request`, and `triage` is gated off `issue_comment`. Correct.
- **Actions security**: clean. No event-payload strings are interpolated into script bodies via `${{ }}` — everything flows through env (integer-validated `ISSUE_NUMBER`) or `context.payload` inside JS; `issue_comment` runs check out the default branch, not attacker-controllable refs; permissions stay `contents: read` + `issues: write`.
- **Workflow/pure-logic parity**: the completeness step re-fetches the issue (fresh body + labels) and feeds `shouldNudge` exactly the shapes the tests cover; the reply job's pass-1 re-run is add-only and can never re-add `needs-triage`/`needs-info` since `labelsForIssueBody` never derives them. Stale payload labels in the reply job are safe: `removeLabel` 404 is swallowed.

**Why request-changes**: one confirmed false positive in placeholder detection (fenced-code-only sections — a very common valid bug-report style) that publicly nudges good reports and, because of the one-nudge-ever marker, permanently consumes the issue's single nudge. Details inline, plus two smaller behavioral gaps.

Per repo policy I am not merging/approving; merge requires explicit human permission.

.github/scripts/issue-triage/needs-info.js
```diff
@@ -0,0 +1,128 @@
...
+// Words that alone convey no information ("tbd", "n/a", ...).
+const PLACEHOLDER_WORDS = new Set(['tbd', 'todo', 'none', 'nil', 'na', 'idk', 'unknown']);
+
+// Minimum meaningful characters for a free-form (non-template) body.
+const NEAR_EMPTY_MIN_CHARS = 30;
+
+// True when a section's content is empty, "_No response_", or contains no
+// meaningful words (only punctuation, list numbering, or placeholder words —
+// e.g. the template's untouched "1. ...\n2. ..." placeholder).
+function isPlaceholderSection(lines) {
```
panghy, line 39: **False positive (blocking): sections whose only content is a fenced code block are assessed as placeholder.** `splitSections` (reused from parse-issue.js) drops all fenced lines — the right behavior for label derivation, but here it erases the most common way people write repro steps and expected-vs-actual: a code block of commands, logs, or a stack trace.

Verified on this branch:

```js
assessCompleteness('### Description\n\nApp crashes on launch with the new build.\n\n### Reproduction steps\n\n```\nmake run\nclick the button\n```\n\n### Expected vs actual\n\nExpected: app opens. Actual: crash dialog.\n')
// → { incomplete: true, reasons: [ 'the "Reproduction steps" section is empty or placeholder-only' ] }
```

Same result for an Expected-vs-actual section containing only a fenced error/log. Impact is doubled by the (correct) one-nudge-ever marker: the false nudge permanently consumes the issue's single nudge, so a later *genuinely* incomplete edit can never be nudged.

Suggested minimal fix: assess completeness on the raw section text (fence contents included) — e.g. add a fence-preserving mode/second return to `splitSections`, or treat a section that contains a fence opener as non-placeholder. Re-verify with a `bug-fenced-repro.md` fixture asserting `incomplete: false`.

.github/scripts/issue-triage/needs-info.js
```diff
@@ -0,0 +1,128 @@
...
+  return true;
+}
+
+// Clear `needs-info` only when the ISSUE AUTHOR comments on a labeled issue.
+// Comments from anyone else — including our own marker-carrying nudge — do
+// not clear it.
+function shouldClearNeedsInfo({ labels, issueAuthor, commentAuthor, commentBody }) {
+  if (!(labels || []).includes(NEEDS_INFO_LABEL)) return false;
+  if (!issueAuthor || !commentAuthor || commentAuthor !== issueAuthor) return false;
+  if (typeof commentBody === 'string' && commentBody.includes(NEEDS_INFO_MARKER)) return false;
```
panghy: **Minor: an author's "Quote reply" to the nudge never clears the label.** GitHub's Quote reply copies the raw markdown *including the hidden HTML marker*, so this substring check rejects a genuine author reply. Verified on this branch: quoting the full nudge (each line `> `-prefixed) and appending real repro details → `shouldClearNeedsInfo(...) === false`.

The author-equality check above already excludes the bot-posted nudge itself (it's authored by `github-actions[bot]`), so this guard only matters for issues authored by the bot — and it currently over-matches. Suggest matching the marker only when it appears as its own exact line (`commentBody.split('\n').some((l) => l.trim() === NEEDS_INFO_MARKER)` — a quote-reply line keeps its `> ` prefix after trim, so it no longer matches), or checking `comment.user.type === 'Bot'` in the workflow instead. Re-verify with a quote-reply test case.

.github/workflows/issue-triage.yml
```diff
@@ -112,3 +124,139 @@ jobs:
...
+            });
+            if (issue.state !== 'open') {
+              core.info(`issue #${issueNumber} is ${issue.state}; skipping completeness check.`);
+              return;
+            }
+
+            const assessment = needsInfo.assessCompleteness(issue.body || '');
+            core.info(`completeness: ${assessment.incomplete ? 'incomplete' : 'ok'}` +
+              (assessment.reasons.length ? ` — ${assessment.reasons.join('; ')}` : ''));
+            if (!assessment.incomplete) return;
```
panghy: **Gap: fixing the issue by *editing the body* never clears `needs-info`.** The nudge tells the author their sections are empty; the natural response is to edit the body (not comment). That `edited` event re-runs this step, which sees `incomplete: false` and returns — leaving the label in place until the author *also* comments. This run already has everything needed to close the loop: if the assessment is complete, `needs-info` is present, and the marker comment exists (i.e. we applied it), remove the label here. That stays consistent with the add-only rule, which governs pass-1 *derived* labels — the needs-info loop already removes this label in the reply job.

.github/workflows/issue-triage.yml
```diff
@@ -112,3 +124,139 @@ jobs:
               labels: toAdd,
             });
             core.info(`Added labels: ${toAdd.join(', ')}`);
+
+      - name: Needs-info completeness check
```
panghy, line 131: **Non-blocking: retroactive nudges on old issues.** This step runs on `edited`/`reopened` with no gating on issue age or triage state, so once merged, any touch of a pre-existing incomplete issue — including ones long since triaged by pass 2 or a human, or actively being worked — posts the nudge and applies `needs-info`. If that's not intended, consider gating the nudge on `needs-triage` still being present (or on `created_at`).

.github/workflows/issue-triage.yml
```diff
@@ -112,3 +124,139 @@ jobs:
...
+
+            const labels = issue.labels.map((l) => (typeof l === 'string' ? l : l.name));
+            const comments = await github.paginate(github.rest.issues.listComments, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issueNumber,
+              per_page: 100,
+            });
+            const commentBodies = comments.map((c) => c.body || '');
+            if (!needsInfo.shouldNudge({ assessment, labels, commentBodies })) {
```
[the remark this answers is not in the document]
panghy: Fixed in 91aa834. The marker no longer gates the label, and the order is reversed: the label is applied first, then the comment. shouldNudge was replaced by nudgeActions({addLabel, postComment}) — the marker only suppresses once the comment actually exists, and a label-present/marker-absent state (comment failed after labeling) retries the comment on the next run. New test covers the recovery case.

.github/workflows/issue-triage.yml
```diff
@@ -112,3 +124,139 @@ jobs:
...
+        with:
+          script: |
+            const path = require('path');
+            const scriptsDir = path.join(process.env.GITHUB_WORKSPACE, '.github/scripts/issue-triage');
+            const { labelsForIssueBody } = require(path.join(scriptsDir, 'parse-issue.js'));
+            const needsInfo = require(path.join(scriptsDir, 'needs-info.js'));
+
+            const issue = context.payload.issue;
+            const comment = context.payload.comment;
+            const labels = issue.labels.map((l) => (typeof l === 'string' ? l : l.name));
```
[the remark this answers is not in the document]
panghy: Fixed in 91aa834. The reply job now re-fetches the issue and decides on the fresh labels (and fresh issue author) instead of the event snapshot, so an author reply that races the opened-triage labeling run still clears needs-info. The single fresh fetch also feeds the pass-1 re-run.

.github/scripts/issue-triage/needs-info.js
panghy: Fixed in 6c35bcc. splitSections now takes an includeFenced option that keeps fenced lines as plain section content (fence delimiters still never derive headings), and assessCompleteness uses it — label derivation keeps the old fence-dropping behavior. Added the bug-fenced-repro.md fixture asserting incomplete: false, plus a test that a fence containing only placeholder words (tbd) is still flagged.

.github/scripts/issue-triage/needs-info.js
panghy: Fixed in 6c35bcc. Marker detection now matches only when the marker is its own exact line (isNudgeComment: split on newlines, trim, strict equality) — a quote-reply keeps the '> ' prefix so it no longer matches. Test covers the full quote-reply case (every nudge line quoted + real repro appended → clears).

.github/workflows/issue-triage.yml
panghy: Fixed in 6c35bcc. Added shouldClearOnEdit: when the assessment is complete, needs-info is present, and the marker comment exists (we applied the label), the completeness step removes it — a human-applied needs-info without our nudge is left alone. Dry-run prints the intended removal; 404 on removal is tolerated.

.github/workflows/issue-triage.yml
panghy: Fixed in 6c35bcc. Outside the opened event, nudging now also requires needs-triage to still be present, so touching an old already-triaged issue (edited/reopened after pass 2 or a human retired needs-triage) never nudges retroactively. Fresh opened issues nudge regardless (the label may not have landed yet in dry-run/manual dispatch). Tests cover both branches.

Issue #3966, opened 2026-08-31
fe: triage 35 renderer actions dispatched with no consumer (orphaned by saga-infra removal 95d908a2d)


panghy:
While confirming the root cause of #3965 (settings/specialist proposal applies are silent no-ops because the proposal-lifecycle saga was deleted and never rebuilt), a systematic scan of packages/cloudlands-fe found the same failure mode across many more actions.

Background: commit 95d908a2d ("chore: remove renderer side effects and saga infrastructure", 2026-06-26) deleted ~137 saga files wholesale; the Themis saga migration (#584 onward) rebuilt effects slice by slice, and several slices were never rebuilt. Individual casualties have been rediscovered one at a time (#1626, #1185, #3965, and the "restore X" fix series, e.g. cloudlands-fe #885/#886/#893).

Method: for each of the 918 createAction exports under src/store/renderer/slices, flag actions that are dispatched from UI code (src/lib, src/routes) but have (a) no reference in any saga/middleware file under src/store and (b) no reducer handler registration. 35 orphans found.

Orphaned actions (grouped by the deleted saga they map to):

- Changes tab git ops (changes-operations-saga, never rebuilt): stageByPathRequested, unstageByPathRequested, unstageChangesRequested, revertByPathRequested, revertChangeRequested, setCachedGitStatus — if confirmed, the staging/revert buttons are silent no-ops.
- Proposals (being fixed in #3965): applyProposalRequested, undoProposalRequested.
- PR branch lookup: requestPrBranchLookup.
- Workspace summaries: loadWorkspaceSummariesRequested, loadWorkspaceSummariesSucceeded.
- Terminals: openTerminalTabRequested, terminalCreated.
- Agent locks: recomputeAgentLocks.
- Background executor: resetExecutor, setExecutorState, startBackgroundOperation (related: #1626).
- Feature codes: fetchFeatures, deactivateFeature.
- Misc: chatQueuedRetryRecordUpdated, createWorkspaceForRepoRequested, delegateExistingTaskRequested, fetchSentryIssues, initializeNotes, initializeScripts, openNewSpaceModalRequested, openTransferModal, refreshUnreadNotes, requestHydrateModelFallbackInfo, retryLoadModels, setMultiSelectSidebarSelectedTabs, setWorkspaceMainPanel, setWorkspaceNavigationWorkspaceStatus, toggleWorkspaceCollapsedNote, toggleWorkspaceMcpServer.

[...]

PR #3970, opened 2026-08-31
docs: protocol v8.8 — daemon-owned agent lock state

panghy:
Part of intent-hq/intent#3966 (PR G1). Docs-only companion to intent-hq/intentd#1603 — the daemon-owned agent-lock computation.

Documents protocol **v8.8** (additive minor bump over 8.7):

- **`changes:agent-locks` event** (06-events.md, `changes` family): `data: { workspaceId, autoCommitEnabled, lockedAgentIds: string[], lockedFilePaths: string[] }` — self-sufficient, published by a daemon recompute worker only when the snapshot changed.
- **`file-tracking.getAgentLocks` method** (methods/change-tracking.md): `{ workspaceId } → { autoCommitEnabled, lockedAgentIds, lockedFilePaths }` — the hydration read — plus the full lock-semantics block (effective auto-commit gate, unstaged/staged attribution rows, actively-working session or non-terminal linked task, retired/deleted exclusion, sorted+deduped arrays, fail-soft empty snapshot).
- **Version history entry** (versioning.md) and version bump (versioning.md, README.md). Like the rest of the hyphenated `file-tracking.*` namespace, the new method sits outside the golden catalog counts, which stay 298 router / 38 fast-path / 338 total.

No submodule pin changes (auto-bump owns those).


panghy (commented):
**Verifier review** (docs companion to intent-hq/intentd#1603, task 0629fd2c).

Checked the four touched files against the intentd#1603 implementation:

- ✅ `versioning.md`: v8.8 entry is accurate — additive minor bump, method + event named correctly, lock semantics match the implementation (effective auto-commit + working-stage tracked change + actively-working, retired/deleted never lock, sorted/deduped arrays, fail-soft empty snapshot, diff-gated debounced emit), and the "no frozen method-catalog count change / 298 router, 38 fast-path, 338 total" note matches the code (the hyphenated `file-tracking.*` namespace sits outside the golden counts).
- ✅ `README.md`: protocol version bumped to 8.8 consistently.
- ✅ `methods/change-tracking.md` (§5.19): `file-tracking.getAgentLocks` row documents `workspaceId` (req) → `{ autoCommitEnabled, lockedAgentIds, lockedFilePaths }`, matching the router arm and `AgentLockSnapshot::to_result_value` exactly; lock-semantics block matches `is_actively_working` / `compute_agent_locks`.
- ✅ `06-events.md` (§6.5): `changes:agent-locks` documented in the `changes` family with the self-sufficient payload `{ workspaceId, autoCommitEnabled, lockedAgentIds, lockedFilePaths }` and the changed-only debounced-emit contract — matches `changes_agent_locks_event` and `refresh_agent_locks`.

No pin bumps in the diff; docs-only, conventional PR title, CI green.

One consistency note tied to the intentd review: I flagged on intentd#1603 that the recompute worker does not subscribe to `agent:retired` / `agent:restored`, so the documented "retired/deleted sessions never lock" property can be served stale until an unrelated event triggers a recompute. If the fix lands as an added subscription (recommended), no docs change is needed here — the documented contract is the intended behavior. Only if the behavior is instead accepted as-is would this doc need a caveat.

panghy:
Re the consistency note: the fix landed on intent-hq/intentd#1603 as the recommended added subscription — commit 33ef7562 adds `agent:retired` / `agent:restored` to `spawn_agent_locks_loop`'s `SubscriptionFilter`, with a loop test covering retire→unlock and restore→relock. The documented "retired/deleted sessions never lock" contract is therefore the actual behavior and no docs caveat is needed here.

Issue #4058, opened 2026-09-01
feat(intentd): expose upstream-tracking info (hasUpstream / unpushed count) on git.status


panghy:
## Context

While bridging the legacy workspace hover-card summary channels to the daemon (intent-hq/cloudlands-fe#1985, workspace-summaries-bridge-seeder.ts), the FE needed the legacy "hasUnpushed" signal: commits ahead of the branch upstream, falling back to "ahead of baseRef" when no upstream exists.

## Problem

The daemon cannot express this today. "git.status" (docs/protocol/methods/git.md, section 5.6) reports ahead/behind vs refs/remotes/origin/<currentBranch> and returns (0, 0) when the upstream ref does not exist (intent-git status.rs ahead_behind) — indistinguishable from "even with upstream". So the FE bridge approximates: hasUnpushed = status.ahead > 0, with a base-ahead fallback. For never-pushed branches (the common pre-PR case) this reports hasUnpushed = false where the legacy implementation reported true.

## Proposal

Additive fields on the "git.status" result (no protocol bump needed per prior additive precedent), e.g.:
- "hasUpstream": boolean — whether refs/remotes/origin/<branch> (or the configured upstream) exists
- optionally "unpushedCount": number — rev-list count upstream..HEAD, omitted when no upstream

The FE bridge can then drop its documented approximation (see the PR body "documented approximations" list on cloudlands-fe#1985).

## References
- FE consumer: packages/cloudlands-fe/src/store/renderer/seeders/workspace-summaries-bridge-seeder.ts (hasUnpushed approximation documented in the header)
- Daemon: packages/intentd/crates/intent-git/src/status.rs (ahead_behind), branches.rs (ahead_behind_vs_origin)
- Raised during PR review of intent-hq/cloudlands-fe#1985

panghy:
Taking this in the assignee-type coordination lane as a two-part batch: (1) intentd — additive hasUpstream + unpushedCount on git.status (unpushedCount omitted when no upstream), riding the existing ahead_behind resolution in status.rs so no new RPC cost class; protocol docs (docs/protocol/methods/git.md §5.6) updated in a companion monorepo docs PR. (2) cloudlands-fe — the workspace-summaries bridge seeder drops the ahead>0 approximation and restores legacy semantics (no upstream → hasUnpushed = ahead-of-baseRef > 0), with a fallback for pre-upgrade daemons. intentd lands first per protocol ordering; the FE PR will carry the Fixes reference since it completes the loop.

[...]

PR #4064, opened 2026-09-01
docs: document git.status hasUpstream + unpushedCount upstream-tracking fields

panghy:
Part of intent-hq/monorepo#4058

Documents the additive `git.status` upstream-tracking fields introduced by [intent-hq/intentd#1622](https://github.com/intent-hq/intentd/pull/1622):

- `hasUpstream` (boolean, always present) — whether `refs/remotes/origin/<branch>` exists.
- `unpushedCount?` (number) — the `upstream..HEAD` commit count (= `ahead`); present even at `0` when the upstream exists, omitted entirely when there is none, so absence is distinguishable from zero and pre-#4058 clients see the previous shape byte-for-byte.

Changes:

- `docs/protocol/methods/git.md` (§5.6): the `git.status` result row gains the two fields, plus a new "`git.status` upstream tracking" contract block (semantics, detached/unborn HEAD behavior, and the no-new-cost-class statement — the fields ride the same repository read the method already performs).
- `docs/protocol/versioning.md`: "Also within 9.0" entry — additive presence-detected response fields, no method-catalog change, so no version bump.

Docs-only; no code changes. Should merge after the intentd PR lands (daemon-first rule for protocol changes).


panghy (commented):
Cross-checked field-by-field against the implementation in [intent-hq/intentd#1622](https://github.com/intent-hq/intentd/pull/1622) (b0be5c6). Everything matches:

- **Field names / wire shape**: `hasUpstream` (always-present bool) and `unpushedCount?` match the camelCase serialization of `GitStatus.has_upstream` (`#[serde(default)]`, no skip) and `unpushed_count` (`skip_serializing_if = "Option::is_none"`). The §5.6 result-row placement after `totalFiles?` matches struct field order.
- **Omission semantics**: "present even at `0` when the upstream exists, omitted entirely when there is none" is exactly `has_upstream.then_some(ahead)`, and the implementation's `upstream_fields_wire_shape` test asserts both sides (I ran it locally — passes).
- **`unpushedCount == ahead`**: holds unconditionally in the implementation, including the `graph_ahead_behind` error arm (both fall back to `0` there).
- **Detached/unborn HEAD**: `hasUpstream: false` + `unpushedCount` omitted matches the `branch.is_empty()` / `repo.head()` guards, each covered by a dedicated unit test.
- **`refs/remotes/origin/<branch>`**: correctly documents the origin-hardcoded convention the existing `ahead`/`behind` computation already uses (a branch tracking a non-origin remote reads `hasUpstream: false`) — faithful to the implementation.
- **No new cost class**: accurate — the fields ride the same `find_reference`/`graph_ahead_behind` calls that already produce `ahead`/`behind`.
- **versioning.md**: the "Also within 9.0" entry follows the established convention for presence-detected additive response fields with no method-catalog change (no version bump), with both cross-repo links; style is consistent with neighboring entries.

Sequencing note from the PR body is correct: per the daemon-first rule this should merge only after intentd#1622 lands.

Not approving for merge per policy — humans own merging.

PR #4095, opened 2026-09-01
docs(protocol): document tcAddress, tc= URI param, and server.tunnel.* settings

panghy:
Documents the Tailcat tunnel wire-contract additions from intent-hq/intentd#1618 and intent-hq/intentd#1623 in `docs/protocol`. Docs-only; no code changes.

## `docs/protocol/05-method-catalog.md`

- `system.status`: `tcAddress` added to the example + a dedicated bullet — additive (intentd#1623), served to local **and** remote callers alongside `localIps` so a connected client can refresh its stored tunnel route; present only while the tunnel sidecar is running, **omitted** (never `null`) when the tunnel is disabled or down, including the restart-backoff window; detect by presence.
- `pairing.getInfo`: the `tc=` query param documented as the **final** URI param (percent-encoded, absent when no tunnel), `tcAddress` added to the component fields, plus the unknown-query-param tolerance note for clients that predate it.
- `server.pairingInfo`: `tcAddress` in the example + bullet with the same field/availability semantics as the other two surfaces.

## `docs/protocol/methods/settings.md` (§5.12 server settings)

- `server.tunnel.enabled` (bool, default `false`, TOML `[server.tunnel]`, intentd#1618): runs the bundled tailcat sidecar; batch hook ordering (applies **after** `server.wsApi.enabled`); live-applied — enable starts the sidecar (start failure fails the update with `-32603`), disable stops it; disabling `server.wsApi.enabled` also stops a running tunnel without touching the persisted key (re-enable brings it back; skipped when the batch carries its own tunnel toggle); disambiguated from the §1.4 `/tunnel` endpoint.
- `server.tunnel.derpUrl` (optional string): self-hosted DERP relay; live-applied via sidecar restart when running, else persisted for the next start.
- `server.tunnel.only` (bool, default `false`): loopback-only WSS bind; live-applied via listener restart (mirrors the `server.bindAddress` hook); pre-persist `-32602` self-termination guard rejecting enable-over-TCP — the guard keys on the raw transport, so tunnel-forwarded connections are rejected too (flip it over UDS).

## Verification

Every claim was checked against the merged intentd code on the current pin: settings registry (keys/defaults/descriptions), pairing/server/control serializers (`tcAddress` / `tc=` spelling, omit-not-null, final-param position), tunnel supervisor (stable address via persisted key, restart backoff), settings hooks (`hook_priority` batch ordering, live-apply arms, both TCP guards), and the pairing tests (URI shape, absent-not-null).


docs/protocol/05-method-catalog.md
```diff
@@ -293,17 +295,19 @@ Returns the structured QR pairing payload so local clients (the `intentd pair` C
...
   "fingerprint": "AA:BB:...",
   "token": "abab...",
-  "version": 1
+  "version": 1,
+  "tcAddress": "tc7f2a91.tailcat.net"
 }
 ```
 
-- `uri` is the plaintext payload encoded in the QR code: `intent://pair?v=1&host=<ip[,ip...]>&port=<p>&fp=<sha256>&token=<t>` (query values percent-encoded where needed). The component fields (`hosts`, `port`, `fingerprint`, `token`, `version`) are provided so clients can render their own payloads.
+- `uri` is the plaintext payload encoded in the QR code: `intent://pair?v=1&host=<ip[,ip...]>&port=<p>&fp=<sha256>&token=<t>[&tc=<addr>]` (query values percent-encoded where needed). The component fields (`hosts`, `port`, `fingerprint`, `token`, `version`, `tcAddress`) are provided so clients can render their own payloads.
```
panghy: Nit (non-blocking): listing `tcAddress` among "The component fields (…) are provided" reads as always-present, while unlike the other five it is conditional (omitted when the tunnel is down). The next bullet does clarify, but a hedge here would remove the tension, e.g. "…`token`, `version`, and — when the tunnel is up — `tcAddress`) are provided…".

docs/protocol/methods/settings.md
```diff
@@ -54,7 +54,7 @@ the overriding flag ("overridden by startup flag …").
...
...nning WSS listener with the new bind set (mirroring the `server.bindAddress` hook). Self-termination guard: enabling it from a direct TCP (WSS) connection is rejected with `-32602` **before anything persists** — the loopback rebind would drop the caller's own connection — so flip it over UDS. The guard keys on the raw transport (like the `server.wsApi.enabled` disable guard), so tunnel-forwarded connections — which arrive at the listener over TCP — are rejected too)*, `server.tls.enabled`, `server.auth.enabled`,`server.auth.token` *(sensitive; read-only / regenerate)*, `server.originAllowList`, `server.maxOutstandingRpcs` *(number, default `256`, min `0`, max `100000`, TOML-backed under `[server]` — the daemon-wide cap on outstanding slow-path RPCs shared across every connection and both transports; over-limit requests are rejected with `-32011` "Server overloaded" (§9) and `0` disables the cap. Read once at boot: a change applies on daemon restart)*. The UDS listener always serves; the TCP/WSS listener is toggled at runtime by `server.wsApi.enabled` (the former `server.listenMode` key is retired — a config.toml still carrying it boots, is discarded, and is stripped from the file).
```
panghy: Two non-blocking notes on the `server.tunnel.derpUrl` entry:

1. "tailcat's default Tailscale-hosted relay" — the settings-catalog description in intentd says only "tailcat's default relay"; "Tailscale-hosted" is an added (plausible, but uncited) claim. Consider dropping the qualifier to stay exactly aligned with the registry description.
2. Mechanically the value is passed to the sidecar as `--derpmap-url` (a DERP *map* URL that resolves the relay region — see `serve_command` in `intentd/src/tunnel.rs`), not a direct relay URL. The doc matches intentd's own catalog wording ("self-hosted DERP relay URL"), so this is consistent as-is — flagging only in case you want the more precise "DERP map URL" phrasing.

panghy (commented):
Reviewed against the merged intentd code on the current monorepo pin (intentd tip `fe8fa96f`). Every technical claim I checked holds:

- **Settings registry** — `server.tunnel.enabled` (bool, default `false`), `server.tunnel.derpUrl` (optional string), `server.tunnel.only` (bool, default `false`), all TOML-backed under `[server.tunnel]` (`TunnelSettings` in `intent-core/src/settings_file.rs`, camelCase → `derpUrl`); descriptions match `intent-services/src/settings.rs`.
- **Hook ordering** — `hook_priority` in `intent-services/src/lib.rs`: value keys 0, `server.wsApi.enabled` 10, `server.tunnel.enabled` 20 — the tunnel toggle applies after the listener toggle, exactly as documented.
- **Live-apply arms** — enable starts the sidecar and wraps any start failure in `Error::Internal` (`-32603`); disable stops it. The `wsApi.enabled=false` arm stops a running tunnel without touching the persisted key; the `=true` arm restarts it from the persisted value, non-fatal (warn) like the boot auto-start, and is skipped when the batch carries its own tunnel toggle (`batch_has_tunnel_enabled`). `derpUrl` restarts a running sidecar, else logs "persisted (tunnel not running)". `tunnel.only` restarts the listener mirroring the `bindAddress` hook.
- **Guards** — pre-flight `server.tunnel.only` enable-over-TCP guard returns `Error::InvalidParams` (`-32602`) before old-value capture/persist; keyed on `control.is_tcp_connection()`, same predicate as the `wsApi.enabled` disable guard, so tunnel-forwarded connections (TCP at the listener) are rejected too.
- **Serializers** — `tcAddress` spelling and omit-not-null confirmed in all three surfaces (`control.rs`, `server.rs`, `pairing.rs`); `tc=` appended last in `build_pairing_uri`, percent-encoded, absent when no tunnel — the pairing tests assert both the final-param position and absent-not-null, and the doc's example address `tc7f2a91.tailcat.net` matches the test fixtures.
- **Restart-backoff omission** — `up.store(false)` in the supervisor's restart loop (commit `fe8fa96f`) backs the "including the restart-backoff window" claim; the stable-address-via-persisted-key claim matches `tunnel.rs` (`tunnel/tailcat.private.json`).
- **Cross-refs** — the three method subsections sit under the §5 fast-path catalog; the §1.4 `/tunnel` disambiguation and §5.12 pointers resolve correctly; additive-field marking with PR links follows the `prettyHostname` (intentd#1466) precedent.

Two non-blocking nits inline. Not approving/merging per workflow — leaving that to a human.

docs/protocol/05-method-catalog.md
panghy: Fixed in ed70814 — the sentence now hedges the conditional field: "…`token`, `version`, and — when the tunnel is up — `tcAddress`) are provided…".

docs/protocol/methods/settings.md
panghy: Both addressed in ed70814. Reworded to the mechanically precise form: the value is a DERP **map** URL passed to the sidecar as `--derpmap-url` (serving a DERP-map JSON document that resolves the relay region for a self-hosted relay setup), it also feeds key generation for region discovery, and the uncited "Tailscale-hosted" qualifier is dropped ("empty/unset uses tailcat's default relay", matching the registry description).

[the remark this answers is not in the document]
panghy: Fixed in ed70814 — the entry now describes the value as a DERP **map** URL passed as `--derpmap-url` (a DERP-map JSON document), not a bare relay endpoint, so users won't point it at a `derper` URL the sidecar cannot use.

PR #4128, opened 2026-09-01
docs(protocol): document idle-deferred attention-request surfacing

panghy:
Documents the idle-deferred attention-surfacing timing change from intent-hq/intentd#1639.

> **Merge sequencing:** intent-hq/intentd#1639 is still open — per the intentd-first rule, do not merge this docs PR until that intentd PR is confirmed merged.

## What changed

A discussion request or blocker raised mid-turn (`ws.agent.requestDiscussion` / `ws.agent.reportBlocker`) no longer surfaces to the user at tool-call time. The user-facing surfacing bundle — the `agent:attention-requested` event, the paired `agent:updated` attention-fields emit, the transcript notice, and the `displayStatus` promotion — is parked on an in-memory deferred-attention marker and flushed when the raising agent goes idle (clean prompt-turn settlement, harness-wake idle, user interrupt, or terminal turn failure), ordered before the paired `agent:idle` / `agent:failed` (one exception: the §6.6 idle-timeout-cap failure, where the drain loop's own `agent:failed` precedes the terminal-failure handler's flush). The session persist, linked-task transition, parent wake, and watcher fan-out stay immediate. A raise with no in-flight turn surfaces immediately as before, and a request cleared before its flush never surfaces.

## Docs updated

- `docs/protocol/methods/agents.md` (§5.5) — new "Idle-deferred surfacing" block after the attention-request step list; steps 1–3 annotated with the deferred timing; step-list count corrected to six.
- `docs/protocol/06-events.md` (§6.5) — `agent:attention-requested` row documents the idle-deferred emit; `workspace:displayStatus-changed` row documents the raise-side recompute moving to the turn-end flush.
- `docs/protocol/methods/workspace.md` (§5.1) — the attention-axes derivation intro notes that a parked mid-turn request does not feed the derivation until the flush.

Docs-only change; no protocol version bump (behavior/timing documentation for the intentd change).


docs/protocol/methods/agents.md
```diff
@@ -359,6 +368,28 @@ delegated or not, with or without a linked task. `reason` is required (trimmed;
...
+while the **user-facing surfacing bundle** — the `agent:attention-requested` event
+(step 3), the paired `agent:updated` attention-fields emit (step 1), the transcript
+notice (step 2), and the `displayStatus` recompute/promotion — is parked on an in-memory
+deferred-attention marker and flushed when the raising agent goes idle: clean prompt-turn
+settlement, harness-wake idle, user interrupt, or terminal turn failure, always ordered
+BEFORE the paired `agent:idle` / `agent:failed` emit so subscribers never see a quiet
+idle that later grows an attention card. A raise with no in-flight turn surfaces
+immediately, as before. While the marker is parked the workspace's `displayStatus`
+derivation skips the pending request (the workspace reads `in_progress` while the
+raising turn runs) and promotes to `blocked` / `needs_attention` exactly at the flush
```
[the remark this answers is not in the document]
panghy: Addressed in 0cc7b6e: all three locations now state the request feeds the derivation from the flush onward, subject to the ordinary §5.1 precedence and eligibility rules — the in_progress → blocked/needs_attention sequence is described as the typical top-level-foreground case, with the terminal-failure failed outrank and the child/background ineligibility called out explicitly.

panghy (commented):
Verified the docs against intentd commit 93599638 (`feat/defer-attention-surfacing`, the intent-hq/intentd#1639 branch). The substance checks out:

- **Op split** — the immediate half (step-1 session persist, step-4 linked-task transition, step-5 parent wake, step-6 watcher fan-out) vs the deferred surfacing bundle (`agent:attention-requested`, the paired `agent:updated` raise, the transcript notice, the `displayStatus` recompute) matches `agent_request_attention_op` / `surface_attention_request` / `flush_deferred_attention` exactly, including the `agent_is_busy` branch (no in-flight turn → immediate surfacing).
- **Flush sites** — all four documented sites exist and (with one exception, see inline comment) order the flush before the paired idle/failed emit: `run_prompt_turn` settlement + its error arm (agent_session.rs), harness-wake idle, the interrupt path (guarded by `!suppress_idle_emit`, matching the interrupt-with-message carve-out), and `handle_terminal_turn_failure`.
- **Cleared-before-flush** — the turn-begin clear consumes the marker (`take_deferred_attention`) before publishing the plain `attentionRequestCleared` emit; `flush_deferred_attention` also retires silently when the session fields are already gone. Matches the docs.
- **Restart caveat** — the registry is a plain in-memory `HashSet` in `Services`; the persisted session fields keep feeding reads and (post-restart, marker gone) the `displayStatus` derivation. Matches.
- **workspace.md §5.1** — `workspace_attention_signals` skips sessions with `attention_surfacing_deferred`, and the flush consumes the marker before the recompute, so the "promotes exactly at the surfacing point" claim is accurate.
- Cross-file consistency between §5.5, the two §6.5 rows, and §5.1 is good; step-number references (1–6) match the §5.5 list.

Two non-inline notes:

1. **Nit (line outside this diff)**: the §5.5 intro above the step list still says the shared op "does five things:" while the list has six steps — pre-existing, but since the new block leans on the step numbering (steps 1–6), this PR is a natural place for the one-word fix.
2. **Sequencing**: intent-hq/intentd#1639 is still **OPEN** (not merged) as of this review, while the PR body says the change "landed". Per the intentd-first rule for protocol docs, this docs PR should not merge until intentd#1639 is confirmed merged (and the body's "landed" wording is accurate at that point).

Not approving/enqueuing — review comments only.

docs/protocol/methods/agents.md
```diff
@@ -359,6 +368,32 @@ delegated or not, with or without a linked task. `reason` is required (trimmed;
...
+[intent-hq/intentd#1639](https://github.com/intent-hq/intentd/pull/1639))*.** A raise from
+inside a live turn does NOT surface to the user at tool-call time. The op splits in two:
+the **immediate** half — the step-1 session persist, the step-4 linked-task transition,
+the step-5 parent wake, and the step-6 watcher fan-out — runs at the raise as before,
+while the **user-facing surfacing bundle** — the `agent:attention-requested` event
+(step 3), the paired `agent:updated` attention-fields emit (step 1), the transcript
+notice (step 2), and the `displayStatus` recompute/promotion — is parked on an in-memory
+deferred-attention marker and flushed when the raising agent goes idle: clean prompt-turn
+settlement, harness-wake idle, user interrupt, or terminal turn failure, always ordered
+BEFORE the paired `agent:idle` / `agent:failed` emit so subscribers never see a quiet
```
panghy, line 381: The "always ordered BEFORE the paired `agent:idle` / `agent:failed` emit" claim has one exception in the landed code: the prompt-idle-timeout **cap** path (§6.6). When the consecutive-timeout cap is spent, the turn worker's drain loop publishes `agent:failed` itself and only then calls `handle_terminal_turn_failure` — whose `flush_deferred_attention` is unconditional but runs after that emit (the handler skips re-publishing the pair via `turn_failure_events_already_emitted`, yet the flush still lands behind the drain loop's own `agent:failed`). So a request raised mid-turn by an agent that then hangs into the timeout cap surfaces AFTER its `agent:failed`, not before. This path is reachable (raise → hung `session/prompt` → cap spent) and the §6.5 lifecycle row already documents that the cap path emits the failed half itself.

Suggest softening "always ordered BEFORE" here (and in the §6.5 `agent:attention-requested` row, which makes the same unqualified claim) with a parenthetical noting the idle-timeout-cap exception — e.g. "(one exception: the §6.6 idle-timeout-cap failure, where the drain loop's own `agent:failed` precedes the terminal-failure handler's flush)".

panghy: Verified against the drain loop's cap arm in agent_manager.rs (it publishes agent:failed itself, then calls handle_terminal_turn_failure whose flush runs after) — you're right. Addressed in fd29967: both the §5.5 block and the §6.5 agent:attention-requested row drop "always" and carry the suggested idle-timeout-cap parenthetical.

panghy:
Addressed all review findings: the ordering exception is documented in fd29967 (§5.5 + §6.5 row, thread resolved), the step-count nit is fixed in cb3a298 ('does six things:'), and the PR body now says the intentd change is from the still-open intent-hq/intentd#1639 with an explicit merge-sequencing note (intentd-first: do not merge this docs PR until #1639 lands).

Issue #1672, opened 2026-08-07
cloudlands-fe: Settings beta-updates toggle desyncs from actual updater channel (dead real-mode hydration + persistence echo)


panghy:
## Summary

The Settings UI can show "Beta updates" as **off** while the auto-updater is actually on the **beta** channel. Observed in the wild: a machine with `betaUpdatesEnabled: true` in `local-prefs.json` (both `intent` and `intent-cloudlands` userData dirs) pulled cloudlands-fe 2.19.0 from the beta feed ~1h before it was promoted to stable, while the Settings toggle displayed off.

## Root cause (two connected defects)

**1. Real-mode hydration of the toggle is dead.**
`live-settings-client.ts#getUserPreferences()` deliberately returns `null` in real (daemon-connected) mode, so `settings-integrations-seeder.ts`'s `if (prefs)` block — the code that syncs `betaUpdatesEnabled` from `autoUpdateClient.getState()` — is skipped entirely. Redux keeps its default `betaUpdatesEnabled: false` and the UI renders "off" regardless of the actual channel.

**2. Fallback hydration races and echoes.**
`user-preferences-beta-persistence-service.ts` hydrates only "on first action" via a fire-and-forget `autoUpdateClient.getState()` read:
- It races main-process `AutoUpdateService.initialize()`; if the IPC read lands before `loadChannelFromSettings()` completes, it sees the default `channel: 'stable'` and dispatches `false`.
- The hydration dispatch is itself a `setBetaUpdatesEnabled` action, which the same middleware treats as a user toggle and calls `autoUpdateClient.setChannel(...)`, which **writes back** to `local-prefs.json`. This echo can silently flip the persisted channel in either direction depending on race timing.

## Impact

- Users on the beta channel believe they are on stable (and vice versa); beta builds install ahead of stable promotion with `autoDownload` + `autoInstallOnAppQuit` always on.
- The persisted channel preference can be corrupted at boot by the hydration echo.

## Affected code (cloudlands-fe)

- `src/lib/client/live/live-settings-client.ts` (`getUserPreferences` returns null)
- `src/store/renderer/seeders/settings-integrations-seeder.ts` (sync gated behind `if (prefs)`)
- `src/store/renderer/middlewares/user-preferences-beta-persistence-service.ts` (first-action hydration race + echo persist)
- `src/features/auto-update/main/auto-update.service.ts` (channel source of truth: `local-prefs.json` via `loadChannelFromSettings`)

## Proposed fix

[...]

PR #4251, opened 2026-09-03
docs(protocol): bind-aware localIps, loopback-free pairing hosts, and the no-dialable-route pairing error

panghy:
Updates `docs/protocol/05-method-catalog.md` to match the advertised-route behavior now on intentd main after [intent-hq/intentd#1656](https://github.com/intent-hq/intentd/pull/1656) and [intent-hq/intentd#1672](https://github.com/intent-hq/intentd/pull/1672).

## What changed

- **`system.status` routing fields**: `localIps` is documented as bind-aware — only addresses the running WSS listener actually answers on (specific bind → exactly those, loopback included on this diagnostic surface; `0.0.0.0` → non-loopback IPv4 enumeration; `::` → IPv4 + global IPv6, dual-stack listener), **empty** with no live TCP listener instead of the historical full enumeration. Dropped the stale "same list as `server.pairingInfo`" claim and clarified that only the interface enumerations are cached (~15s TTL) while the bind-set filter runs live.
- **`pairing.getInfo`**: new `hosts` bullet — bind-aware and **never containing loopback** even when explicitly bound; a loopback-only bind with the tunnel up pairs hostless (`host=` empty, `tc=` carries the route). Error prose restructured into the two `-32603` variants: listener-down (with the `error.data.code = "listener-down"` discriminator, unchanged) and the new **no-dialable-route** error with actionable guidance (loopback-only bind with no tunnel names both remediations; the no-non-loopback-IPv4 enumeration case keeps its plain message). Notes the behavior change: pre-#1672 a loopback-only bind returned a success payload advertising loopback.
- **`server.pairingInfo`**: `localIps` documented as the same dialable pairing host set as `pairing.getInfo` `hosts` (loopback-free, bind-aware, may be empty), explicitly contrasted with the diagnostic `system.status` `localIps`.

## Not documented (verified non-wire)

The tailcat "picker note" from intentd#1672 is a CLI-only stderr line in the `intentd pair` interactive picker, not a JSON-RPC field — pairing payloads on main carry only `tcAddress` as the tunnel-related addition, so no new field is documented.

## Verification

Every claim checked against the merged intentd diffs (`gh pr diff 1656 / 1672`) and current main sources: `crates/intent-transport/src/pairing.rs` (error variants, exact messages, hostless URI), `crates/intent-transport/src/server.rs` (`pairing_hosts` loopback filter, `advertised_hosts` bind-set semantics), `crates/intentd/src/main.rs` (status route snapshot: live bind filter over cached enumerations, empty when listener down), plus the updated e2e tests in `e2e_wss_runtime_control.rs` / `e2e_wss_server_pairing.rs`. Error code `-32603` for both variants confirmed via `intent-core/src/error.rs` (`Unsupported` and `ListenerDown` both map there). Other protocol docs (01-transport, 02-authentication, 09-error-codes, 10-thin-client) checked — their existing mentions remain accurate.

Docs-only monorepo change — no release monitoring needed. Follow-up to [#4095](https://github.com/intent-hq/intent/pull/4095).


docs/protocol/05-method-catalog.md
```diff
@@ -306,10 +306,12 @@ Returns the structured QR pairing payload so local clients (the `intentd pair` C
...
-- Hosts, TLS fingerprint, and bearer token come from the same sources as `intentd pair`, so all pairing surfaces stay consistent.
+- `hosts` is the **dialable pairing host set** — bind-aware and **never containing loopback** ([intent-hq/intentd#1656](https://github.com/intent-hq/intentd/pull/1656), [intent-hq/intentd#1672](https://github.com/intent-hq/intentd/pull/1672)): a listener bound to specific addresses advertises exactly those, an unspecified bind (`0.0.0.0` / `::`) enumerates the machine's local addresses (virtual/container interfaces skipped; `::` adds global IPv6 — its listener is dual-stack), and loopback (`127.0.0.1` / `::1`) is filtered out even when explicitly bound: pairing hosts feed remote devices, and loopback is not dialable from another device. With a loopback-only bind and the tunnel **up**, `hosts` is **empty** and the payload pairs **hostless** — the `host=` URI param is empty and the `tc=` route alone carries the connection. TLS fingerprint and bearer token come from the same sources as `intentd pair` and `server.pairingInfo`, so all pairing surfaces stay consistent.
```
panghy: Nit: the `uri` template on the line above still reads `host=<ip[,ip...]>`, which implies at least one IP. Since this bullet now documents the legitimate empty-`host=` (hostless) form, consider annotating the template, e.g. `host=<ip[,ip...]>` → "`host=` may be empty (hostless pairing, see the `hosts` bullet)". Purely optional — this bullet does cover it.

docs/protocol/05-method-catalog.md
```diff
@@ -306,10 +306,12 @@ Returns the structured QR pairing payload so local clients (the `intentd pair` C
...
+  - **No dialable route** ([intent-hq/intentd#1672](https://github.com/intent-hq/intentd/pull/1672)) — the listener is up but the advertised host set is empty and no tunnel is active. A loopback-only `server.bindAddress` gets actionable guidance naming both remediations: `"listener is bound to loopback only and no tunnel is active — set server.bindAddress to a LAN address or enable the tunnel (server.tunnel.enabled) before pairing"`. An unspecified/unknown bind whose enumeration found no non-loopback IPv4 address keeps the plain `"no non-loopback IPv4 address found — connect this machine to a network before pairing"`. Neither no-route variant carries `error.data`. (Before intentd#1672 a loopback-only bind returned a success payload advertising loopback; clients must now handle the error.)
```
panghy: Nit: "An unspecified/unknown bind whose enumeration found no non-loopback IPv4 address" — for a `::` bind the enumeration also includes global IPv6, so the condition is really "found no hosts at all" (no non-loopback IPv4 *and* no global IPv6). The daemon's message text only names IPv4 (source wording, quoted correctly), but the framing prose here could avoid restating IPv4 as the trigger condition, e.g. "whose enumeration found no dialable address". Minor; the current text is only imprecise for the `::` case.

panghy (commented):
Reviewed against merged intentd main (intent-hq/intentd#1656 / #1672 diffs, `crates/intent-transport/src/pairing.rs`, `server.rs` — `pairing_hosts` / `advertised_hosts` / `collect_local_ipv6s`, `crates/intentd/src/main.rs` route snapshot + 15s sampler, `intent-core/src/error.rs`).

**Verified accurate:**
- `system.status` `localIps`: bind-aware via `advertised_hosts`, loopback kept on this surface, empty when `port` is `None` (UDS-only / stopped WSS), enumerations from the 15s TTL cache with the bind-set filter on the read path — all match `main.rs` exactly.
- `pairing.getInfo` `hosts`: `pairing_hosts` applies the loopback `retain` filter after `advertised_hosts`, so loopback never appears even when explicitly bound; hostless pairing with the tunnel up is confirmed by the e2e test asserting `intent://pair?v=1&host=&port=…&tc=…` (empty `host=` param) and `hosts: []`.
- Errors: both variants map to `-32603` (`ListenerDown` and `Unsupported` both map there in `error.rs`); both quoted messages are byte-identical to the source strings; only listener-down carries `error.data = { "code": "listener-down" }`; the pre-#1672 behavior note (loopback-only bind previously returned a success payload advertising loopback) matches the tests #1672 rewrote.
- `server.pairingInfo` `localIps`: same `pairing_hosts` call as `pairing.getInfo`, may be empty, no error on empty (correctly not claimed) — and the contrast with the diagnostic `system.status` surface matches the source comments.
- Style: inline PR reference links, presence-detection phrasing, and bolding match the surrounding doc conventions; no stale contradicting mentions remain in the other protocol docs.

**Non-blocking consideration:** `docs/protocol/versioning.md` has precedent for recording behavior-only, client-visible changes as "Also within X.Y …" entries (e.g. the intentd#1065 listener-down discriminator). The #1672 success→error flip on `pairing.getInfo` for a loopback-only bind is a client-visible contract change and could get such an entry. The predecessor doc PR #4095 also skipped versioning.md, so precedent is mixed — leaving this as a suggestion, not a blocker.

Verdict: accurate and complete; two inline nits, no blockers.

docs/protocol/05-method-catalog.md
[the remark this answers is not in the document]
panghy: Fixed in e5ca33f — verified against `collect_local_ipv6s` on intentd main: the enumeration skips only loopback and link-local (`fe80::/10`, unusable without a zone index), so ULA (`fd00::/8`) addresses are advertised. All three locations (`system.status` line 182, `pairing.getInfo` line 309, `server.pairingInfo` line 339) now say "non-link-local IPv6" instead of "global IPv6", with the line-182 bullet spelling out the link-local/ULA distinction.

docs/protocol/05-method-catalog.md
```diff
@@ -334,7 +336,7 @@ Returns the raw pairing/connection material — bearer token, TLS cert fingerpri
...
 - `port` is the bound WSS port, or `null` when the TCP (WSS) listener is not running; `path` is always `"/ws"`.
-- `localIps` lists non-loopback IPv4 addresses (virtual/container interfaces such as `docker*`/`veth*` are skipped) — the same host set `pairing.getInfo` reports, so all pairing surfaces stay consistent.
+- `localIps` is the same **dialable pairing host set** `pairing.getInfo` reports as `hosts` ([intent-hq/intentd#1656](https://github.com/intent-hq/intentd/pull/1656), [intent-hq/intentd#1672](https://github.com/intent-hq/intentd/pull/1672)): bind-aware (a specific bind advertises exactly its addresses; an unspecified bind enumerates local addresses, with virtual/container interfaces such as `docker*`/`veth*` skipped and global IPv6 added for `::`) and **never containing loopback**, even when loopback is explicitly bound. **May be empty** (e.g. a loopback-only bind) — with the tunnel up, `tcAddress` then carries the only dialable route. Note the contrast with `system.status` `localIps`, which is the diagnostic surface and keeps bound loopback entries.
```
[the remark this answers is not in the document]
panghy: Fixed in e5ca33f — confirmed against `stop_ws_listener` in `crates/intentd/src/main.rs` (sets `bind_addresses = None` and `port = None`), so a listener-down `server.pairingInfo` re-enumerates non-loopback IPv4 while `port` is `null`. The bullet now scopes the "same set as `pairing.getInfo` `hosts`" claim to while the listener is up, and documents the listener-down shape explicitly: `port: null` is the listener-down signal, and `localIps` is address material rather than proof of a live route (unlike `pairing.getInfo`, which errors listener-down).

docs/protocol/05-method-catalog.md
panghy: Done in e5ca33f — the template line now annotates that `host=` may be empty (the hostless form), pointing at the `hosts` bullet.

docs/protocol/05-method-catalog.md
panghy: Done in e5ca33f — the framing prose now reads "whose enumeration found no dialable address" (the quoted daemon message keeps its source wording naming IPv4). Also took the non-blocking suggestion from the review body: added a behavior-only entry to versioning.md — under the current 9.3 block rather than 8.8, since the branch base had advanced past 8.8 by the time this landed.

Issue #4254, opened 2026-09-03
intentd: add completion-persistent agent watches


Wattenberger:
## Problem

ws.agent.watch is one-shot even when its wake is only an advisory that the child is monitoring-idle. Coordinators that need the actual terminal completion must wake, inspect watchStillArmed, and manually re-arm.

In the Remote app development DX workspace, one verifier with background monitoring required six re-arm turns before its genuine completion. Those turns added coordinator latency and created repeated opportunities to miss the final wake.

Workspace evidence: intent://local/starting-run

## Minimal API shape

Allow an explicit persistent terminal mode:

- ws.agent.watch(agentId, { until: "completion" })
- Monitoring-idle and attention advisories may still wake the caller, but do not consume this watch.
- Consume it only when the target completes, fails, or is deleted.
- Every wake continues to expose watchStillArmed so callers can verify the contract.

The existing one-shot behavior remains the default for compatibility.

## Acceptance criteria

- A monitoring-idle advisory does not retire a watch created with until: completion.
- The same watch later delivers one genuine terminal wake and then retires.
- Failure and deletion are terminal.
- Tests cover ordinary one-shot watches and completion-persistent watches.

## Related, not duplicate

- #2051 clarified that ordinary watches must be manually re-armed; it did not add a persistent option.
- #4248 covers retry behavior when an advisory targets a retired parent.

panghy:
Design decision (workspace: Completion-persistent agent watches): we are NOT adding an 'until: completion' parameter. Instead the default contract changes so it is uniform across all watch shapes:

- A watch (ungrouped or grouped) is consumed ONLY by terminal settlement: completion, failure, deletion, or retirement.
- The monitoring-idle advisory never retires a watch. It delivers watchStillArmed: true, and each NEW continuous waiting period may advise again (still at most one advisory per period via the persisted advisory_wake_delivery marker).
- Attention wakes and reportToParent wakes already never consume; the ungrouped advisory was the one remaining non-terminal wake that did. Grouped (after_all) watches already behave the proposed way, so this aligns the ungrouped shape with them.
- completion_only (ask-only) watches will skip advisories entirely, matching their documented strictly-terminal contract.

[...]

PR #4263, opened 2026-09-03
docs(protocol): monitoring-idle advisories never consume completion watches

panghy:
Documents the behavioral change from intent-hq/intent#4254 (implemented on intentd branch `feat/persistent-agent-watches`): **monitoring-idle advisory wakes never consume completion watches**. Only terminal settlement (completion / failure / deletion / retirement) retires a watch.

## Changes

- **`docs/protocol/methods/agent-aux.md`**
  - §Completion-watch persistence, monitoring-idle advisory wake: the ungrouped advisory no longer retires the watch. Both shapes stay armed (`watchStillArmed: true`), the advisory message names `ws.agent.unwatch` as the opt-out instead of instructing a re-arm, and the stable message identity is now `advisory-wake:<watchId>:<eventId>` (the idle event id is a per-period discriminator, since the armed watch keeps its id across waiting periods).
  - Ordering & failure semantics: the retirement transaction is gone — the shared once-per-period marker is recorded best-effort after the wake is durable; one duplicate advisory is the accepted worst case for a failure between the two. A failed ungrouped delivery still leaves the watch armed and schedules the per-child retry.
  - Ask-registered completion-only watches (the chief `ws.app.agents.ask` shape) skip advisories entirely — they wait strictly for terminal settlement and neither consult nor write the period marker.
  - `watchStillArmed` paragraph: the advisory moves from the `false` list to the `true` list (both shapes).
- **`docs/protocol/versioning.md`**: supersession note on the v8.6 entry plus a new "Also within 9.3" entry (behavioral change on additive wake metadata; no method-catalog or wire-shape change).

The matching intentd MCP help-text update rides the `feat/persistent-agent-watches` branch (commit bc668a8f).

Refs intent-hq/intent#4254.


panghy (commented):
## Docs review — verified against `intentd` `feat/persistent-agent-watches`

**Verdict: no blocking findings.** The diff accurately describes the implementation; internal consistency between `agent-aux.md` and `versioning.md` holds; no live (non-changelog) text in `docs/protocol` still claims advisories consume watches.

### Verified claims (doc → implementation)

| Doc claim | Evidence |
|---|---|
| Advisories never consume a watch; `watchStillArmed: true` on grouped and ungrouped shapes | `intent-services/src/lib.rs` `deliver_advisory_wake_for_watch` — no watch removal on the advisory path; e2e `monitoring_idle_advisories_leave_watch_armed_until_genuine_completion_over_wss` asserts `"watchStillArmed":true`, `stays armed` / `no re-arm needed` text, and absence of `consumed your one-shot watch` |
| Stable identity `advisory-wake:<watchId>:<eventId>` | `lib.rs:6751` `format!("advisory-wake:{}:{}", watch.id, event.id)`; doc comment at 6667–6668 explains why the bare `advisory-wake:<watchId>` form would dedup-suppress every later period |
| Once per continuous waiting period; new period advises again | `has_advisory_wake_delivery(parent, child)` marker gate in the same function; e2e covers same-period suppression and the fresh-period re-advise |
| Ask-registered completion-only watches never receive an advisory and neither consult nor write the marker | `lib.rs:6698` `if watch.completion_only { … return false }` placed *before* the marker read; `agent_subscriptions.rs:281` `completion_only: true` in `register_completion_watch_strict_durable`, called from the chief `ws.app.agents.ask` handler (`agent_ops.rs:9957`) |
| Only terminal settlement retires; `watchStillArmed: false` on idle/failed/deleted | e2e asserts `"watchStillArmed":false` on idle (l.962), failure (l.1210) wakes; blocker wake asserts `true` (l.1019) |
| `ws.agent.unwatch` is the opt-out | e2e `agent_unwatch_stops_further_wakes_over_wss` (WATCH-2c): `removed=true`, no wake after unwatch |
| Version 9.3, "no method-catalog or wire-shape change" | `intent-transport/src/protocol.rs:345` `PROTOCOL_VERSION = "9.3"` — unchanged on the feat branch, consistent with a behavior-only entry |

### Stale-text scan

Searched `docs/protocol/**` on the PR branch for `advisor` / `consum` / `re-arm`. The only remaining "ungrouped watches are CONSUMED by the advisory" wording is the v8.6 changelog paragraph in `versioning.md`, which already carries the "(later superseded within 9.3 — see below)" note. `agents.md` (reportToParent progress wake, `watchStillArmed: true`) and `06-events.md` are consistent with the new contract. Nothing else to update.

### Non-blocking suggestions

1. **Ask-reuse nuance.** `register_completion_watch_strict_durable` reuses an *existing* ungrouped watch between the same parent/child (`agent_subscriptions.rs:256–269`) without flipping it to `completion_only`. So "Ask-registered completion-only watches … never receive an advisory" is exact for watches the ask path creates; if the chief already held a `ws.agent.watch` on the target, that pre-existing (advisory-eligible) watch is what `ask` returns and it can still advise. Consider a short parenthetical, e.g. "(an `ask` that finds an existing ungrouped watch reuses it as-is, so a prior `ws.agent.watch` on the same target keeps its advisory eligibility)". Optional — arguably an implementation detail.
2. **Marker read fails closed.** The implementation skips the advisory (watch stays armed) when the once-per-period marker read errors (`lib.rs` comment above the `has_advisory_wake_delivery` call). If the doc wants to be complete about failure modes it could mention this in the "Once per waiting period" sentence; not required for accuracy.

Not merging per policy; leaving that to a human.

panghy:
Addressed both non-blocking suggestions in f89eb8c (`docs/protocol/methods/agent-aux.md`):

1. **Ask-reuse nuance** — added a parenthetical after the "Ask-registered completion-only watches … never receive an advisory" sentence: it holds for watches the ask path creates; an `ask` that finds an existing ungrouped watch for the same parent/child pair reuses it as-is, so a prior `ws.agent.watch` on the same target keeps its advisory eligibility. Verified against `register_completion_watch_strict_durable` (`agent_subscriptions.rs:256–269` reuses the existing watch without setting `completion_only`).
2. **Marker read fails closed** — added a parenthetical to the "Once per waiting period" sentence (read error → advisory skipped, watch stays armed), cross-referencing the existing statement under "Ordering & failure semantics". Verified against the `has_advisory_wake_delivery` call in `lib.rs` (`unwrap_or_else(… true)`).

No inline review threads were opened, so there is nothing to resolve.

panghy:
Follow-up from the intent-hq/intentd#1686 review: 39032c1 adds the monitoring-idle advisory wake to the list of notifications in docs/protocol/methods/agents.md that can carry watchStillArmed: true (it leaves the watch armed on both the ungrouped and grouped shapes, consistent with the advisory section in agent-aux.md).

Issue #4082, opened 2026-09-01
Merge held legacy-import removal PRs once feature is confirmed removable


panghy:
Tracking issue for the held legacy-workspace-import removal. The removal was reverted (restored on main) pending confirmation that the feature is actually removable; the removal is re-filed as DRAFT PRs to be merged only once that confirmation lands.

## Held draft PRs (do NOT mark ready or merge until confirmed)

- cloudlands-fe: [intent-hq/cloudlands-fe#2001](https://github.com/intent-hq/cloudlands-fe/pull/2001) — re-applies [#1980](https://github.com/intent-hq/cloudlands-fe/pull/1980)
- intentd: [intent-hq/intentd#1630](https://github.com/intent-hq/intentd/pull/1630) — re-applies [#1616](https://github.com/intent-hq/intentd/pull/1616) (squash 615f5dc3)
- monorepo docs: [intent-hq/intent#4081](https://github.com/intent-hq/intent/pull/4081) — re-applies [#4035](https://github.com/intent-hq/intent/pull/4035)

## Merge order (removal direction)

1. cloudlands-fe#2001 (drop the FE callers first)
2. intentd#1630 (drop the wire method — becomes protocol **10.0** at that point)
3. intent#4081 (docs last)

Note this is the removal direction, the reverse of the usual intentd-first rule for additive protocol changes: the FE callers must be gone before the daemon method is removed.

## Required refresh before merging

The intentd and docs drafts are pure reverts-of-the-reverts, so their protocol/versioning content is stale: they currently reset the protocol to 9.0 and drop the 9.1 restoration records. Since 9.1 has been released, the eventual removal is a new breaking change — before merge, refresh intentd's PROTOCOL_VERSION / version-history / golden catalog counts and the docs versioning entries to record the removal as **protocol 10.0**, keeping the 9.0 removal + 9.1 restoration history. Rebase all three drafts onto then-current main.

## Shipped-version history

- Removal shipped in cloudlands-fe alpha **v2.123.0** / intentd **0.9.0** (protocol 9.0): [cloudlands-fe#1980](https://github.com/intent-hq/cloudlands-fe/pull/1980), [intentd#1616](https://github.com/intent-hq/intentd/pull/1616), docs [#4035](https://github.com/intent-hq/intent/pull/4035).
[...]

PR #4363, opened 2026-09-04
feat(triage): set Priority and Effort issue fields instead of priority labels

panghy:
## Summary

The issue-triage pipeline stops using the priority:P0-P3 labels and instead sets the org-level GitHub issue fields **Priority** (Urgent / High / Medium / Low) and **Effort** (High / Medium / Low).

- **Pass 1 (deterministic)** - parse-issue.js exports priorityForIssueBody (accepts the new Urgent/High/Medium/Low dropdown options and the legacy P0-P3 tokens: P0->Urgent, P1->High, P2->Medium, P3->Low). The workflow fills the Priority field via the setIssueFieldValue GraphQL mutation in both the triage step and the needs-info-reply re-run. Field and option ids are resolved at runtime from repository.issueFields; the write is fill-only (never overwrites an existing value) and fail-soft (core.warning, job stays green).
- **Pass 2 (agentic)** - agentic-triage.js gains fetchIssueContext / planIssueFields / setIssueFields. The prompt asks for priority and effort (with an Effort rubric), planActions plans field writes gated on the current field values instead of labels, and the summary comment reports Priority set / Effort set with reasons. Suspected security issues still escalate to Urgent. Suggest-don't-destroy is unchanged: human-set values always win.
- **Templates and docs** - bug_report.yml Severity options renamed to Urgent/High/Medium/Low (same descriptions); docs/README.md severity taxonomy updated.

## Follow-up on this branch
A --fields-only backfill mode and the one-off migration of open issues (labels -> fields, Effort estimated, priority label definitions deleted) is in progress and will be pushed to this branch.

## Verification
- node --test over the three triage test files: 70 pass, 0 fail
- Dry-runs: node .github/scripts/issue-triage/agentic-triage.js --dry-run 3696 / 4082 plan field writes and no priority label
- Read queries and mutation input shape validated against the live GraphQL schema
- Not yet exercised: a real setIssueFieldValue write with the Actions github.token on the ORG_ONLY fields; the fail-soft path covers a rejection until the first live run confirms it.

### DIFF
--- a/.github/scripts/issue-triage/agentic-triage.js
+++ b/.github/scripts/issue-triage/agentic-triage.js
@@ -53,6 +53,18 @@ const PRIORITIES = ['P0', 'P1', 'P2', 'P3'];
 // resolved at runtime from `repository.issueTypes`, never hardcoded.
 const ISSUE_TYPE_BY_LABEL = { bug: 'Bug', enhancement: 'Feature', question: 'Task' };
 
+// Org-level single-select issue fields (the sidebar "Fields" section, not a
+// Projects v2 board). Field and option IDs are resolved at runtime from
+// `repository.issueFields` by name, never hardcoded. A P0–P3 priority from
+// the model maps onto the Priority field vocabulary; the field option
+// names themselves are also accepted so the response schema can switch
+// vocabularies without touching the planner.
+const PRIORITY_FIELD = 'Priority';
+const EFFORT_FIELD = 'Effort';
+const PRIORITY_OPTION_BY_LABEL = { P0: 'Urgent', P1: 'High', P2: 'Medium', P3: 'Low' };
+const PRIORITY_OPTIONS = ['Urgent', 'High', 'Medium', 'Low'];
+const EFFORT_OPTIONS = ['Low', 'Medium', 'High'];
+
 const MAX_ISSUE_BODY_CHARS = 4000;
 const MAX_CANDIDATE_BODY_CHARS = 400;
 const MAX_CANDIDATES = 12;
@@ -353,6 +365,66 @@ function planIssueType({ response, currentLabels, currentIssueType, issueTypes }
   };
 }
 
+// Resolve a single-select option by field name + option name against the
+// runtime `repository.issueFields` list. Null when the field or option is
+// unknown (fields disabled, lookup failed, option renamed) — never a guess.
+function resolveIssueFieldOption(issueFields, fieldName, optionName) {
+  if (!optionName) return null;
+  const field = (issueFields || []).find(
+    (f) => f && f.name === fieldName && f.id && Array.isArray(f.options)
+  );
+  if (!field) return null;
+  const option = field.options.find((o) => o && o.name === optionName && o.id);
+  return option ? { fieldId: field.id, optionId: option.id, name: option.name } : null;
+}
+
+// Decide the Priority / Effort field values to set. Pure, so the gating is
+// unit-testable; the result carries only the fields to write:
+//   - an existing field value always wins (never overwritten,
+//     suggest-don't-destroy) — `currentFieldValues` is field name → option
+//     name as read from `issue.issueFieldValues`,
+//   - Priority comes from the model's pick (P0–P3 mapped, or a field option
+//     name); a suspected security issue escalates to Urgent regardless,
+//   - Effort comes from `response.effort` (Low / Medium / High) when present,
+//   - both must resolve to a runtime field option by name — an empty list
+//     (lookup failed, fields unavailable) plans nothing.
+function planIssueFields({ response, currentFieldValues, issueFields }) {
+  const current = currentFieldValues || {};
+  const plan = {};
+  const r = response || {};
+  const reasons = r.reasons || {};
+
+  let priorityName = null;
+  if (r.security) {
+    priorityName = 'Urgent';
+  } else if (PRIORITY_OPTION_BY_LABEL[r.priority]) {
+    priorityName = PRIORITY_OPTION_BY_LABEL[r.priority];
+  } else if (PRIORITY_OPTIONS.includes(r.priority)) {
+    priorityName = r.priority;
+  }
+  if (!current[PRIORITY_FIELD]) {
+    const priority = resolveIssueFieldOption(issueFields, PRIORITY_FIELD, priorityName);
+    if (priority) {
+      plan.priority = {
+        ...priority,
+        reason: r.security
+          ? 'suspected security impact escalates to Urgent'
+          : reasons.priority || 'inferred from the issue text',
+      };
+    }
+  }
+
+  const effortName = EFFORT_OPTIONS.includes(r.effort) ? r.effort : null;
+  if (!current[EFFORT_FIELD]) {
+    const effort = resolveIssueFieldOption(issueFields, EFFORT_FIELD, effortName);
+    if (effort) {
+      plan.effort = { ...effort, reason: reasons.effort || 'inferred from the issue text' };
+    }
+  }
+
+  return plan;
+}
+
 // The one auditable, marker-idempotent summary comment.
 function buildSummaryComment(plan) {
   const lines = ['### Automated triage', ''];
@@ -411,6 +483,7 @@ module.exports = {
   extractSearchQueries,
   parseTriageResponse,
   planActions,
+  planIssueFields,
   planIssueType,
   sanitizeText,
 };
@@ -439,13 +512,25 @@ function ghJson(args) {
   return JSON.parse(gh(args));
 }
 
-// One repository-scoped GraphQL read for everything the Type decision
-// needs: the enabled Types (IDs resolved at runtime, never hardcoded), the
-// issue's current Type, and its node id for the mutation. Repository scope
-// (not `organization.issueTypes`) is what the Actions token's
-// `issues: write` can read. Fail-soft: on any error the caller gets an
-// empty Type list, so planIssueType sets nothing and triage continues.
-function fetchIssueTypeContext(repo, issueNumber) {
+// Field name → option name for the single-select values on an issue, from
+// an `issueFieldValues.nodes` list (non-single-select nodes are skipped).
+function singleSelectValuesByField(nodes) {
+  const values = {};
+  for (const v of nodes || []) {
+    if (v && v.field && v.field.name && v.value) values[v.field.name] = v.value;
+  }
+  return values;
+}
+
+// One repository-scoped GraphQL read for everything the Type and field
+// decisions need: the enabled Types and the single-select issue fields
+// (IDs resolved at runtime, never hardcoded), the issue's current Type and
+// field values, and its node id for the mutations. Repository scope (not
+// `organization.issueTypes` / `organization.issueFields`) is what the
+// Actions token's `issues: write` can read. Fail-soft: on any error the
+// caller gets empty Type and field lists, so planIssueType /
+// planIssueFields set nothing and triage continues.
+function fetchIssueContext(repo, issueNumber) {
   const [owner, name] = repo.split('/');
   try {
     const data = ghJson([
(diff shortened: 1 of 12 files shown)


.github/workflows/issue-triage.yml
```diff
@@ -120,31 +132,74 @@ jobs:
...
+              }
+              const option = (field.options || []).find((o) => o.name === priority);
+              if (!option) throw new Error(`Priority field has no "${priority}" option`);
+              if (dryRun) {
+                core.notice(`[dry-run] Would set Priority = ${priority} on #${issueNumber}`);
+                return;
+              }
+              await github.graphql(
+                `mutation($issueId: ID!, $fieldId: ID!, $optionId: ID!) {
+                  setIssueFieldValue(input: { issueId: $issueId, issueFields: [{ fieldId: $fieldId, singleSelectOptionId: $optionId }] }) {
```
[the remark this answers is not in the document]
panghy: Acknowledged; this window is narrowed as far as the API allows rather than eliminated. IssueFieldCreateOrUpdateInput has no conditional / only-if-empty precondition, so a compare-and-set is not expressible. Mitigations in place: (1) in issue-triage.yml both steps read the current Priority value and issue the mutation within the same script block, with no model call or other long operation in between, so the window is a single request round-trip; (2) in agentic-triage.js the long LLM call is bracketed by a pre-write re-read in setIssueFields (the re-read drops any field a human filled while the model ran), so the residual window is again one round-trip. Given the fields are org-only suggestions and the pass runs once per issue, we accept that residual race. Resolving.

.github/scripts/issue-triage/agentic-triage.js
```diff
@@ -42,17 +45,29 @@ const POSSIBLE_DUPLICATE_LABEL = 'possible-duplicate';
...
 // Type label → GitHub issue Type name. The repo enables exactly Task, Bug,
 // and Feature (no "Question" Type), so questions map to Task. Type IDs are
 // resolved at runtime from `repository.issueTypes`, never hardcoded.
 const ISSUE_TYPE_BY_LABEL = { bug: 'Bug', enhancement: 'Feature', question: 'Task' };
 
+// Org-level single-select issue fields (the sidebar "Fields" section, not a
+// Projects v2 board). Field and option IDs are resolved at runtime from
+// `repository.issueFields` by name, never hardcoded. The model emits the
+// Priority field vocabulary directly; the planner also maps the legacy
+// P0–P3 tokens so either response vocabulary plans the same field write.
```
panghy: This is not reachable from the real pipeline: `parseTriageResponse` (line 264) clamps `priority` to `PRIORITY_OPTIONS`, so a legacy `"P1"` from the model becomes `null` before `planIssueFields` ever sees it (the test at agentic-triage.test.js:87 asserts exactly that). The map is only exercised by tests that set `response.priority` directly. Either map the legacy tokens in the parser too (more robust if the model regresses to the old vocabulary) or reword this comment so it does not promise it.

.github/scripts/issue-triage/agentic-triage.js
```diff
@@ -171,17 +187,22 @@ function buildPrompt(issue, candidates) {
     '  surface, same steps).',
     '- component/type: infer from stack traces, file paths, and vocabulary;',
     '  use null when genuinely unsure.',
-    '- priority rubric: P0 = crash, data loss/corruption, or a suspected',
-    '  security vulnerability; P1 = major functionality broken with no',
-    '  workaround; P2 = default for ordinary bugs; P3 = cosmetic/papercut.',
+    '- priority rubric: Urgent = crash, data loss/corruption, or a suspected',
+    '  security vulnerability; High = major functionality broken with no',
+    '  workaround; Medium = default for ordinary bugs; Low = cosmetic/papercut.',
     '  Feature requests and questions usually take no priority (null).',
```
panghy: Not a bug, flagging for the spec: this rule means Tasks/Features get no Priority from pass 2 (confirmed by a dry-run on #4082: `Priority: none (unchanged)`), while the spec's backfill acceptance criterion says Priority + Effort should be populated on all open issues. If the backfill on this branch reuses this prompt, that criterion will not be met for feature requests; decide which one wins.

.github/scripts/issue-triage/parse-issue.js
```diff
@@ -96,17 +112,24 @@ function labelsForIssueBody(body) {
...
-module.exports = { labelsForIssueBody, splitSections };
+// Derive the Priority field value the Severity dropdown maps onto:
+// 'Urgent' | 'High' | 'Medium' | 'Low', or null when the section is
+// missing, unanswered ("_No response_"), or free-form text.
+function priorityForIssueBody(body) {
+  if (!body || typeof body !== 'string') return null;
+  const sections = splitSections(body);
+  if (!sections['Severity']) return null;
+  const selection = firstContentLine(sections['Severity']);
+  const m = selection.match(/^([A-Za-z][A-Za-z0-9]*)\b/);
```
panghy, line 130: Case-insensitive matching works (`urgent`, `URGENT`, `p0` all resolve) but no test pins it down; one assertion in the legacy/current-vocabulary tests would guard the `toLowerCase()`. Minor: free text like `High-priority issue` maps to `High` because `\b` stops at the hyphen - same class of behaviour as the old `^P([0-3])\b`, and the dropdown is required, so no action needed.

.github/workflows/issue-triage.yml
```diff
@@ -120,31 +132,74 @@ jobs:
...
+              await github.rest.issues.addLabels({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: issueNumber,
+                labels: toAdd,
+              });
+              core.info(`Added labels: ${toAdd.join(', ')}`);
+            }
+
+            // Priority field: fill only when empty; ids resolved at runtime;
```
panghy, line 166: Schema, fill-only gate and fail-soft all check out. Non-blocking: this block is byte-identical to lines 384-420 in `needs-info-reply`. Extracting it into `.github/scripts/issue-triage/` (a function taking `github`, `core`, `{ owner, repo, issueNumber, priority, dryRun }`) would keep both steps in sync and let the gate be unit-tested with a fake `graphql`, like `parse-issue.js` already is.

panghy (commented):
Review of 673229e (verifier pass; no files modified, no mutations run).

**Verified (evidence-based)**
- GraphQL shapes checked by introspection: `SetIssueFieldValueInput { issueId: ID!, issueFields: [IssueFieldCreateOrUpdateInput!]! }`, `IssueFieldCreateOrUpdateInput { fieldId: ID!, singleSelectOptionId: ID, ... }`, payload exposes `clientMutationId` / `issue` / `issueFieldValues`; `IssueFieldSingleSelectValue { value: String!, optionId, field: IssueFields }`. Both the inline-variable mutation in the two workflow steps and the `$fields: [IssueFieldCreateOrUpdateInput!]!` form in `setIssueFields` are schema-valid.
- The `fetchIssueContext` read query and the `gh api graphql --input -` (JSON query+variables on stdin) pattern both execute correctly against `intent-hq/intent` (read-only). Note the live `repository.issueFields.nodes` list contains empty `{}` nodes for non-single-select fields; the script's filter and the workflow's `f && f.name === 'Priority'` lookup both handle that.
- Fill-only + fail-soft: both workflow steps check `existing.value` before writing and wrap lookup+write in try/catch -> `core.warning`; `setIssueFields` re-reads immediately before the mutation, drops any field filled meanwhile, and on failure returns `false` so `main()` removes the fields from the plan before the summary comment is built (comment never claims a write that did not happen).
- Argv/token safety: every `gh` call is `execFileSync` argv; the mutation body goes via stdin; log/warn strings contain field/option ids and names only; `runAuggie` still scrubs `GH_TOKEN`/`GITHUB_TOKEN` and denies all tools.
- Prompt-injection surface: unchanged pattern (untrusted-data rule, output clamped to `PRIORITY_OPTIONS`/`EFFORT_OPTIONS`, `reasons.effort` sanitized; the invented-response fixture now exercises a marker + @-mention in the effort reason).
- `node --test .github/scripts/issue-triage/*.test.js` -> 70/70 pass (Node 24). `node agentic-triage.js --dry-run 4082` (PR-head copy) -> fields resolved at runtime, `Priority: none (unchanged)` for a Task, `Effort: none -> set High`, no `priority:*` label anywhere in the plan.

**Findings** (all low severity / non-blocking; details inline)
1. `agentic-triage.js:62-67, 402-403` - the P0-P3 -> option map in `planIssueFields` is unreachable through `parseTriageResponse`, which clamps `priority` to the field vocabulary (test at `agentic-triage.test.js:87` asserts `P0 -> null`). The header comment claims either vocabulary plans the same write; today a model that falls back to `"P1"` produces no Priority. Either accept legacy tokens in the parser or correct the comment.
2. `parse-issue.js:130` - lowercase/uppercase tokens (`urgent`, `URGENT`, `p0`) are accepted via `toLowerCase()` but there is no test for it; suggest one assertion.
3. `issue-triage.yml:160-201` and `:384-420` - the Priority block is duplicated verbatim across two `github-script` steps; a small helper in `.github/scripts/issue-triage/` taking `(github, core, opts)` would keep the two in sync and make the fill-only gate unit-testable with a fake `graphql`.
4. `issue-triage.yml:12-16` - the add-only tradeoffs list should mention the analogous field case: a Priority cleared by hand is re-filled from the body on the next `edited` event / author reply.
5. Stale wording not covered by the diff: `issue-triage.yml:67-68` ("print the intended label operations") and `:431-437` (the agentic-job comment describes only the Type write, not `setIssueFieldValue`); `CONTRIBUTING.md:93-94` still describes the severity taxonomy as "P0 crash/data-loss, P1 broken feature, P2 papercut".
6. Spec tension (not a code bug): the prompt (`agentic-triage.js:193`) tells the model that feature requests/questions usually take `null` priority, while the spec's acceptance criterion for the backfill says Priority should be populated on all open issues. Worth deciding explicitly before the backfill commit lands on this branch.

.github/workflows/issue-triage.yml
```diff
@@ -1,17 +1,26 @@
...
+# severity dropdown fills the issue's Priority FIELD (Urgent / High /
+# Medium / Low — formerly the P0–P3 priority labels; the parser still
+# accepts the old P0–P3 option tokens), and `needs-triage` marks the
+# untriaged queue on open. The parser only ever ADDS labels — it never
+# removes any — so re-runs are idempotent and this job never fights pass 2
+# or human re-classification.
 #
 # Add-only tradeoffs (accepted for pass 1):
 #   - A label the CURRENT body still derives is re-added on the next
 #     `edited` event, so removing such a label by hand does not stick; to
```
panghy, line 14: Suggest adding the field analogue here: because the fill is body-derived and runs on `edited` and on author replies, a Priority a human has *cleared* is re-filled from the Severity dropdown on the next event (an existing value is respected, an empty one is not). Same tradeoff as the labels, worth stating.

.github/scripts/issue-triage/agentic-triage.test.js
```diff
@@ -224,6 +329,131 @@ test('type: nothing is set without a signal, a matching enabled Type, or a resol
...
+    },
+    effort: {
+      fieldId: 'IFSS_effort',
+      optionId: 'IFSSO_e_medium',
+      name: 'Medium',
+      reason: 'Touches one crate and needs a regression test',
+    },
+  });
+
+  // The P0–P3 map covers every label; the field vocabulary is accepted as-is.
```
panghy, line 362: These assertions bypass `parseTriageResponse`, which is what makes the P0-P3 branch look covered while it is unreachable end to end (see the comment on agentic-triage.js:64). If the parser is changed to accept legacy tokens, a parser-level assertion (`parseTriageResponse('{"priority":"P1"}').priority`) would be the meaningful one.

.github/scripts/issue-triage/agentic-triage.js
panghy: Resolved by design: the backfill (--fields-only mode, landing on this branch) uses a variant of the prompt that asks for a Priority on every issue, so all open issues end with both fields set; the regular pass-2 rule (features/questions usually null) is unchanged from the previous label-based behavior and is intentional.

.github/workflows/issue-triage.yml
panghy: Agreed the block is duplicated; extracting a unit-testable helper module is a reasonable follow-up but out of scope for this PR (it would also change the needs-info path's checkout/require layout). Leaving as-is here.

.github/scripts/issue-triage/parse-issue.js
panghy: fixed in 8379de5

.github/workflows/issue-triage.yml
panghy: fixed in 8379de5

panghy:
Review finding 5 (stale wording: issue-triage.yml dry-run comment and agentic-job comment, CONTRIBUTING.md severity taxonomy) fixed in 8379de5. Repo grep outside packages/ found no other stale P0–P3 / priority-label references (remaining P0–P3 hits are intentional legacy-token coverage).

.github/scripts/issue-triage/agentic-triage.js
panghy: Fixed in 782ae23 — parseTriageResponse maps legacy P0–P3 tokens (case-insensitive) onto the field vocabulary before the clamp; parser-level assertions added for P0/P1/p2/P3 → Urgent/High/Medium/Low.

.github/scripts/issue-triage/agentic-triage.test.js
panghy: Fixed in 782ae23 — parseTriageResponse maps legacy P0–P3 tokens (case-insensitive) onto the field vocabulary before the clamp; parser-level assertions added for P0/P1/p2/P3 → Urgent/High/Medium/Low.

Issue #4367, opened 2026-09-04
make test aborts on nextest retry ids (#2) from flaky e2e_wss_agent_watch tests


panghy:
## Summary

A flaky test that PASSES on nextest retry turns into a hard 'make test' failure: the resumable gate wrapper does not recognise nextest's retry suffix on the test identifier and aborts the whole run. Two pre-existing flaky tests in 'intentd::e2e_wss_agent_watch' trigger it regularly.

## Observed (local 'make test' from the monorepo root, intentd at 838087af)

Run 1:

    Summary [ 110.207s] 4816/8265 tests run: 4816 passed (1 flaky, 1 leaky), 4 skipped
      FLAKY 2/2 [  33.028s] (4816/8265) intentd::e2e_wss_agent_watch report_progress_keeps_original_watch_for_terminal_completion_over_wss
    error: test run failed
    [test-intentd] ERROR: nextest test identifier was not listed: 'intentd::e2e_wss_agent_watch$report_progress_keeps_original_watch_for_terminal_completion_over_wss#2'
    make: *** [Makefile:287: test-intentd] Error 2

Run 2 ('make test RESUME=1'):

    Summary [  96.577s] 7233/8265 tests run: 7233 passed (1 flaky, 1 leaky), 4 skipped
      FLAKY 2/2 [   3.925s] (7233/8265) intentd::e2e_wss_agent_watch agent_watch_rearm_on_idle_but_waiting_target_defers_over_wss
    error: test run failed
    [test-intentd] ERROR: nextest test identifier was not listed: 'intentd::e2e_wss_agent_watch$agent_watch_rearm_on_idle_but_waiting_target_defers_over_wss#2'

A third 'RESUME=1' segment completed the suite: 8265/8265 passed, 0 failures. So every test is green, but a clean 'make test' needs up to three invocations.

### 1. Gate wrapper: retry suffix not handled (monorepo 'scripts/resumable_nextest.py')

[...]

PR #4371, opened 2026-09-04
fix(scripts): accept nextest retry suffix in resumable gate pass events

panghy:
## Summary

nextest names retried attempts `<suite>$<test>#<attempt>` in the libtest-json event stream. `scripts/resumable_nextest.py` looked the raw name up in the map built from `nextest list`, so a FLAKY-but-passed test (e.g. `intentd::e2e_wss_agent_watch$report_progress_keeps_original_watch_for_terminal_completion_over_wss#2`) aborted the whole gate with `nextest test identifier was not listed`.

`parse_passed_event` now strips a trailing `#<digits>` fragment from the test part before the lookup and records the **base** test name as passed, so a later `RESUME=1` filters it correctly. A name that genuinely is not in the list output still raises — there is no blanket swallowing of unknown identifiers, and the strip only applies to a trailing `#<digits>` fragment.

Refs intent-hq/intent#4367 (defect 1; the intentd test flakiness itself is tracked separately).

## Changes

- `scripts/resumable_nextest.py`: add `RETRY_SUFFIX_RE` and strip it in `parse_passed_event`.
- `scripts/test_resumable_nextest.py`: regression test `test_pass_event_strips_retry_suffix` — a `...$module::passes#2` event resolves to the same `(binary_id, "module::passes")` as the un-suffixed event; an unlisted `...$module::missing#2` still raises; a non-numeric trailer (`#2x`) is not stripped.

## Verification

- `python3 -m py_compile scripts/resumable_nextest.py` — OK
- `python3 -m unittest scripts/test_resumable_nextest.py -v` — 9 tests, OK

### DIFF
--- a/scripts/resumable_nextest.py
+++ b/scripts/resumable_nextest.py
@@ -19,6 +19,7 @@ SCHEMA_VERSION = 1
 MAX_AGE_SECONDS = 7 * 24 * 60 * 60
 KEY_RE = re.compile(r"^[0-9a-f]{64}$")
 RUST_FLAG_ENV = {"RUSTFLAGS", "CARGO_ENCODED_RUSTFLAGS", "CARGO_BUILD_RUSTFLAGS"}
+RETRY_SUFFIX_RE = re.compile(r"#[0-9]+$")
 
 
 def run(command: list[str], cwd: Path, env: dict[str, str] | None = None) -> str:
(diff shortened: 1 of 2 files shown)


panghy:
Review (agent): no issues found. Checked against nextest-runner 0.122.1 libtest.rs: on TestFinished nextest appends '#<run_statuses.len()>' to the test name when retried and emits event ok only for Success / Flaky-Pass (flaky-treated-as-failure emits failed), so stripping a trailing '#<digits>' before the (suite, test) lookup is correct. The lookup still requires the stripped key to exist in the nextest-list map and Rust test identifiers cannot contain '#', so no genuinely unlisted name can be accepted; the recorded name is the stripped one, consistent with the resume-time known_tests check and the exact-match default-filter. scripts/test_resumable_nextest.py passes (9 tests) on the branch tip.

Issue #4409, opened 2026-09-04
fix(intentd): preserve assistant metadata in chat terminal reconciliation


Wattenberger:
Current protocol §7.1 guarantees that a chat subscription snapshot reduced with terminal deltas equals a fresh agent.getConversation result. The current terminal mapper in packages/intentd/crates/intent-transport/src/subscriptions.rs reads the persisted assistant row but emits entities built only from block, sequence, timestamp, and streaming completion; it does not lift the row metadata. Reproduction against disposable intentd 0.9.12 shows agent.getConversation returning interrupted metadata including interrupted, stopReason, status, and interruptReason while the terminal chat delta omits it. Static review confirms the current source retains the omission. This prevents clients using only chat.subscribe from rendering the canonical interrupted state.

panghy:
This fix is included in intentd v0.9.25.

panghy:
This fix is included in cloudlands-fe v2.137.0 (bundles intentd v0.9.25).

PR #4416, opened 2026-09-05
docs(protocol): chat.subscribe terminal deltas carry assistant metadata

panghy:
Documents the terminal-delta metadata contract for intent-hq/intent#4409.

## What changed

- `docs/protocol/07-agent-streaming.md` §7.1 — new "Terminal entities carry the row's `metadata`" paragraph under the terminal-reconcile invariant: every entity in the `agent:stream:end` terminal frame carries the re-read assistant row's `metadata` verbatim (present only when the persisted row has a non-null map, omitted otherwise, never `null` — the same presence-detected rule as the `agent:message` non-assistant row deltas). States that the degraded best-effort terminal frame (no persisted row) carries no `metadata`, and that live chunk deltas never do. The example prose after the delta frame gains an interrupted-turn terminal entity showing the lifted `metadata`, and the §7.2 `chat.subscribe` consequence paragraph now points at the terminal entities as the carrier.
- `docs/protocol/versioning.md` — "Also within 9.7" history note (additive entity field, presence-detected, no version bump) referencing intent-hq/intent#4409.

## Ordering

Protocol docs describe daemon behavior: **merge only after the intentd PR implementing the lift has merged** (intentd-first rule for protocol changes). Implementing PR: intent-hq/intentd#1746 (head 9c24dde). Zero-block rows (no entity, no carrier) are called out in the docs and tracked in intent-hq/intent#4417.

Refs intent-hq/intent#4409


docs/protocol/07-agent-streaming.md
```diff
@@ -585,6 +585,20 @@ live, carrying the authoritative `messageSeq`/`timestamp`/`streamingComplete:tru
...
 
+**Terminal entities carry the row's `metadata`** ([intent-hq/intent#4409](https://github.com/intent-hq/intent/issues/4409)).
+When the re-read assistant row carries `metadata` (the §7.2 interrupt tag set —
+`interrupted` / `stopReason` / `interruptReason` / `interruptedBy` — or the §7.3 `finishReason`),
+every entity in the terminal frame additionally carries it **verbatim** as `metadata`, under the
+same rule as the non-assistant row deltas above: present only when the persisted row has a
+non-null metadata map, omitted entirely otherwise, **never** `null` (additive, presence-detected —
+rows without metadata keep the lean entity shape). So a `chat.subscribe`-only client renders the
+interrupted / "Stopped" / abnormal-finish state at the terminal frame exactly as an
+`agent.getConversation` reader would, with no refetch. The **degraded best-effort terminal frame**
```
panghy: **Accuracy gap — zero-block rows.** Terminal entities are per persisted content block (`reconcile` iterates `contentBlocks` and builds one entity per block). The §7.2 pre-first-token marker row and the §7.3 zero-output abnormal row are persisted with `contentBlocks: []`, so their terminal frame has **no entities** — and therefore nothing to carry `metadata` on. The unqualified claim here ("renders the interrupted / \"Stopped\" / abnormal-finish state at the terminal frame exactly as an `agent.getConversation` reader would, with no refetch") does not hold for exactly the case §7.3 calls "the only durable record of the ending".

Either the intentd change adds a block-less carrier for that case (then the doc needs to describe it), or this paragraph needs a caveat, e.g.: *"A zero-output marker row (`contentBlocks: []`, §7.2 / §7.3) yields a terminal frame with no entities and so carries no `metadata`; a client rendering that ending from `chat.subscribe` alone must take it from the `agent:stream:end` payload (`stopReason` / `interruptReason` / `finishReason`) or refetch."*

docs/protocol/07-agent-streaming.md
```diff
@@ -585,6 +585,20 @@ live, carrying the authoritative `messageSeq`/`timestamp`/`streamingComplete:tru
...
+— emitted when the re-read fails and the channel falls back to sealing the live-accumulated blocks
+in place — has no persisted row to lift from and therefore carries **no** `metadata`; a client that
```
panghy: Nit (precision): the fallback fires after the re-read fails **and its one retry** also fails (or when a successful read carries no `messages` array). Suggest "when the re-read (and its retry) fails" so a reader doesn't assume a single transient error degrades the frame.

docs/protocol/07-agent-streaming.md
```diff
@@ -585,6 +585,20 @@ live, carrying the authoritative `messageSeq`/`timestamp`/`streamingComplete:tru
 `removedIds` for any orphaned live block). The guarantee: **the seq-0 snapshot reduced with every
 delta — honoring `removedIds` — equals a fresh `agent.getConversation` snapshot.**
 
+**Terminal entities carry the row's `metadata`** ([intent-hq/intent#4409](https://github.com/intent-hq/intent/issues/4409)).
+When the re-read assistant row carries `metadata` (the §7.2 interrupt tag set —
+`interrupted` / `stopReason` / `interruptReason` / `interruptedBy` — or the §7.3 `finishReason`),
```
panghy, line 591: Nit: since the lift is **verbatim**, the entity carries whatever the row has — today that includes `status: "interrupted"` on interrupted rows (`agent_session.rs` interrupt flush), which is not part of the §7.2 documented tag set. Suggest phrasing the parenthetical as illustrative ("e.g. the §7.2 interrupt tag set … or the §7.3 `finishReason`") rather than as the closed set of what can appear.

docs/protocol/07-agent-streaming.md
```diff
@@ -622,7 +636,12 @@ A block's first appearance arrives as `added` with the same `block.id`; each gro
 carrying the **full** block (or, incremental mode, the fragment-only `textDelta`). A tool call
 arrives as an `added` `tool_use` block, then a `tool_result`
 block once output lands. The terminal frame (after `agent:stream:end`) carries the persisted blocks
-with `streamingComplete:true` and any orphan ids in `removedIds` — full `text` in both encodings.
+with `streamingComplete:true` and any orphan ids in `removedIds` — full `text` in both encodings —
+plus the row's `metadata` when present (e.g. an interrupted turn's terminal entity:
+`{ "agentId":"agent-123","messageId":"0190a200-asst","role":"assistant","messageSeq":8,
+"timestamp":"2026-06-27T01:00:09.000Z","streamingComplete":true,
+"metadata":{ "interrupted":true,"stopReason":"interrupted","interruptReason":"user_stop" },
```
panghy: Example vs "verbatim": the persisted interrupted row also carries `"status":"interrupted"`, so a real terminal entity's `metadata` would include it. Either add it to the example or mark the object as abbreviated, otherwise the example contradicts the verbatim claim two paragraphs up.

docs/protocol/07-agent-streaming.md
```diff
@@ -622,7 +636,12 @@ A block's first appearance arrives as `added` with the same `block.id`; each gro
 carrying the **full** block (or, incremental mode, the fragment-only `textDelta`). A tool call
 arrives as an `added` `tool_use` block, then a `tool_result`
 block once output lands. The terminal frame (after `agent:stream:end`) carries the persisted blocks
-with `streamingComplete:true` and any orphan ids in `removedIds` — full `text` in both encodings.
+with `streamingComplete:true` and any orphan ids in `removedIds` — full `text` in both encodings —
+plus the row's `metadata` when present (e.g. an interrupted turn's terminal entity:
+`{ "agentId":"agent-123","messageId":"0190a200-asst","role":"assistant","messageSeq":8,
```
panghy: Nit: `"messageSeq":8` is inconsistent with the worked example — the seq-0 snapshot has the user row at `seq:0` with `totalMessages:1`, so the assistant row that follows would be `messageSeq:1`.

docs/protocol/versioning.md
```diff
@@ -30,6 +30,8 @@ Also within 9.7 (additive manifest / archive fields, presence-detected per the c
...
 
+Also within 9.7 (additive entity field, presence-detected per the convention below; no method-catalog or wire-shape change, so no version bump — [intent-hq/intent#4409](https://github.com/intent-hq/intent/issues/4409)): **`chat.subscribe` terminal deltas carry the persisted assistant row's `metadata`** (§7.1 "Terminal entities carry the row's `metadata`") — every entity in the `agent:stream:end` terminal reconcile frame lifts the re-read row's `metadata` map verbatim (the §7.2 interrupt tag set `interrupted` / `stopReason` / `interruptReason` / `interruptedBy`, or the §7.3 `finishReason`), under the same rule the non-assistant row deltas have followed since intentd#747: present only when the persisted row has a non-null metadata map, omitted entirely otherwise, never `null`. Rows without metadata keep the lean entity shape byte-for-byte, live chunk deltas are unchanged, and the degraded best-effort terminal frame (no persisted row to lift from) carries no `metadata` — so a `chat.subscribe`-only client renders the interrupted / abnormal-finish state at the terminal frame identically to an `agent.getConversation` reader, with no refetch.
```
panghy: Two nits on this note:

1. **History attribution:** the metadata lift on non-assistant row deltas landed in [intent-hq/intentd#773](https://github.com/intent-hq/intentd/pull/773) (`227c5466`, "carry row metadata on message deltas"), not #747 — #747 introduced the user-row deltas themselves without `metadata`. Suggest "since intentd#773".
2. **Wording convention:** "no method-catalog or wire-shape change" is self-contradictory for an additive wire field. Precedent entries for additive fields write "no method-catalog change" (9.4 `isBackground?`, 8.6 `contextUsage?`) or "no method-catalog or **existing**-wire-shape change" (8.7 hook schedule kinds). Suggest the latter.

Also, per the PR body, the intentd PR link still needs adding here before merge.

panghy (commented):
Review of the docs-only change (verified against the current `intent-transport/src/subscriptions.rs` reconcile path and `agent_session.rs` row metadata). Wording/placement conventions and links are fine; the versioning note sits correctly in the 9.7 block and follows the "Also within" pattern. One substantive accuracy gap (zero-block marker rows) and a handful of factual nits below. Not approving/merging — merge stays gated on the intentd PR per the body.

docs/protocol/07-agent-streaming.md
```diff
@@ -585,6 +585,20 @@ live, carrying the authoritative `messageSeq`/`timestamp`/`streamingComplete:tru
 `removedIds` for any orphaned live block). The guarantee: **the seq-0 snapshot reduced with every
 delta — honoring `removedIds` — equals a fresh `agent.getConversation` snapshot.**
 
+**Terminal entities carry the row's `metadata`** ([intent-hq/intent#4409](https://github.com/intent-hq/intent/issues/4409)).
+When the re-read assistant row carries `metadata` (the §7.2 interrupt tag set —
+`interrupted` / `stopReason` / `interruptReason` / `interruptedBy` — or the §7.3 `finishReason`),
+every entity in the terminal frame additionally carries it **verbatim** as `metadata`, under the
+same rule as the non-assistant row deltas above: present only when the persisted row has a
+non-null metadata map, omitted entirely otherwise, **never** `null` (additive, presence-detected —
+rows without metadata keep the lean entity shape). So a `chat.subscribe`-only client renders the
```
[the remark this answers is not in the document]
panghy: Addressed in c804260: the §7.1 paragraph now scopes the no-refetch claim to rows with content blocks and adds a "Zero-block rows have no carrier" sentence pointing clients at the agent:stream:end payload (stopReason / interruptReason / interruptedBy / finishReason) or a refetch. The versioning note carries the same caveat.

docs/protocol/07-agent-streaming.md
panghy: Addressed in c804260 (same fix as the thread above): explicit zero-block caveat + agent:stream:end payload / refetch fallback.

docs/protocol/07-agent-streaming.md
panghy: Addressed in c804260: now "when the re-read (and its retry) fails".

docs/protocol/07-agent-streaming.md
panghy: Addressed in c804260: the parenthetical is now "e.g. …" and the sentence states the lift is the whole persisted map, including keys this document does not enumerate.

docs/protocol/07-agent-streaming.md
panghy: Addressed in c804260: the example metadata now includes "status":"interrupted" (verified against the interrupt flush in agent_session.rs).

docs/protocol/07-agent-streaming.md
panghy: Addressed in c804260: messageSeq is now 1.

docs/protocol/versioning.md
panghy: Addressed in c804260: attribution corrected to intentd#773 and wording changed to "no method-catalog or existing-wire-shape change". The intentd PR link will be added once that PR is open.

PR #4427, opened 2026-09-05
feat(triage): classify issues by the Type field and retire the bug/enhancement labels

panghy:
## Summary

Issues on intent-hq/intent are classified by the GitHub issue **Type** field (Bug / Feature / Task) only. The redundant "bug" and "enhancement" type labels are retired: they are no longer documented, no longer applied by the issue templates or by triage, and the automatic triage removes them when someone still applies one. The "question" label is unchanged and stays a regular label.

## Changes

- **Docs** (AGENTS.md, CONTRIBUTING.md, docs/README.md): Type is the sole classification; "gh issue create --type Bug" recipe (gh >= 2.94.0) plus a GraphQL "updateIssue" fallback for older gh; labels bullet limited to component:* + agent-filed.
- **Issue templates**: dropped the "labels:" front-matter; both forms keep their "type:".
- **Triage pass 1** (issue-triage.yml, parse-issue.js): new deterministic step on opened / edited / reopened, and on `labeled` when the applied label is "bug" / "enhancement" (a labeled run executes only this step) — when an issue carries "bug" or "enhancement", fill the Type if empty (bug -> Bug, enhancement -> Feature; ids resolved at runtime from repository.issueTypes; the Type is re-read right before the write so a Type set in the meantime is kept), then remove the label. The label is removed only once the issue has a Type; any lookup/write failure logs a warning and leaves the label in place (fail-soft). Dry-run honored. This is the one documented exception to the pass-1 add-only contract.
- **Triage pass 2** (agentic-triage.js): the model's type inference now feeds the Type field directly and is never applied as a bug/enhancement label; legacy labels are removed under the same rule as pass 1 (after a Type exists, skipped when the Type write fails); shared label->Type mapping imported from parse-issue.js; prompt, summary comment and --dry-run output updated.

## Verification

- node --test on parse-issue / needs-info / agentic-triage: 84 pass, 0 fail (same command as the ci.yml triage-parser-test job).
- Workflow YAML parses; the new github-script step was extracted and executed against mocked github/core objects across 10 scenarios (set+remove, existing Type -> remove only, Types absent/disabled, write failure, dry-run, question-only no-op, removeLabel 404, ...).
- GraphQL shapes (Repository.issueTypes, UpdateIssueInput.issueTypeId) validated read-only against the live repo.
- Post-merge check: gh workflow run issue-triage.yml -f issue_number=<N> -f dry_run=true on an issue that carries "enhancement" should print the intended Type write / label removal notices.

## Follow-up (manual, after merge)

Delete the "bug" and "enhancement" labels from the repository (gh label delete) — this strips them from all existing issues at once. Not done by automation.

### DIFF
--- a/.github/scripts/issue-triage/agentic-triage.js
+++ b/.github/scripts/issue-triage/agentic-triage.js
@@ -417,6 +417,22 @@ function planIssueType({ response, currentLabels, currentIssueType, issueTypes }
   };
 }
 
+// Reconcile the plan with the outcome of the Type write (setIssueType):
+//   - `true`: written by this run — the plan stands,
+//   - `'existing'`: a human set a Type between plan and write — nothing
+//     to report as set, but the issue HAS a Type, so the legacy type
+//     labels are still retired,
+//   - `false`: the write failed — no Type on the issue, so the legacy
+//     label stays (the issue remains classified; pass 1 retires it once a
+//     Type is present).
+// Pure, so the gating is unit-testable; returns the (mutated) plan.
+function applyTypeWriteOutcome(plan, outcome) {
+  if (outcome === true) return plan;
+  plan.issueType = null;
+  if (outcome !== 'existing') plan.removeLabels = [];
+  return plan;
+}
+
 // Resolve a single-select option by field name + option name against the
 // runtime `repository.issueFields` list. Null when the field or option is
 // unknown (fields disabled, lookup failed, option renamed) — never a guess.
@@ -598,6 +614,7 @@ module.exports = {
   NEEDS_INFO_LABEL,
   NEEDS_TRIAGE_LABEL,
   POSSIBLE_DUPLICATE_LABEL,
+  applyTypeWriteOutcome,
   buildPrompt,
   buildSummaryComment,
   extractSearchQueries,
@@ -695,12 +712,15 @@ function fetchIssueContext(repo, issueNumber) {
   }
 }
 
-// Set the Type via `updateIssue(issueTypeId)`. Returns true on success;
-// a failure (e.g. a token that cannot set Types) is a warning, not an
+// Set the Type via `updateIssue(issueTypeId)`. Returns true on success,
+// `'existing'` when the issue already has a Type, false on failure — a
+// failure (e.g. a token that cannot set Types) is a warning, not an
 // abort — the rest of the pass still completes. The planning read is a
 // point-in-time snapshot and `updateIssue` replaces unconditionally, so
 // the Type is re-read immediately before the write: a Type set by a human
-// in the meantime (or during the label write) is left alone.
+// in the meantime (or during the label write) is left alone (`'existing'`,
+// distinct from a failed write so the caller can still retire the legacy
+// label — see applyTypeWriteOutcome).
 function setIssueType(issueNodeId, issueType) {
   try {
     const fresh = ghJson([
@@ -714,7 +734,7 @@ function setIssueType(issueNodeId, issueType) {
       console.log(
         `issue Type is now ${node.issueType.name} (set since the plan was made); leaving it unchanged.`
       );
-      return false;
+      return 'existing';
     }
     gh([
       'api', 'graphql',
@@ -1089,14 +1109,15 @@ function main() {
   }
 
   // Order matters for partial-failure recovery: labels, then the Type,
-  // then the legacy type-label removal (only once the Type write
-  // succeeded, or the issue already had a Type), then the Priority /
-  // Effort fields, then the summary comment (the audit record +
-  // idempotency marker), and only then retire needs-triage — a failure
-  // before that point leaves the issue in the triage queue for a re-run or
-  // a human. The comment is built after the Type, label-removal and field
-  // writes so a fail-soft failure (or a value filled by a human in the
-  // meantime) is not reported as applied.
+  // then the legacy type-label removal (only once the issue has a Type —
+  // written by this run, pre-existing, or set by a human in the meantime;
+  // fail-soft, a failed removal is a warning and the label is retried by
+  // pass 1's next run), then the Priority / Effort fields, then the
+  // summary comment (the audit record + idempotency marker), and only then
+  // retire needs-triage — a failure before that point leaves the issue in
+  // the triage queue for a re-run or a human. The comment is built after
+  // the Type, label-removal and field writes so a fail-soft failure (or a
+  // value filled by a human in the meantime) is not reported as applied.
   if (plan.addLabels.length > 0) {
     gh([
       'issue', 'edit', String(issueNumber), '--repo', repo,
@@ -1105,27 +1126,27 @@ function main() {
     console.log(`added labels: ${plan.addLabels.join(', ')}`);
   }
   if (plan.issueType) {
-    if (setIssueType(context.issueNodeId, plan.issueType)) {
+    const outcome = setIssueType(context.issueNodeId, plan.issueType);
+    if (outcome === true) {
       console.log(`set issue Type: ${plan.issueType.name}`);
-    } else {
-      // No Type was written by this run: keep the legacy label so the
-      // issue stays classified (pass 1 retires it on the next edit once a
-      // Type is present).
-      plan.issueType = null;
-      if (plan.removeLabels.length > 0) {
-        console.log(
-          `issue Type not set by this run; keeping legacy label(s): ${plan.removeLabels.join(', ')}`
-        );
-        plan.removeLabels = [];
-      }
+    } else if (outcome === false && plan.removeLabels.length > 0) {
+      console.log(
+        `issue Type not set; keeping legacy label(s): ${plan.removeLabels.join(', ')}`
+      );
     }
+    applyTypeWriteOutcome(plan, outcome);
   }
   if (plan.removeLabels.length > 0) {
-    gh([
-      'issue', 'edit', String(issueNumber), '--repo', repo,
-      ...plan.removeLabels.flatMap((l) => ['--remove-label', l]),
-    ]);
-    console.log(`removed legacy type labels: ${plan.removeLabels.join(', ')}`);
+    try {
+      gh([
+        'issue', 'edit', String(issueNumber), '--repo', repo,
+        ...plan.removeLabels.flatMap((l) => ['--remove-label', l]),
+      ]);
+      console.log(`removed legacy type labels: ${plan.removeLabels.join(', ')}`);
+    } catch (e) {
+      warn(`could not remove legacy type labels ${plan.removeLabels.join(', ')} (fail-soft): ${e.message}`);
+      plan.removeLabels = [];
+    }
   }
   const plannedFields = ['priority', 'effort'].filter((k) => plan.issueFields[k]);
   if (plannedFields.length > 0) {
--- a/.github/scripts/issue-triage/agentic-triage.js
+++ b/.github/scripts/issue-triage/agentic-triage.js
@@ -2,16 +2,21 @@
 // Agentic triage for the issue-triage workflow (pass 2).
 //
 // LLM judgment pass over a newly opened issue, run after the deterministic
-// pass 1: duplicate-candidate detection, component/type inference when the
+// pass 1: duplicate-candidate detection, component inference when the
 // template checkboxes left the issue unlabeled, the issue Type field (Bug /
 // Feature / Task) when the issue has none, the Priority (Urgent / High /
 // Medium / Low, with security escalation) and Effort (Low / Medium / High)
 // issue fields when they are empty, and ONE auditable summary comment.
+// Classification is Type-native: the model's type inference feeds the
+// Type field directly and is never written as a `bug` / `enhancement`
+// label (those are retired; `question` stays a regular label).
 // Suggest-don't-destroy: it only ever ADDS labels and comments and only
 // ever FILLS an empty Type or field — it never closes issues, never edits
-// bodies, never overwrites an existing Type or field value, and the only
-// label it removes is `needs-triage` (the triage queue marker) after a
-// successful pass.
+// bodies, never overwrites an existing Type or field value. The only labels
+// it removes are `needs-triage` (the triage queue marker) after a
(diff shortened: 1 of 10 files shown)


AGENTS.md
```diff
@@ -223,6 +223,23 @@ dogfooding intentd + cloudlands-fe for daily development work), file a GitHub is
 [intent-hq/intent](https://github.com/intent-hq/intent/issues) — the single tracker
 for all components.
 
+- **Type**: classification is the GitHub issue **Type** field — `Bug`, `Feature`,
+  or `Task` — not a label. The `bug` and `enhancement` type labels are retired: do
+  not apply them (triage removes them after setting the matching Type; the
```
panghy: **should-fix (doc accuracy)** — "triage removes them after setting the matching Type" reads as if applying the label anywhere triggers retirement, but the workflow only runs on `opened` / `edited` / `reopened` (no `labeled` trigger). A `bug` label applied from the sidebar to an existing issue stays until the next edit/reopen (or a manual `workflow_dispatch`). Suggest making the timing explicit, e.g. "(triage retires them on the issue's next open / edit / reopen, after setting the matching Type; …)". The PR body's "removes them when someone still applies one" has the same nuance. Adding `labeled` to the trigger list would close the gap but re-runs the whole pass-1 job (incl. the needs-info check) on every label change, so a follow-up rather than this PR.

.github/workflows/issue-triage.yml
```diff
@@ -5,9 +5,26 @@ name: Issue triage
...
+# and `enhancement` labels are retired. When an issue still carries one,
+# the step fills the Type if it is EMPTY (bug -> Bug, enhancement ->
+# Feature; the first label in LEGACY_TYPE_LABELS order decides when both
+# are present; Type ids resolved at runtime from `repository.issueTypes`,
+# an existing Type is never overwritten) and THEN removes the label(s). A
+# label is removed only once the issue has a Type — if the Type cannot be
+# set (lookup/write failure, Types disabled) the label stays and a warning
+# is logged (fail-soft). `question` is not legacy: it is never removed.
+# Runs in the `triage` job only (opened / edited / reopened / dispatch);
+# the `needs-info-reply` re-run deliberately skips it — the retirement is
```
panghy: **nit (doc accuracy)** — "the retirement is not trivial to mirror there" undersells it: it is the same step body. A more precise rationale is that an author reply on a needs-info issue cannot introduce a type label, so there is nothing new to retire and the next `edited` / `reopened` run covers it.

.github/workflows/issue-triage.yml
```diff
@@ -207,6 +230,104 @@ jobs:
               core.warning(`Could not set Priority field on #${issueNumber} (${e.message}); continuing.`);
             }
 
+      # The one pass-1 removal (see header): convert a legacy `bug` /
+      # `enhancement` label into the issue Type when empty, then remove it.
+      # Type first, label second, so a failed Type write leaves the label in
+      # place; every lookup/write is fail-soft (warning, job stays green).
+      - name: Retire legacy type labels into the issue Type field
```
panghy, line 255: **nit (test coverage, follow-up)** — the pure planner is well covered, but the step's fail-soft branches (query failure, Types absent, mutation failure, `removeLabel` 404 vs other, dry-run) live inline in YAML and are outside `node --test` / `triage-parser-test`. I exercised them by extracting the script and running it against mocked `github` / `core` (11 scenarios, all as documented). Moving the body into e.g. `.github/scripts/issue-triage/retire-type-labels.js` exporting `run({ github, core, context, issueNumber, dryRun })` would let CI keep covering it. Consistent with the existing inline Priority step, so not blocking.

.github/workflows/issue-triage.yml
```diff
@@ -207,6 +230,104 @@ jobs:
...
+                }`,
+                { owner: context.repo.owner, name: context.repo.repo, number: issueNumber }
+              ));
+              if (!repository || !repository.issue) throw new Error('empty repository/issue in response');
+            } catch (e) {
+              core.warning(`Could not read issue Types/labels for #${issueNumber} (${e.message}); legacy type labels left in place.`);
+              return;
+            }
+            const issue = repository.issue;
+            const labels = (issue.labels.nodes || []).map((l) => l && l.name).filter(Boolean);
```
panghy: **nit** — `issue.labels.nodes` is dereferenced outside the `try`, while `repository.issueTypes` on the next lines is guarded. `LabelConnection` is nullable in the schema (in practice always present for issues), so a `TypeError` here would fail the step rather than warn. `((issue.labels && issue.labels.nodes) || [])` keeps the whole step fail-soft.

.github/workflows/issue-triage.yml
```diff
@@ -207,6 +230,104 @@ jobs:
...
+
+            if (dryRun) {
+              if (plan.setType) core.notice(`[dry-run] Would set issue Type = ${plan.setType.name} on #${issueNumber}`);
+              core.notice(`[dry-run] Would remove legacy type labels from #${issueNumber}: ${plan.removeLabels.join(', ')}`);
+              return;
+            }
+
+            if (plan.setType) {
+              try {
+                await github.graphql(
```
panghy: **nit** — pass 2's `setIssueType` re-reads the Type immediately before `updateIssue` (documented there: the plan is a snapshot and `updateIssue` replaces unconditionally); this pass-1 write relies on the read a few lines above. The window is a single round-trip so the risk is negligible — noting the asymmetry only.

.github/scripts/issue-triage/agentic-triage.js
```diff
@@ -1070,9 +1108,25 @@ function main() {
     if (setIssueType(context.issueNodeId, plan.issueType)) {
```
panghy: **nit** — `setIssueType` returns `false` both when the write failed and when a human set the Type between plan and write. In the second case the issue *does* have a Type, but `removeLabels` is cleared below, so the legacy label survives until the next `edited` / `reopened` pass-1 run. Returning a distinct value for the already-set case (e.g. `'existing'`) would let the removal proceed. Already recorded as an accepted nit in the workspace spec.

.github/scripts/issue-triage/agentic-triage.js
```diff
@@ -1070,9 +1108,25 @@ function main() {
...
+          `issue Type not set by this run; keeping legacy label(s): ${plan.removeLabels.join(', ')}`
+        );
+        plan.removeLabels = [];
+      }
     }
   }
+  if (plan.removeLabels.length > 0) {
+    gh([
+      'issue', 'edit', String(issueNumber), '--repo', repo,
+      ...plan.removeLabels.flatMap((l) => ['--remove-label', l]),
```
panghy: **nit** — a failing `gh issue edit --remove-label` throws out of `main()` (not fail-soft, unlike the Type / field writes), aborting before the summary comment and the `needs-triage` retirement. Consistent with the `--add-label` write above, recoverable on re-run (the Type is already set, so the re-run removes the label), and the job is `continue-on-error` — just noting the asymmetry with the header comment's "fail-soft" framing.

CONTRIBUTING.md
```diff
@@ -87,13 +87,15 @@ Keep the relevant checks green before opening a PR:
 ## Filing issues
 
 - Use the [issue forms](https://github.com/intent-hq/intent/issues/new/choose)
-  (bug report / feature request) and pick the affected component(s).
+  (bug report / feature request) and pick the affected component(s). The forms
+  set the issue **Type** (Bug / Feature / Task) — that field is the
+  classification; there is no `bug` or `enhancement` label to add.
```
panghy: **nit** — "there is no `bug` or `enhancement` label to add" is not literally true until the manual label deletion in the PR's follow-up happens (they still exist on the repo and in the label picker). "do not add a `bug` or `enhancement` label" is accurate before and after.

panghy (commented):
## Review (agent) — no blockers

Reviewed against the workspace spec (Type-only classification; pass 1 converts a legacy label into the Type and removes it only once a Type is present; pass 2 never adds `bug` / `enhancement` and removes them under the same rule; `question` unchanged).

**Verified**
- `node --test` on parse-issue / needs-info / agentic-triage: 84 pass, 0 fail (same command as the `triage-parser-test` CI job).
- GraphQL shapes validated read-only against the live repo: `Repository.issueTypes { id name isEnabled }` (Task / Bug / Feature, all enabled), `Issue.issueType { name }`, `Issue.labels`, and `UpdateIssueInput.issueTypeId` exists in the schema.
- The new github-script step, extracted from the YAML and run against mocked `github` / `core` across 11 scenarios: set+remove; existing Type → remove only; Types list empty / `issueTypes` null → warning, label kept; query failure → warning, label kept; mutation failure → warning, label kept; dry-run → two notices, no writes; `question`-only → no-op, no warning; `removeLabel` 404 tolerated; non-404 → warning, continue; invalid issue number → `setFailed`. All behaved as documented. Workflow YAML parses.
- Pass 2: `planActions` has no path that pushes `bug` / `enhancement` into `addLabels` (only `QUESTION_LABEL`); `removeLabels` is gated on `currentIssueType || issueType` (precedence is `(a || b) ? … : []`, correct); `--dry-run` prints the removal plan and returns before any write; `--fields-only` returns before the label code. Type write precedes label removal and a failed Type write clears `removeLabels`.
- Docs: `gh issue create --type` landed in gh v2.94.0 (cli/cli#13057), so the `≥ 2.94.0` note is accurate; the GraphQL fallback recipe matches the live schema.
- Neither `labeled`/`unlabeled` nor `typed` are workflow triggers, so the removal / Type write cannot re-trigger the job.

**Findings** are inline: 1 should-fix (doc wording overstates when retirement happens — the workflow has no `labeled` trigger), the rest nits.

Not approving/merging — human decision.

.github/workflows/issue-triage.yml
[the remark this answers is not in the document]
panghy: Fixed in 2e2b0f9: `labeled` is now in the `issues` trigger types. The `triage` job runs on a `labeled` event only when `github.event.label.name` is `bug` or `enhancement`, and on such a run only the retirement step executes (the label/Priority step and the needs-info step carry `if: github.event.action != 'labeled'`). `agentic-triage` stays `opened`-only. Header comment updated.

AGENTS.md
panghy: Done in 2e2b0f9 — went with adding the gated `labeled` trigger (only the retirement step runs on it, so the needs-info check and label derivation are not re-run on label changes), and the AGENTS.md sentence now reads "triage retires them — on the issue's open / edit / reopen, or as soon as the label is applied — after setting the matching Type". PR body updated with the same nuance.

.github/workflows/issue-triage.yml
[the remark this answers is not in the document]
panghy: Fixed in 2e2b0f9: the step re-reads the Type via `node(id) { ... on Issue { issueType { name } } }` immediately before `updateIssue` (same pattern as pass 2's `setIssueType`). If a Type is present by then the write is skipped and the issue is treated as already typed, so the label removal still proceeds; a failing re-read is a warning and leaves the label in place. Exercised against mocked `github`/`core` (12 scenarios incl. the race and re-read-failure paths).

panghy: Mirrored in 2e2b0f9: pass 1 now re-reads the Type right before `updateIssue`, skipping the write (and still retiring the label) when a Type appeared in between.

.github/scripts/issue-triage/agentic-triage.js
```diff
@@ -1070,9 +1108,25 @@ function main() {
     if (setIssueType(context.issueNodeId, plan.issueType)) {
       console.log(`set issue Type: ${plan.issueType.name}`);
     } else {
+      // No Type was written by this run: keep the legacy label so the
+      // issue stays classified (pass 1 retires it on the next edit once a
+      // Type is present).
       plan.issueType = null;
+      if (plan.removeLabels.length > 0) {
```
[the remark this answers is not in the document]
panghy: Fixed in adce0a4: `setIssueType` returns `'existing'` when the pre-write re-read finds a Type (distinct from `false` on a failed write). The gating moved into a pure `applyTypeWriteOutcome(plan, outcome)`: only a real failure clears `removeLabels`; `'existing'` drops the Type from the report but keeps the removal. Unit-tested for all three outcomes.

.github/scripts/issue-triage/agentic-triage.js
panghy: Done in adce0a4 — `setIssueType` now returns `'existing'` for the already-set case and the removal proceeds; gating extracted into `applyTypeWriteOutcome` with tests.

.github/scripts/issue-triage/agentic-triage.js
panghy: Done in adce0a4: the `--remove-label` write is wrapped in try/catch — a failure logs a warning, clears `removeLabels` (so the summary comment does not report a removal that did not happen) and the pass continues to the comment and the `needs-triage` retirement. Header comments in the script and workflow updated to say so.

.github/workflows/issue-triage.yml
panghy: Done in 2e2b0f9: `((issue.labels && issue.labels.nodes) || [])`.

.github/workflows/issue-triage.yml
panghy: Reworded in 2e2b0f9: the header now says an author reply cannot introduce a type label, so there is nothing new to retire, and the next `edited` / `reopened` / `labeled` run covers it (the `needs-info-reply` job comment says the same).

CONTRIBUTING.md
panghy: Done in 2e2b0f9: "do not add a `bug` or `enhancement` label".

.github/workflows/issue-triage.yml
panghy: Keeping the step inline in this PR for consistency with the existing Priority step; noted as a follow-up to move both pass-1 field/Type steps into `.github/scripts/issue-triage/` modules so `triage-parser-test` covers their fail-soft branches. For this round the step was re-exercised against mocked `github`/`core` (12 scenarios incl. the new re-read path and the null-labels guard).

PR #4693, opened 2026-09-11
fix(make): refresh FE deps when pnpm-lock.yaml changes

panghy:
## Summary

The eight FE targets in the root `Makefile` (`dev-ui`, `dev-sandbox-ui`, `dev-sandbox-app`, `dev-sandbox-stack`, `fe-launch`, `run-fe-local`, `build-sidecar`, `dev`) guarded their `pnpm install` with `[ -d $(FE_DIR)/node_modules ]`, so a lockfile change on a checkout with an existing tree never reinstalled and the target ran against stale deps.

This replaces those guards with one shared `FE_DEPS_FRESH` condition: fresh only when `node_modules/.pnpm/lock.yaml` exists **and** is byte-identical to `pnpm-lock.yaml` (pnpm writes the lockfile it installed from to that path). Each target keeps its existing install command (corepack + `--frozen-lockfile` for sandbox targets, plain `pnpm install` elsewhere) and log prefix; only the condition and the log text change.

Companion to intent-hq/cloudlands-fe#2321 (FE-side freshness gate); this implementation is inline shell and does not depend on it. Incident that motivated the work: intent-hq/cloudlands-fe#2310, where a lockfile change on an existing `node_modules` tree left targets running against stale deps.

## Changes

- `Makefile`: add `FE_DEPS_FRESH` / `FE_DEPS_INSTALL_MSG` next to `FE_DIR`; use them in all eight guards; update the `fe-launch` comment.
- `docs/fe/DEVELOPER_GUIDE.md`: the "installs when `node_modules` is missing" sentence now also mentions a mismatched lockfile.

## Verification

| Check | Result |
|---|---|
| `grep -n 'node_modules \]' Makefile` | no output |
| `make -n dev-sandbox-ui` / `fe-launch` / `build-sidecar` / `dev` | each shows the new `FE_DEPS_FRESH \|\| (...)` guard with its original install command |
| `make -n status`, `make help`, `make docs-check` | pass |
| Condition extracted from `make -n`, fresh install | no-op |
| Same, after appending a line to `pnpm-lock.yaml` (then reverted) | install branch triggers |
| Same, with `FE_DIR` pointing at a dir without `node_modules` | install branch triggers |

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,13 @@ INTENTD_DIR = packages/intentd
 FE_DIR = packages/cloudlands-fe
 IOS_DIR = packages/ios
 
+# FE dependency freshness guard shared by every target that needs an installed
+# frontend tree. pnpm writes the lockfile it installed from to
+# node_modules/.pnpm/lock.yaml, so the tree is stale when that copy is missing
+# or no longer byte-identical to pnpm-lock.yaml. Use as `$(FE_DEPS_FRESH) || (...)`.
+FE_DEPS_FRESH = [ -f $(FE_DIR)/node_modules/.pnpm/lock.yaml ] && cmp -s $(FE_DIR)/pnpm-lock.yaml $(FE_DIR)/node_modules/.pnpm/lock.yaml
+FE_DEPS_INSTALL_MSG = installing/refreshing FE deps (lockfile changed or node_modules missing)
+
 # cargo install may place subcommands outside PATH when cargo itself comes from
 # a distro package. Make every recipe discover the effective install bin dir.
 CARGO_BIN_DIR ?= $(or $(CARGO_INSTALL_ROOT),$(CARGO_HOME),$(HOME)/.cargo)/bin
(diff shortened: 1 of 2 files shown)


panghy (commented):
## Review result

No blocking code findings at e35253c6a3685b5a5e74c4553db71e4d45d98329. This is a COMMENT review, not an approval.

### Non-blocking: correct the companion PR link

The Summary currently calls intent-hq/cloudlands-fe#2310 the companion FE-side freshness check. That PR is actually `refactor(store): drop the browserClients tab mirror`, the incident motivating this work. Please identify [intent-hq/cloudlands-fe#2321](https://github.com/intent-hq/cloudlands-fe/pull/2321) as the companion freshness-gate PR and retain #2310 separately as the incident reference. This is PR-description metadata only, not an implementation defect.

### Verification performed

- All eight expanded guard recipes tested against isolated fixtures using Linux `/bin/sh` (dash) and `bash --posix`, with `-e` enabled: fresh locks, differing locks, absent node_modules, absent installed lock, installed-lock path that is a directory, and absent source lock. Each case exercised mocked installer success and failure: 192 executions passed. Fresh guards exited successfully without invoking an installer; stale/missing guards invoked the correct installer; installer exit 23 propagated and prevented continuation. No real installs, builds, services, or repository edits were performed.
- Compared every replacement to its parent revision: the four `corepack pnpm install --frozen-lockfile` and four plain `pnpm install` commands, plus all target-specific log prefixes, are unchanged. No old directory-only guard remains.
- POSIX shell constructs and `cmp -s` are suitable for Linux/macOS. Native macOS execution was not available; that platform assessment is static. The fixed, relative FE_DIR path is safe as used; support for whitespace in a custom FE_DIR remains an existing limitation of the recipes, not a regression introduced here.
- `make -n status`, `make help`, `make docs-check`, all eight target dry runs, and `git diff --check HEAD^ HEAD` passed.
- Diff is limited to Makefile and the intended docs sentence; no submodule pin, CI, or AGENTS changes. The docs accurately describe dev-sandbox-ui, and both the sole commit and PR title are conventional.

panghy:
Fixed the companion link from review https://github.com/intent-hq/intent/pull/4693#pullrequestreview-5175642494: the Summary now names intent-hq/cloudlands-fe#2321 as the companion FE-side freshness gate and cites intent-hq/cloudlands-fe#2310 separately as the motivating incident. PR-description change only; no code changes.

PR #4716, opened 2026-09-11
fix: make scripts/shipped-in.sh run on stock macOS Bash 3.2

### LINKED ISSUE #4706: Release detector fails on stock macOS Bash
## Observed failure
After cloudlands-fe PR #2313 merged (squash a6c67ebfd080a40d1303ce3f80e91c000c0d2f9e), the documented release watcher failed during registration. scripts/shipped-in.sh cloudlands-fe a6c67ebfd080a40d1303ce3f80e91c000c0d2f9e exits 2 at line68: declare: -A: invalid option. Host is macOS with /bin/bash 3.2.57; make status reports doctorOk=true. No newer Bash is available in the standard Homebrew executable locations.

## Cause and expected behavior
scripts/shipped-in.sh uses Bash4 associative arrays (lines68/99) and mapfile (line96), but its env bash entrypoint resolves the stock macOS shell. The script should run on supported macOS hosts without requiring agents to install an undocumented interpreter or reimplement release detection. Preserve the existing tag scan, containment semantics, manifests, and exit codes.

## Proposed fix and verification
Make the existing entrypoint portable to the supported shell/runtime, and extend scripts/shipped-in.test.sh with an explicit stock-macOS Bash3.2 run plus modern Bash/Linux coverage. Keep all fake-gh fixtures for oldest carrying version, rolling-channel filtering, intentd manifest pins, behind/diverged rejection, API failure, and no-hit exit3. Use offline tests; no release publication or credentials required.

## Scope and dedup
Separate monorepo tooling follow-up, not part of the video feature. Searched all issues for shipped-in in:title, shipped-in with bash, and Bash3; no matching detector portability issue. Existing DX-setup release-test plan concerns frontend release transformations, not this detector. A bounded read-only Python/gh fallback is temporarily tracking this workspace without changing repo files or dependencies.

panghy:
Fixes intent-hq/intent#4706

## Problem

`scripts/shipped-in.sh` used `declare -A` and `mapfile`, so on stock macOS `/bin/bash` 3.2.57 it exited 2 at line 68 (`declare: -A: invalid option`) and agents on Mac could not run the documented shipped-in check. `bash -n` accepts the broken script, so a syntax check alone does not catch this.

## Change

**`scripts/shipped-in.sh`** — Bash 3.2-compatible, same `#!/usr/bin/env bash` entrypoint:
- The two associative-array caches (manifest `intentdVersion` per tag, compare status per pinned intentd version) become newline-separated `<key> <value>` string records with a small `cache_get` helper.
- `mapfile -t tags < <(...)` becomes a `while read` loop into an indexed array.
- Tag scan, containment semantics (`ahead`/`identical` carry; `behind`/`diverged` never), manifest handling and exit codes (0 hit / 3 no-hit / 2 usage / 1 gh failure) are unchanged.

**`scripts/shipped-in.test.sh`** — Bash 3.2 compatibility pass appended after the existing fixtures:
- The interpreter running the script under test defaults to the one running the suite (`$BASH`), so `/bin/bash scripts/shipped-in.test.sh` on macOS exercises 3.2 directly.
- Static gate on both files: `bash -n` plus a grep rejecting Bash 4+ constructs (associative-array declarations, `mapfile`/`readarray`, case-modification expansions, `&>>`, `|&`, case fall-through, `coproc`, namerefs).
- When a Bash 3 interpreter is found — `BASH3_BIN`, `bash3` on PATH, Homebrew `bash@3`, or a 3.x `/bin/bash` — the whole fixture suite is re-executed under it; otherwise it prints that the real 3.2 run was skipped (this is what CI on ubuntu does).

## Verification

Bash 3.2.57 was built from the GNU tarball locally to run the suite for real:

| Run | Result |
|---|---|
| `bash scripts/shipped-in.test.sh` (no Bash 3 present) | passes under 5.2; prints "real 3.2 run skipped, static gate only" |
| `BASH3_BIN=<bash-3.2.57> bash scripts/shipped-in.test.sh` | passes under 5.2, then all fixtures pass again under 3.2.57(4)-release |
| `<bash-3.2.57> scripts/shipped-in.test.sh` | all fixtures pass under 3.2.57 as the top-level interpreter |
| new test against `origin/main`'s script | fails: gate reports lines 68, 96, 99 (`declare -A`, `mapfile`, `declare -A`) |
| gate regex sample check | 14/14 positive samples match; 0/10 legitimate expansions (`${x%% *}`, `${1#--limit=}`, `${a[0]}`, `${r/\//__}`, `;;`, `2>&1 >>log`, …) match |

Not covered: a Docker-based Bash 3 runner. The fixtures symlink host `bash`/`python3`/`grep` into a stub `PATH`, so a container cannot execute them as-is; `BASH3_BIN` accepts any native Bash 3 executable instead.

### DIFF
--- a/scripts/shipped-in.test.sh
+++ b/scripts/shipped-in.test.sh
@@ -190,11 +190,59 @@ echo "shipped-in tests passed under $("$script_bash" -c 'echo "bash $BASH_VERSIO
 # Stock macOS /bin/bash is 3.2 (intent-hq/intent#4706). `bash -n` alone
 # accepts Bash 4+ builtins and expansions, so reject them by pattern too,
 # then rerun the fixtures under a real Bash 3 when one can be found:
-# BASH3_BIN, a bash3 on PATH, Homebrew bash@3, or a 3.x /bin/bash.
+# BASH3_BIN, a bash3 on PATH, Homebrew bash@3, or a 3.x /bin/bash. The
+# pattern gate is a best-effort guard for hosts without a Bash 3; the real
+# Bash 3 fixture run is the authoritative check.
 bash -n "$script" || fail "shipped-in.sh does not parse"
 bash -n "${BASH_SOURCE[0]}" || fail "shipped-in.test.sh does not parse"
-bash4_constructs='(declare|local|typeset) +-[A-Za-z]*[An]|mapfile|readarray|\$\{[A-Za-z_][A-Za-z_0-9]*(\[[^]]*\])?(\^\^?|,,?)[^}]*\}|&>>|\|&|;;?&|coproc'
-gate_hits=$(grep -nE "$bash4_constructs" "$script" "${BASH_SOURCE[0]}" | grep -v -F 'bash4_constructs' || true)
+bash4_constructs='(^|[^A-Za-z0-9_])(declare|local|typeset)([[:blank:]]+-[A-Za-z]+)*[[:blank:]]+-[A-Za-z]*[An]([^A-Za-z]|$)|(^|[^A-Za-z0-9_])(mapfile|readarray|coproc)([^A-Za-z0-9_]|$)|\$\{([A-Za-z_][A-Za-z_0-9]*|[0-9]+|[@*#?!$-])(\[[^]]*\])?(\^\^?|,,?)[^}]*\}|&>>|\|&|;;?&'
+# Full-line comments, the pattern itself and the gate_sample table below are
+# not scanned.
+gate_matches() {
+  grep -nE "$bash4_constructs" "$@" | grep -vE '^([^:]*:)?[0-9]+:[[:blank:]]*#' |
+    grep -v -F -e 'bash4_constructs' -e 'gate_sample' || true
+}
+gate_sample() {
+  local expected=$1 sample=$2 hit
+  hit=$(printf '%s\n' "$sample" | gate_matches)
+  case "$expected:${hit:+hit}" in
+    hit:hit | miss:) ;;
+    *) fail "gate regex $expected sample misclassified: $sample" ;;
+  esac
+}
+gate_sample hit 'declare -A m=()'
+gate_sample hit 'local -gA x'
+gate_sample hit 'declare -r -A cache=()'
+gate_sample hit $'declare\t-A m'
+gate_sample hit 'typeset -n ref=x'
+gate_sample hit 'mapfile -t a'
+gate_sample hit 'readarray a <f'
+gate_sample hit 'coproc x'
+gate_sample hit 'echo ${var,,}'
+gate_sample hit 'echo ${var^^}'
+gate_sample hit 'echo ${var^}'
+gate_sample hit 'echo ${1^^}'
+gate_sample hit 'echo ${@,,}'
+gate_sample hit 'echo ${arr[1],,[a-z]}'
+gate_sample hit 'cmd &>> log'
+gate_sample hit 'cmd |& tee'
+gate_sample hit 'x) y ;;&'
+gate_sample hit 'x) y ;&'
+gate_sample miss 'local head=$1 status'
+gate_sample miss 'declare -a arr'
+gate_sample miss 'local -r x=1'
+gate_sample miss 'echo ${record%% *}'
+gate_sample miss 'echo ${1#--limit=}'
+gate_sample miss 'echo ${tags[0]}'
+gate_sample miss 'echo ${repo/\//__}'
+gate_sample miss 'a || b'
+gate_sample miss 'x) y ;;'
+gate_sample miss 'echo ${tag##*.}'
+gate_sample miss 'cmd 2>&1 >>log'
+gate_sample miss '# mapfile is unavailable on Bash 3'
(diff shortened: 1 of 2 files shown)


scripts/shipped-in.test.sh
```diff
@@ -180,4 +184,38 @@ run_script cloudlands-fe "$sha" --limit 0
...
+echo "shipped-in tests passed under $("$script_bash" -c 'echo "bash $BASH_VERSION"')"
+[[ -z "${SHIPPED_IN_TEST_BASH:-}" ]] || exit 0
+
+# Stock macOS /bin/bash is 3.2 (intent-hq/intent#4706). `bash -n` alone
+# accepts Bash 4+ builtins and expansions, so reject them by pattern too,
+# then rerun the fixtures under a real Bash 3 when one can be found:
+# BASH3_BIN, a bash3 on PATH, Homebrew bash@3, or a 3.x /bin/bash.
+bash -n "$script" || fail "shipped-in.sh does not parse"
+bash -n "${BASH_SOURCE[0]}" || fail "shipped-in.test.sh does not parse"
+bash4_constructs='(declare|local|typeset) +-[A-Za-z]*[An]|mapfile|readarray|\$\{[A-Za-z_][A-Za-z_0-9]*(\[[^]]*\])?(\^\^?|,,?)[^}]*\}|&>>|\|&|;;?&|coproc'
```
panghy: [P3] Treat this gate as a heuristic, or cover its simple spelling variants

The current files are compatible with real Bash 3.2, so this is non-blocking. Direct sample probes show the regex misses an associative-array declaration separated by a tab, split declaration flags such as `declare -r -A cache=()`, and positional case conversion `${1^^}`. Conversely, a comment mentioning `mapfile` or a valid identifier such as `readarray_count=0` fails the gate. On Linux without Bash 3, future code using one of those missed forms could pass the fallback check.

Consider adding table-driven gate samples, allowing horizontal whitespace/split declaration options and using token boundaries for builtins. At minimum document that this is a best-effort guard rather than an exhaustive compatibility check; the real Bash 3 fixture run remains the authoritative check.

panghy (commented):
Non-blocking review: no production correctness or Bash 3.2 compatibility regression found in 505bcb8. The default Bash 5.2 suite, Bash 5.2 to real Bash 3.2.57 re-execution, and direct Bash 3.2.57 suite all pass. Additional offline probes preserved main parity for empty/channel-only listings, usage errors, empty/malformed intentd manifests, all four containment statuses, nonadjacent compare deduplication, and prefix-distinct pinned versions. Detection also correctly rejected a Bash 5 BASH3_BIN and found Bash 3 via PATH and a mocked Homebrew prefix. The new suite fails against the original main script at its three incompatible constructs. All existing fixture assertions remain intact. One optional static-gate hardening note is attached below; no approval or merge action is being submitted.

scripts/shipped-in.test.sh
panghy: Addressed in 79d4062: the pattern now allows horizontal whitespace and split declaration flags (`declare -r -A`), matches positional/special-parameter case conversion (`${1^^}`, `${@,,}`), uses token boundaries for `mapfile`/`readarray`/`coproc` (so `readarray_count=0` passes), and skips full-line comments. The regex is asserted against a table of 18 hit / 14 miss samples on every run, and the comment documents the gate as best-effort with the real Bash 3 fixture run as the authoritative check. All three modes (default, BASH3_BIN re-exec, direct 3.2.57) pass and the gate still rejects origin/main's script.

panghy: Re-verification of 79d4062: all three suite modes pass and 22 additional legitimate-syntax probes pass on Bash 5.2 and 3.2.57. One declaration-matching regression remains: `[An]([^A-Za-z]|$)` requires A/n to be the last option letter, so `declare -Ar cache=()`, `declare -Ax cache=()`, `local -nr ref=x`, `typeset -Anr ref=x`, and `declare -r -Ax cache=()` are now missed even though the previous regex caught them. Please allow trailing alphabetic option letters after A/n and add positive samples for these combined/split flag forms. The production script is unaffected; this is confined to the new gate hardening.

panghy: Fixed in 227eb4a: the declaration alternative now allows trailing option letters after `A`/`n` (`-[A-Za-z]*[An][A-Za-z]*`), so `declare -Ar`, `declare -Ax`, `declare -r -Ax`, `local -nr` and `typeset -Anr` are caught again, with hit samples added for each. All three suite modes pass and the gate still rejects origin/main's script.

panghy: Re-verified 227eb4a: the trailing-option regression is fixed, including combined and split flag forms. All three requested suite modes pass (Bash 5.2 default, nested real Bash 3.2.57, and direct Bash 3.2.57). Independent probes pass for 22 legitimate Bash-3 forms and 10 incompatible forms under both shells. The current suite still rejects the original main script at its three incompatible constructs. No further findings in this scoped review; the thread can remain resolved. This is a comment-only verification report, not a GitHub approval or merge action.

PR #4734, opened 2026-09-11
fix(dev-sandbox): survive repeated or late signals during cleanup and deflake the TERM tests

panghy:
## Summary

Root-causes and fixes the flaky supervised-recipe external-TERM assertion in scripts/dev-sandbox.test.sh (state file 'ui.json' occasionally left behind). The flake was not a slow poll: GNU make blocks until its local child exits when TERMed, so by the time the test's wait returns the script's EXIT trap has already run. Residual state meant cleanup was aborted or skipped. Two signal-handling defects in scripts/dev-sandbox.sh are fixed and each is covered by a committed regression test that fails against the pre-fix script; a third robustness gap in state removal is fixed and was verified with ad-hoc probes.

### scripts/dev-sandbox.sh
- A second HUP/INT/TERM arriving during cleanup used to longjmp out of the EXIT trap and skip the child KILL escalation. cleanup() now masks HUP/INT/TERM as its very first statement. Committed test: the double-TERM block.
- bash checks pending signal traps before the first command of the EXIT trap string, so a TERM landing between a normal exit and cleanup starting skipped cleanup entirely (the residual own-pid ui.json case). The signal traps now run cleanup themselves via on_signal() and then exit 130/143; the EXIT trap remains the normal path and cleanup stays idempotent. Committed test: the 20-iteration signal-window block.
- remove_state_file no longer depends on a python3 helper to read the state file's pid: if that helper failed to start under host load, the own-pid check silently did not match and the state stayed in place. The check is now a bash string match on the own-pid token, and every non-removal of an existing file emits a WARNING (unreadable file, pid-token mismatch with the token seen, failed rm). Verified with ad-hoc probes (python unavailable, rm failure), not a committed test.

### scripts/dev-sandbox.test.sh
- New double-TERM regression block (fake frontend ignores TERM; second TERM sent mid-cleanup) and a 20-iteration exit-path signal-window block (two back-to-back TERMs to the script pid). Against the respective pre-fix scripts they fail (78/120 residual in an ad-hoc harness for the exit-path case); with the fixes they pass. Combined added suite time about 11 s.
- The supervised-recipe assertion waits deterministically (script pid gone, then state file, 10 s deadlines) and dumps the residual JSON, the process group and the script output before failing.
- The 'pgrep -P | head -1' frontend lookup could exit the suite silently under the test's set -euo pipefail on a transient no-match; replaced at all three sites by a bounded, command-matched find_frontend_pid helper, and an errexit-gated ERR trap now names the line and command of any plain-command failure.

## Verification
- bash -n on both scripts; suite passes standalone (about 38 s).
- Loaded stability at final head: 10/10 (implementor) + 8/8 (independent verifier) full-suite runs with dev-ports/dev-status test loops running concurrently, zero failures and zero ERR diagnostics; 20/20 at the commit that finalized dev-sandbox.sh.
- Negative controls: with only dev-sandbox.sh reverted to its parent, the double-TERM block fails; reverted to the pre-exit-path-fix version, the signal-window block fails.
- INT/HUP/TERM still exit 130/143/143 with state removed and the listener closed.

Monorepo-only change (scripts); no release monitoring needed.

### DIFF
--- a/scripts/dev-sandbox.sh
+++ b/scripts/dev-sandbox.sh
@@ -362,9 +362,10 @@ elif [[ "$mode" == stop ]]; then
   exit $?
 fi
 
-# Fork-free so it cannot fail under host load: write_state_file emits compact
-# JSON, so "pid":$$, is a stable own-pid token. Every non-removal of an
-# existing file is reported so a leftover state file is never silent.
+# The own-pid ownership check uses only bash builtins (no python helper), so a
+# missing or failed helper can no longer leave the state in place; write_state_file
+# emits compact JSON, so "pid":$$, is a stable own-pid token. rm remains external,
+# and every non-removal of an existing file is reported so it is never silent.
 remove_state_file() {
   local state_line="" pid_token=""
   [[ -e "$state_file" ]] || return 0
--- a/scripts/dev-sandbox.sh
+++ b/scripts/dev-sandbox.sh
@@ -373,10 +373,10 @@ remove_state_file() {
 }
 
 cleanup() {
-  local pid
   # bash runs pending signal traps between commands even inside the EXIT trap,
   # so a repeated HUP/INT/TERM would otherwise longjmp out mid-cleanup.
   trap '' HUP INT TERM
+  local pid
   [[ "$cleaning" -eq 0 ]] || return
   cleaning=1
   remove_state_file
--- a/scripts/dev-sandbox.sh
+++ b/scripts/dev-sandbox.sh
@@ -362,25 +362,21 @@ elif [[ "$mode" == stop ]]; then
   exit $?
 fi
 
+# Fork-free so it cannot fail under host load: write_state_file emits compact
+# JSON, so "pid":$$, is a stable own-pid token.
 remove_state_file() {
-  python3 - "$state_file" "$$" <<'PY'
-import json
-import os
-import sys
-
-path, expected_pid = sys.argv[1], int(sys.argv[2])
-try:
-    with open(path, encoding="utf-8") as handle:
-        state = json.load(handle)
-    if int(state.get("pid", -1)) == expected_pid:
-        os.unlink(path)
-except (FileNotFoundError, ValueError, TypeError, json.JSONDecodeError):
-    pass
-PY
+  local state_line=""
+  [[ -e "$state_file" ]] || return 0
+  IFS= read -r state_line 2>/dev/null <"$state_file" || true
+  [[ "$state_line" == *"\"pid\":$$,"* ]] || return 0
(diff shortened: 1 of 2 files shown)


panghy (commented):
Reviewed all six commits through 19bff168d4ad7e58558151def7a4ed94ee476ccf against origin/main. No blocking implementation findings. Two documentation findings:

1. **[nit] scripts/dev-sandbox.sh:365-367 — narrow the fork-free claim.** The ownership check is now builtins-only, but remove_state_file still invokes external rm at line 383, and unlink can fail under load or for filesystem reasons. Therefore 'Fork-free so it cannot fail under host load' promises more than the implementation provides (and contradicts the guarded-rm warning immediately below). Suggested wording: 'Avoid the Python JSON-decoding helper during cleanup; check ownership with Bash builtins and diagnose removal failures.' The required removal of the Python dependency is implemented correctly.

2. **[non-blocking] PR description — distinguish verified signal races from the fork-failure hardening.** The description says the previous Python fork ran 'inside a set -e trap' and that its failure 'aborted cleanup'; scripts/dev-sandbox.sh starts with set -u only, unchanged from the base. In the normal invocation a failed Python helper could leave state behind, but errexit would not abort the rest of cleanup. The introduction also says all three reasons have committed regression coverage, whereas the committed new blocks cover the two signal races; the Python-unavailable/rm-failure probes are recorded as ad-hoc verification in the task notes. Please adjust the causal/coverage wording (or cite a specific inherited-errexit reproducer and committed fork-failure test if that stronger claim is intended). No code change is required for this wording issue.

Verification: separate bash -n checks on both files and git diff --check pass; bash scripts/dev-sandbox.test.sh passes in 37.453s; live INT/HUP/TERM probes return 130/143/143 with state absent and the frontend reaped. Extracted ERR-trap probes preserve status 1/7, report pipeline status, and remain silent in expected set +e/conditional paths. Extracted remover probes under set -euo pipefail pass for own/missing/foreign/unreadable state and failed rm with the Python helper unavailable. Audited the prior historical negative-control logs and retained load-run evidence accepted in the spec. PR shell-tests is green.

Mask ordering, idempotence and the retained normal EXIT path look correct. Runtime remains reasonable for this suite; the new polling counts are sleep budgets plus command/scheduling overhead, not strict wall-clock deadlines. Added Bash syntax is compatible with Bash 3.2 by static review; pre-existing setsid, pgrep/process-substitution and array usage are not introduced by this PR. macOS/Bash 3.2 runtime and ShellCheck were unavailable here. No repository files edited.

panghy (commented):
Thanks — PR body corrected: the script runs under set -u only, so the python-helper issue is described as a silent non-match (state left in place), not a set -e abort; and the body now distinguishes the two committed signal regression blocks from the ad-hoc python/rm probes. The remove_state_file comment wording is being fixed in a follow-up commit.

Issue #2355, opened 2026-08-14
Chat: reason-specific stopped indicator labels regressed in workspace redesign (fe#1031); system_suspend has no label


panghy:
## Regression

`feat(chat): reason-specific stopped indicator labels` (intent-hq/cloudlands-fe@bfe18f585, PR intent-hq/cloudlands-fe#757) rendered the persisted `interruptReason` / `interruptedBy` interrupted-row metadata (PROTOCOL §7.2, intent-hq/intentd#919) as reason-specific labels in the inline Stopped indicator:

- "Interrupted by a new message" / "Interrupted by {name}" (`preempted_by_message`)
- "Stopped — daemon restarted" (`daemon_shutdown`)
- "Stopped — agent terminated" (`agent_stopped`)
- generic "Stopped" fallback (`user_stop`, legacy rows, unknown future reasons)

The workspace redesign (intent-hq/cloudlands-fe@df7aa1f12, PR intent-hq/cloudlands-fe#1031) rewrote `ChatMessage.svelte` and reverted the indicator to the generic `m.chat_chatMessage_stopped_label()` ("Stopped") for every interruption. `resolveStoppedIndicatorLabel()` in `message-display-utils.ts` is now dead in production (only its unit tests call it); the i18n keys still exist in all 8 locales. Users currently cannot distinguish why a turn stopped.

## Gap: `system_suspend` has no label

The daemon also emits `interruptReason: system_suspend` (laptop sleep mid-turn; the daemon auto-resumes on wake), but the FE resolver has no case for it — it falls back to generic "Stopped" with no hint that the turn resumes automatically.

## Fix

- Re-wire `resolveStoppedIndicatorLabel()` into `ChatMessage.svelte` (restore the pre-#1031 derived label, including the compile-time exhaustiveness check).
- Add a `system_suspend` case with a new label ("Stopped — system suspended; resumes on wake") across all 8 locales.
- Add a component-level regression test so a future redesign cannot silently drop the wiring again.

panghy:
Fixed in cloudlands-fe v2.38.0 (alpha), bundles intentd v0.6.16.

PR #4786, opened 2026-09-12
docs(fe): document the chat-motion perf trace harness

panghy:
## Summary

Adds a 30-line Chat-motion performance traces subsection to docs/fe/DEVELOPER_GUIDE.md describing the reusable harness added in intent-hq/cloudlands-fe#2375: the pnpm perf:chat-motion command, its flags and defaults, exit codes (0 success, 1 runtime failure, 2 usage / refused overwrite), the artifact list, the summary.json fields (motions per mark, mark-derived scroll labels, both quiescence waits), why the quiescence waits and the no-toggle control exist (the #2355 hydration false positives), and the recommended layout-thrash negative control.

The cloudlands-fe AGENTS.md Where to look row in that PR links to this subsection. This PR touches docs only; the packages/cloudlands-fe pointer is not changed (the auto-bump workflow advances it after the submodule PR merges).

## Verification

- make docs-check and the doc-link tests / link checker pass.
- Contents cross-checked against scripts/perf at cloudlands-fe feat/chat-motion-perf-harness HEAD 1f66e3f5.


panghy (commented):
## Independent documentation review at 6cb5611d

### Should-fix: document the fixture requirements that make the command runnable
Location: docs/fe/DEVELOPER_GUIDE.md:141-149.

The prerequisites currently only require a workspace with a long scrollable transcript. The default footer scenario actually waits for an event-subscriptions footer header, which need not exist in such a workspace. The task evidence already records dry-python timing out for exactly this reason. Default inflation also requires a rendered donor row with more than 20 descendant elements and height below 2000px; a transcript represented as one very tall row is rejected. The spec Results calls out both limitations, but a new agent following the durable guide cannot discover them before trying the command.

Please mention the event-subscriptions footer prerequisite for the footer scenario and the bounded-row requirement for inflation; document --inflate 0 as the way to disable synthetic inflation when the real transcript is already suitable. This can replace/condense the existing prerequisite prose rather than growing the subsection budget.

No other docs findings: CLI defaults, exit codes, scenario artifact filenames, summary fields and both quiescence waits match implementation. make docs-check, node scripts/check-doc-links.mjs and both git diff --check commands pass (exit 0). No source edits or commits.

The blocking measurement synchronization finding belongs to the companion code PR: https://github.com/intent-hq/cloudlands-fe/pull/2375.

panghy (commented):
## Re-review: prerequisite finding resolved

Verified pushed HEAD 7fd594f081adb014b7705f1f39c52981a9ba560d. The guide now names the event-subscriptions footer / context-well requirements, both donor restrictions (more than 20 descendants and under 2000 px), and the --inflate 0 escape hatch. These agree with the runner. The only changes since my review are the scoped prerequisite additions and paragraph compression.

make docs-check, node scripts/check-doc-links.mjs, and git diff --check origin/main...HEAD all pass (exit 0). No remaining blocking or should-fix documentation findings. My earlier finding was a general review comment, not an inline thread, so this review records its resolution.

Documentation review verdict: APPROVED. Companion frontend measurement fix remains under review; this is not merge authorization.

PR #4810, opened 2026-09-12
docs(protocol): one PR monitor per PR per workspace (§5.42)

panghy:
## Summary

Document the one-active-monitor-per-PR-per-workspace rule for ws.pr.monitor in docs/protocol/methods/pr.md section 5.42:

- the refusal payload { ok: false, refused: true, reason: "already-monitored", repo, prNumber, ownerAgentId, ownerAgentName?, monitorId, instruction }, verified field-for-field against the intentd implementation;
- own re-register is idempotent; the owner's ws.pr.unmonitor frees the PR; coordination via ws.agent.send (ask the owner to relay events or relinquish);
- repo identity is case-insensitive (GitHub slug semantics);
- the already-monitored precheck refuses before any forge fetch; the rare insert-race loser is refused after its fetch with the same payload.

Companion daemon change: intent-hq/intentd#1802. Docs only; the packages/intentd gitlink is not moved (auto-bump-submodules owns pin advancement).


panghy (commented):
## Review: protocol docs §5.42 — one PR monitor per PR per workspace

Compared `docs/protocol/methods/pr.md` §5.42 (commits `49eb3ca`, `ef63ec8` on `reduce-notification`) field-for-field against the intentd implementation (intentd PR #1802, `e03de08c..HEAD`).

Checked each claim against code:
- "at most one ACTIVE monitor per `(repo, prNumber)` ... enforced at the store by a partial unique index over active rows (`idx_pr_monitor_workspace_identity`, migration `0118`)" — matches `0118_pr_monitor_workspace_identity.sql`.
- "Repo identity is case-insensitive ... for both this workspace rule and the per-agent re-arm identity ... COLLATE NOCASE (migration 0119)" — matches `0119_pr_monitor_identity_nocase.sql`, which updates both `idx_pr_monitor_identity` and `idx_pr_monitor_workspace_identity`.
- "the stored `repo` label keeps the casing supplied by the first registration" — matches `PrMonitorRefusal::to_wire()`, which reports `self.owner.repo_owner`/`repo_name` (the stored row's casing), not the caller's override casing.
- Refusal payload shape `{ ok: false, refused: true, reason: "already-monitored", repo, prNumber, ownerAgentId, ownerAgentName?, monitorId, instruction }` — matches `PrMonitorRefusal::to_wire()` exactly, including `ownerAgentName` being omitted (not `null`) when the owner session is unnamed, and `monitorId` being the owner's monitor id rather than a full `PrMonitor` row (a documented spec decision).
- "refuses before the forge fetch and before the per-agent cap check" — matches `pr_monitor_try_register`'s ordering: `pr_monitor_refusal` check precedes the cap check, which precedes `fetch_snapshot`.
- "the one exception is the insert-race loser ... which has already fetched by the time the store's unique index rejects its row" — matches: the fetch happens before `insert_pr_monitor`, and the second `pr_monitor_refusal` check (post-insert-failure) happens after that fetch. This correctly qualifies what an earlier verification pass flagged as an overstatement in an older revision.
- "the owner's own re-register is never refused ... returns `{ ok: true, monitor, requirements }`" — matches (`owner.agent_id == *agent_id` short-circuits to no refusal).
- "freed once the owner's monitor leaves the active state ... owner's `ws.pr.unmonitor`, an FE `prMonitor.cancel`, or completion" — matches unchanged cancel/completion code paths (untouched by this PR, correctly scoped as a non-goal).
- "Cross-workspace uniqueness is not enforced" and "no daemon-side relay/transfer primitive ... `ws.agent.send` ... `ws.pr.unmonitor`" — matches the spec's non-goals and the refusal `instruction` text in `pr_monitor.rs`.

No discrepancies found between the doc and the implementation. The wording is precise about the prefetch-vs-insert-race distinction, which a prior review round had flagged as under-qualified — that gap is resolved here.

### Verdict
Matches the implementation field-for-field. Approving.

PR #4866, opened 2026-09-12
feat: add make test-changed for changed-only Rust nextest runs

panghy:
## Summary

Adds a supported pre-merge-queue gate that runs only the nextest targets a branch actually touched, so a targeted local run replaces "skip make test because the host is loaded".

- **scripts/rust-changed-tests.sh** — diffs the packages/intentd checkout against origin/main (merge-base, plus uncommitted and untracked files, read NUL-delimited) and maps:
  - crates/C/tests/T.rs → -p C --test T
  - crates/C/tests/common/** or nested test dirs, deleted test files → -p C --tests
  - crates/C/src/** → -p C --lib --bins --tests (--lib omitted for bin-only crates; prints a note that downstream crates are not selected)
  - other files under crates/C/ (fixtures, migrations) → -p C
  - benches/examples and paths outside crates/ are ignored
  - Exit 3 on build-wide changes (Cargo.toml, Cargo.lock, crate Cargo.toml, build.rs, .config/, .cargo/, rust-toolchain.toml); exit 2 on an unresolvable BASE; exit 0 with "nothing to test" on an empty selection. DRY_RUN=1 prints the plan only.
- **make test-changed** — same nextest install check, TEST_THREADS / BUILD_JOBS caps and progress-bar env as make test; turns script exit 3 into the full make test.
- **scripts/rust-changed-tests.test.sh** — hermetic suite (throwaway git repo, stub cargo) covering every mapping rule, the fallback/exit codes, quoted paths (spaces, Unicode), renames, and DRY_RUN; wired into the shell-tests CI job.
- One bullet in the root AGENTS.md "Resuming local Rust gates" section.

Motivation: intent-hq/intentd#1825 was ejected from the merge queue after the pre-queue full make test was skipped for host load; the two failing binaries run in under a minute in isolation. A follow-up PR on intent-hq/intentd will mention the target in packages/intentd/AGENTS.md.

## Verification

- bash scripts/rust-changed-tests.test.sh, scripts/shipped-in.test.sh, scripts/docs-check.test.sh: pass
- make docs-check, make help: pass / target listed
- make test-changed DRY_RUN=1: "nothing to test" on a clean tree; exact plan -p intentd --test e2e_wss_repo_warm_cache with a probe test-file edit; fallback announced with a Cargo.lock probe; exit 2 with BASE=nonexistent-ref
- Real make test-changed with the probe edit: build 2m13s, 4/4 tests in exactly one binary (intentd::e2e_wss_repo_warm_cache), exit 0
- Not covered: reverse-dependency propagation (by design; make test remains the complete gate); Bash 3.2 verified by static inspection only (no Bash 3 on the host)

### DIFF
--- a/scripts/rust-changed-tests.sh
+++ b/scripts/rust-changed-tests.sh
@@ -10,8 +10,8 @@
 # are inherited as-is (the Makefile sets them).
 #
 # The changed set is `git diff --name-only $(git merge-base HEAD BASE)` inside
-# INTENTD_DIR (committed, staged and unstaged edits) plus the untracked paths
-# from `git status --porcelain`. Paths map to per-crate nextest selections:
+# INTENTD_DIR (committed, staged and unstaged edits) plus the untracked files
+# from `git ls-files --others`. Paths map to per-crate nextest selections:
 #   crates/<c>/tests/<t>.rs          -p <c> --test <t>   (--tests once deleted)
 #   crates/<c>/tests/<dir>/**        -p <c> --tests
 #   crates/<c>/src/**                -p <c> --lib --bins --tests
@@ -88,18 +88,6 @@ git rev-parse --git-dir >/dev/null 2>&1 || die 2 "$display_dir is not a git chec
 merge_base=$(git merge-base HEAD "$base" 2>/dev/null) ||
   die 2 "cannot resolve BASE '$base' in $display_dir; run 'git -C $display_dir fetch origin main' or set BASE=<ref>"
 
-# Committed + staged + unstaged edits since the merge base, then untracked
-# paths (`?? <path>`; renames are reported as `<old> -> <new>`).
-changed=$(
-  git diff --name-only "$merge_base" --
-  git status --porcelain --untracked-files=all | while IFS= read -r line; do
-    [[ "$line" == '??'* ]] || continue
-    path=${line:3}
-    printf '%s\n' "${path##* -> }"
-  done
-)
-changed=$(printf '%s\n' "$changed" | sort -u)
-
 is_fallback() {
   case "$1" in
     Cargo.toml | Cargo.lock | rust-toolchain.toml | .config/nextest.toml | .cargo/*) return 0 ;;
@@ -133,14 +121,27 @@ map_path() {
 
 fallback=""
 mapped=""
-while IFS= read -r path; do
-  [[ -n "$path" ]] || continue
+seen=""
+classify() {
+  local path=$1
+  [[ -n "$path" && "$seen" != *$'\n'"$path"$'\n'* ]] || return 0
+  seen+=$'\n'"$path"$'\n'
   if is_fallback "$path"; then
     fallback+="$path"$'\n'
   else
     mapped+="$(map_path "$path")"$'\n'
   fi
-done <<<"$changed"
+}
+
+# Committed + staged + unstaged edits since the merge base, then untracked
+# files. Both are read NUL-delimited so paths arrive raw: line-oriented
+# porcelain output C-quotes names with spaces or non-ASCII characters.
+while IFS= read -r -d '' path; do
+  classify "$path"
+done < <(git diff --name-only -z "$merge_base" --)
+while IFS= read -r -d '' path; do
+  classify "$path"
+done < <(git ls-files -z --others --exclude-standard)
 
 if [[ -n "$fallback" ]]; then
   echo "[test-changed] build-wide change(s) vs $base need the full suite -- run 'make test':" >&2
--- a/scripts/rust-changed-tests.sh
+++ b/scripts/rust-changed-tests.sh
@@ -1,0 +1,237 @@
+#!/usr/bin/env bash
+# Run only the nextest targets a packages/intentd branch touched.
+#
+#   scripts/rust-changed-tests.sh [--dry-run] [--base REF] [--intentd-dir DIR]
+#                                 [--build-jobs N] [--test-threads N]
+#
+# Each flag falls back to an environment variable: DRY_RUN=1, BASE (default
+# origin/main), INTENTD_DIR (default packages/intentd next to this script),
+# BUILD_JOBS, TEST_THREADS. NEXTEST_SHOW_PROGRESS / CARGO_TERM_PROGRESS_WHEN
+# are inherited as-is (the Makefile sets them).
+#
+# The changed set is `git diff --name-only $(git merge-base HEAD BASE)` inside
+# INTENTD_DIR (committed, staged and unstaged edits) plus the untracked paths
+# from `git status --porcelain`. Paths map to per-crate nextest selections:
+#   crates/<c>/tests/<t>.rs          -p <c> --test <t>   (--tests once deleted)
+#   crates/<c>/tests/<dir>/**        -p <c> --tests
+#   crates/<c>/src/**                -p <c> --lib --bins --tests
+#                                    (--lib only when src/lib.rs exists)
+#   crates/<c>/benches|examples/**   ignored (nextest does not run them)
+#   crates/<c>/<anything else>       -p <c>  (all test targets of <c>)
+#   outside crates/                  ignored
+# A broader selection subsumes narrower ones for the same crate. Cargo applies
+# target flags to every -p package on one command line, so crates with
+# different selections run as separate `cargo nextest run` invocations and
+# crates with identical selections share one. Reverse dependencies are NOT
+# propagated: crates/<c>/src selects <c>'s own tests only; `make test` stays
+# the complete gate.
+#
+# Exit codes: 0 = nothing to test, or every invocation passed; 2 = usage error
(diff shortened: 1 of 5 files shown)


scripts/rust-changed-tests.sh
```diff
@@ -0,0 +1,238 @@
...
+    mapped+="$(map_path "$path")"$'\n'
+  fi
+}
+
+# Committed + staged + unstaged edits since the merge base, then untracked
+# files. Both are read NUL-delimited so paths arrive raw: line-oriented
+# porcelain output C-quotes names with spaces or non-ASCII characters.
+while IFS= read -r -d '' path; do
+  classify "$path"
+done < <(git diff --name-only -z "$merge_base" --)
```
panghy: [P1] Preserve the source side when a rename crosses selection boundaries

Git name-only rename output contains only the destination. In an isolated committed fixture, git mv Cargo.lock Cargo.lock.backup makes this script return 0 and print nothing to test, rather than the required build-wide fallback exit 3. Renaming crates/alpha/build.rs to crates/alpha/retired.rs likewise runs only -p alpha instead of falling back. Moving crates/alpha/src/old.rs to crates/alpha/tests/old.rs selects only --test old, losing the source-change --lib --bins --tests requirement; moving a test between crates omits the originating crate entirely.

Please classify both sides of cross-boundary renames (or conservatively disable rename detection so they are treated as deletion plus addition). Add regressions for a renamed build-wide path, src-to-tests move, and cross-crate move, and rerun the suite. The current same-directory test rename fixture only exercises the safe destination-only case.

scripts/rust-changed-tests.sh
```diff
@@ -0,0 +1,238 @@
...
+while IFS= read -r -d '' path; do
+  classify "$path"
+done < <(git ls-files -z --others --exclude-standard)
```
panghy: [P2] Fail closed when either Git path enumeration fails

The two git commands are process substitutions, so their exit statuses are not the while loops' exit statuses; set -euo pipefail does not propagate their failures. Reproduced with a tracked integration-test edit and local git config diff.renames not-a-valid-value: git diff prints a fatal error, but the script then exits 0 with nothing to test. The Makefile consequently treats an uncomputed selection as a successful gate. A failure from git ls-files is subject to the same problem and can hide untracked tests.

Please explicitly check both producers before accepting the plan, retaining NUL-safe path handling and Bash 3.2 compatibility. Add failure-injection tests for git diff and git ls-files; each should exit nonzero, not print a success/nothing-to-test message, and not invoke cargo. Keep status 3 reserved for a known full-suite fallback.

panghy (commented):
Reviewed commits 91fb6b0, 965ec24, and 62c5f26 against the workspace acceptance criteria. Two correctness findings need fixes before this is a trustworthy pre-queue gate (inline): rename sources are dropped, and Git enumeration failures produce a successful empty plan. Posting a COMMENT review as requested, not an approval.

Validation passed: bash scripts/rust-changed-tests.test.sh; bash scripts/docs-check.test.sh; bash scripts/shipped-in.test.sh; make docs-check; make help; make test-changed DRY_RUN=1; bash -n; git diff --check. Independent temporary-repo tests confirmed exact warm-cache/source target argv, checkout paths containing spaces, Makefile progress env/caps, ordinary Cargo.lock exit-3 fallback dispatch, DRY_RUN fallback safety, and invalid BASE exit 2. No source edits or real full-suite run were made. Bash 3 is unavailable here; static compatibility checks passed on Bash 5.2. GitHub currently reports no PR checks, so CI execution is not verified.

Non-blocking test-isolation note: scripts/rust-changed-tests.test.sh:104-109 inherits the very variables its comment says are isolated. BASE=HEAD bash scripts/rust-changed-tests.test.sh fails the clean-tree expectation, and DRY_RUN=1 bash scripts/rust-changed-tests.test.sh fails the first cargo expectation. Unset the supported script input variables once at suite startup, while retaining explicit per-case overrides, and regression-test those ambient-env invocations. Restricted PATH, cargo stubbing, and temporary cleanup otherwise look sound.

scripts/rust-changed-tests.sh
panghy: Fixed in 842e981 (+ test in dde3824): the diff now runs with `--no-renames`, so both sides of a move are changed paths and the source keeps its classification. Hermetic cases: `git mv Cargo.lock Cargo.lock.backup` and `git mv crates/beta/build.rs crates/beta/retired.rs` -> exit 3 naming the source; `git mv crates/alpha/src/util/mod.rs crates/alpha/tests/old.rs` -> `-p alpha --lib --bins --tests`; a moved integration test is delete + add -> `-p alpha --tests`.

scripts/rust-changed-tests.sh
panghy: Fixed in 842e981: `git diff --name-only -z --no-renames` and `git ls-files -z --others --exclude-standard` are written to a `mktemp` file with each exit status checked; a failure is exit 2 with `[test-changed] git <cmd> ... failed in <dir> (exit N)` after git's own fatal line, never "nothing to test". Paths stay NUL-delimited / Bash 3.2. The suite has a `git` wrapper honouring `GIT_STUB_FAIL=<subcommand>` with cases for `diff` and `ls-files`; the `diff.renames not-a-valid-value` repro on the real checkout now exits 2.

scripts/rust-changed-tests.sh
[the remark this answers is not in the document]
panghy: Fixed in 842e981 (+ test in dde3824): the diff now runs with `--no-renames`, so both sides of a move are changed paths and the source keeps its classification. Hermetic cases: `git mv Cargo.lock Cargo.lock.backup` and `git mv crates/beta/build.rs crates/beta/retired.rs` -> exit 3 naming the source; `git mv crates/alpha/src/util/mod.rs crates/alpha/tests/old.rs` -> `-p alpha --lib --bins --tests`; a moved integration test is delete + add -> `-p alpha --tests`.

panghy: Independently re-verified the fix in 842e981c7f6d74f53c313165afe5064e150d5d75. Disabling rename detection preserves both sides. Original temporary-repo repros now pass: Cargo.lock -> Cargo.lock.backup and build.rs -> retired.rs exit 3 with the original path named; src/old.rs -> tests/old.rs selects -p alpha --lib --bins --tests; moving an integration test across crates selects old-crate --tests plus the new-crate named test. The suite passes at the fix commit and at dde3824e8d991f7c66ba2f4bebb7e2e829673bfa (additional build.rs regression). Resolving this finding; no PR approval submitted.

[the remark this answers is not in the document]
panghy: Fixed by 842e981c7f6d74f53c313165afe5064e150d5d75 using --no-renames. Independent temporary-repo repros matching this comment now pass: moving a source file into benches or outside crates retains -p alpha --lib --bins --tests, and moving .config/nextest.toml outside its protected path exits 3 and names the original config. Cargo.lock/build.rs renames and cross-crate moves also pass. Bash suite/docs checks pass; resolving this duplicate rename finding without approving the PR.

scripts/rust-changed-tests.sh
panghy: Independently re-verified 842e981c7f6d74f53c313165afe5064e150d5d75 with the original real Git repro (tracked test edit plus invalid local diff.renames): now exit 2, original fatal diagnostic retained, and no success plan. Independent injected failures from git diff and git ls-files likewise exit 2 with empty stdout. The committed hermetic suite asserts no cargo call for both failures. Also checked that the temporary path-list file is removed on success, fallback, and both failures, including TMPDIR containing spaces. Resolving; no PR approval submitted.

scripts/rust-changed-tests.sh
```diff
@@ -0,0 +1,244 @@
...
+    crates/*/Cargo.toml | crates/*/build.rs) return 0 ;;
+  esac
+  return 1
+}
+
+# Prints "<crate>\t<kind>[\t<test>]" for a path under crates/, kinds ranked
+# test < tests < src < all; prints nothing for ignored paths.
+map_path() {
+  local path=$1 rest crate sub name
+  [[ "$path" == crates/*/* ]] || return 0
```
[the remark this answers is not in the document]
panghy: Fixed in f777e2b: outside `crates/` the script now fails closed. Only inert paths are ignored (`*.md`, `LICENSE`, `NOTICE`, `.gitignore`, `.github/**`, `docs/**`, `deny.toml`, `release-plz.toml`, `dist-workspace.toml`); any other path outside `crates/` (`scripts/**`, `packaging/**`, unknown root files) triggers the full-suite fallback (exit 3) with a reason line naming the path, alongside the existing build-wide triggers. Hermetic cases: `scripts/install.sh`, `scripts/tool.sh`, `packaging/deb/control`, `unknown.toml` each combined with a test-file edit -> exit 3 naming only the non-crate path; `README.md`, `docs/guide.md`, `.github/workflows/x.yml` and the other inert files -> still "nothing to test".

PR #4931, opened 2026-09-13
fix: run Rust make gates under the pinned rustup toolchain

panghy:
## Summary

On hosts where a non-rustup cargo (e.g. Homebrew's rust formula) is first on PATH, the Rust make gates ran that toolchain instead of the one pinned by packages/intentd/rust-toolchain.toml, because rustup's proxies are keg-only on Homebrew-rustup hosts and never live in CARGO_BIN_DIR — making the existing PATH prepend a no-op. Gate results then diverge from CI (observed: clippy/fmt under 1.98.0 vs pinned 1.96.0, workspace round-ram / intent-hq/intentd#1853).

Two changes:
- Makefile: resolve the pinned toolchain's cargo via rustup which (with RUSTUP_AUTO_INSTALL=0) and prepend its directory ahead of CARGO_BIN_DIR, so every Rust recipe — fmt, clippy, lint-repo-slug, test (incl. resumable_nextest.py), list-tests, coverage — runs the pin. Graceful fallback when rustup or the toolchain is absent: behavior unchanged, no new hard requirement.
- scripts/bootstrap-dev-host.sh: make doctor now emits a non-failing [warn] when the caller-visible plain cargo does not run the pinned toolchain version, naming the shadowing binary, both versions, and the remediation. The caller PATH is captured before the script's own prepend, and the Makefile-injected prefix is stripped when invoked through make (MAKELEVEL). Covered by new cases in scripts/bootstrap-dev-host.test.sh.

## Verification (on a reproducing host: Homebrew cargo 1.98.0 first on PATH, no ~/.cargo/bin/cargo)
- Probe recipe through the Makefile resolves ~/.rustup/toolchains/1.96.0-aarch64-apple-darwin/bin/cargo (1.96.0).
- make fmt, full make clippy, and make list-tests ARGS="-p intent-core" pass under the pin; cargo-nextest still discoverable.
- Rustup-absent PATH: gates fall back to the available cargo unchanged.
- make doctor prints the [warn] for /opt/homebrew/bin/cargo 1.98.0 vs pinned 1.96.0 with exit code and gap count unchanged; silent when a pin-honoring cargo leads the caller PATH (direct and through make).
- bash scripts/bootstrap-dev-host.test.sh passes (shadowed, silent, gap-count-invariance, MAKELEVEL-strip cases).

Companion docs PR: intent-hq/intentd#1858 (raw cargo gate invocations must use the rustup-pinned toolchain).


scripts/bootstrap-dev-host.sh
```diff
@@ -9,6 +9,20 @@ FE_DIR=${FE_DIR:-"$ROOT_DIR/packages/cloudlands-fe"}
...
 CARGO_HOME=${CARGO_HOME:-"$HOME/.cargo"}
+# Caller-visible cargo, resolved BEFORE this script's own PATH prepend below.
+# Under make (MAKELEVEL set) the Makefile has already prepended the pinned
+# rustup toolchain dir and CARGO_BIN_DIR (see the PATH export in the Makefile);
+# strip that exact prefix so the shadowing-cargo check still sees what a plain
+# `cargo` resolves to in the caller's shell.
+CALLER_PATH="$PATH"
+if [[ -n ${MAKELEVEL:-} ]]; then
+  make_rustup_cargo=$(cd "${INTENTD_DIR}" 2>/dev/null && RUSTUP_AUTO_INSTALL=0 rustup which cargo 2>/dev/null)
+  make_prefix="${make_rustup_cargo:+${make_rustup_cargo%cargo}:}${CARGO_INSTALL_ROOT:-$CARGO_HOME}/bin:"
```
panghy: nit: this reconstruction assumes the Makefile prepended `${CARGO_INSTALL_ROOT:-$CARGO_HOME}/bin`, but the Makefile actually uses `$(CARGO_BIN_DIR)`, which is `?=`-overridable. If a caller overrides `CARGO_BIN_DIR` (env var or `make CARGO_BIN_DIR=...`), the strip never matches, `CALLER_CARGO` resolves to the make-injected pinned cargo, and the check goes silent — a false negative, so it fails safe, but silently. Minimal fix for the env-var case: `${CARGO_BIN_DIR:-${CARGO_INSTALL_ROOT:-$CARGO_HOME}/bin}` (make exports it into the recipe environment via the PATH export? it does not — `CARGO_BIN_DIR` itself is not exported, so this only covers a caller-exported override). If that is not worth handling, a one-line comment documenting the assumption would do.

scripts/bootstrap-dev-host.sh
```diff
@@ -112,6 +130,21 @@ active_toolchain_ready() {
...
+# PATH (e.g. Homebrew's) ignores the rust-toolchain.toml pin. Advisory only —
+# never counts toward FAILURES. Absence of any cargo stays covered by the
+# existing checks.
+report_shadowing_cargo() {
+  [[ -n "$TOOLCHAIN" && -n "$CALLER_CARGO" ]] || return 0
+  local line version
+  line=$( (cd "$INTENTD_DIR" 2>/dev/null && PATH="$CALLER_PATH" RUSTUP_AUTO_INSTALL=0 "$CALLER_CARGO" --version 2>/dev/null) | head -n 1 )
+  version=${line#cargo }
+  version=${version%% *}
+  [[ "$version" == "$TOOLCHAIN" ]] && return 0
```
panghy, line 153: nit: this equality assumes `channel` is an exact `X.Y.Z` pin. If the pin ever moved to a named channel (`stable`) or a partial version (`1.96`), `cargo --version` output would never equal it and the warn would fire permanently — including for a rustup proxy that honors the pin, the exact "good" case. The repo pins exact versions today, so non-blocking; if you want it future-proof, only run the comparison when `[[ "$TOOLCHAIN" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]`, and skip (or resolve via `rustup run "$TOOLCHAIN" cargo --version`) otherwise.

panghy (commented):
Reviewed the diff against the spec (Makefile probe correctness, prepend ordering, graceful degradation, caller-PATH capture + MAKELEVEL strip, test coverage, portability/quoting, commit hygiene).

Verified:
- `RUSTUP_CARGO` probe: `rustup which cargo` runs from `$(INTENTD_DIR)` so `rust-toolchain.toml` applies, `RUSTUP_AUTO_INSTALL=0` prevents a download, and an absent rustup/toolchain expands empty — the `$(if ...)` keeps the PATH export byte-identical to the previous `$(CARGO_BIN_DIR):$(PATH)` shape in that case. Toolchain dir is prepended ahead of `CARGO_BIN_DIR`, so the pin wins while `cargo-nextest` stays discoverable. FE targets only gain an inert PATH entry.
- Strip prefix match: `${make_rustup_cargo%cargo}:` keeps the trailing slash, matching make's `$(dir ...)`, and `${CARGO_INSTALL_ROOT:-$CARGO_HOME}/bin` matches `$(or ...)` semantics (both treat empty as unset; the script defaults `CARGO_HOME` to `$HOME/.cargo` first). The `while` loop handles nested-make double prepends and always terminates. The script is `set -u` only, so the unguarded `make_rustup_cargo=$(...)` assignment is safe when rustup is absent.
- `report_shadowing_cargo` never touches `FAILURES`, returns 0 on all paths, and probes with `RUSTUP_AUTO_INSTALL=0` so doctor cannot trigger a toolchain download. Quoting is space-safe throughout (`"$make_prefix"*`, `${CALLER_PATH#"$make_prefix"}` with quoted pattern).
- Tests: all four requested cases present (shadowed warn, silent match, gap-count invariance, MAKELEVEL strip + no-MAKELEVEL control), and `run_doctor` now pins `CARGO_HOME`/`CARGO_INSTALL_ROOT`/`MAKELEVEL` for hermeticity. `bash scripts/bootstrap-dev-host.test.sh` passes locally on this host; `shell-tests`, `shell-tests-bash3`, and `bootstrap-install` are green.
- Commit hygiene: multi-commit PR, title `fix: ...` is a valid conventional commit (it is what lands as the squash title); both branch commits are conventional; no breaking-change footer token anywhere (dedicated CI check also green).

Two non-blocking nits inline. One informational note, no change requested: `$(dir $(RUSTUP_CARGO))` word-splits on spaces, so a HOME containing spaces would mangle the prepend — consistent with the Makefile's existing `CARGO_BIN_DIR` assumption and it degrades to pre-fix behavior.

scripts/bootstrap-dev-host.sh
panghy: Fixed in a29ffec: the strip now uses `${CARGO_BIN_DIR:-${CARGO_INSTALL_ROOT:-$CARGO_HOME}/bin}`, so a caller-exported `CARGO_BIN_DIR` is honored and the prefix match no longer false-negatives into a silent check. As you note, a `make CARGO_BIN_DIR=...` command-line override is not exported into the recipe environment, so it stays invisible — the comment above the strip now documents that limit. Covered by a new test case in scripts/bootstrap-dev-host.test.sh (MAKELEVEL set + exported `CARGO_BIN_DIR` pointing at a custom bin dir: the warn still fires on the caller's shadowing cargo).

scripts/bootstrap-dev-host.sh
panghy: Fixed in a29ffec: `report_shadowing_cargo` now guards the comparison with `[[ "$TOOLCHAIN" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]` and returns silently for a named (`stable`) or partial (`1.96`) channel, so a pin-honoring rustup proxy can never draw a permanent warn under a non-exact pin. New test cases in scripts/bootstrap-dev-host.test.sh assert no warn for `stable` and `1.96` channels and that the exact-pin warn still fires for `1.96.0`.

Makefile
```diff
@@ -37,7 +37,15 @@ FE_DEPS_INSTALL_MSG = installing/refreshing FE deps (lockfile changed or node_mo
...
 CARGO_BIN_DIR ?= $(or $(CARGO_INSTALL_ROOT),$(CARGO_HOME),$(HOME)/.cargo)/bin
-export PATH := $(CARGO_BIN_DIR):$(PATH)
+# Route every Rust recipe through the rustup toolchain pinned by
+# packages/intentd/rust-toolchain.toml, even when another cargo (Homebrew,
+# distro) is first on the caller's PATH. `rustup which` resolves the pin from
+# $(INTENTD_DIR) without triggering a toolchain download (RUSTUP_AUTO_INSTALL=0
+# guards rustup >= 1.28 auto-install). When rustup or the toolchain is absent
+# this expands to empty and behavior is unchanged. cargo-nextest and other
+# installed subcommands stay discoverable via $(CARGO_BIN_DIR).
+RUSTUP_CARGO := $(shell cd $(INTENTD_DIR) 2>/dev/null && RUSTUP_AUTO_INSTALL=0 rustup which cargo 2>/dev/null)
```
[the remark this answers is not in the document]
panghy: Same issue as the sibling thread on this line — fixed in d350f53.

`RUSTUP_CARGO` is now deferred (recursive `=`) with `ORIG_PATH := $(PATH)` snapshotted so the recursive `export PATH` cannot self-reference, and the probe is gated on a readable `rust-toolchain.toml`. The gate matters for the case this comment describes: on a fresh clone `packages/intentd` exists but is empty, so `rustup which cargo` does not fail — it answers with the caller's default toolchain, which would have been memoized as if it were the pin.

Verified on GNU Make 3.81 and 4.4.1 that a probe makefile with `INTENTD_DIR` pointed at an empty dir, populated by a prerequisite, has the pinned `1.96.0` dir at the head of the consuming recipe's PATH — so the first `make fmt/clippy/check` of a clone is pinned. Full detail, including the overhead measurement and the `$(eval)` memo I dropped, is in the sibling thread.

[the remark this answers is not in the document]
panghy: Valid — fixed in d350f53.

`RUSTUP_CARGO` is now deferred (recursive `=`) instead of `:=`, with `ORIG_PATH := $(PATH)` snapshotted first so the recursive `export PATH` cannot reference itself. Make rebuilds a recipe's exported environment per recipe, so the probe runs after `ensure-intentd-submodule` has had its turn.

Simulating the fresh-clone ordering surfaced a second failure mode worth noting: a fresh clone has `packages/intentd` present but *empty*, so `cd $(INTENTD_DIR) && rustup which cargo` succeeds and answers with the caller's **default** toolchain rather than failing. Deferring alone therefore still produced an unpinned PATH (`stable-aarch64-apple-darwin/bin/`, cargo 1.98.0). The probe is now gated on a readable pin file, which is both the correct precondition for `rustup which` and the signal that the submodule is populated:

```make
ORIG_PATH := $(PATH)
RUSTUP_CARGO = $(shell cd $(INTENTD_DIR) 2>/dev/null && [ -r rust-toolchain.toml ] && RUSTUP_AUTO_INSTALL=0 rustup which cargo 2>/dev/null)
export PATH = $(addsuffix :,$(dir $(RUSTUP_CARGO)))$(CARGO_BIN_DIR):$(ORIG_PATH)
```

Verification (GNU Make 3.81, the macOS system make, and a locally built 4.4.1, which exports variables into `$(shell)` and so exercises the recursion path):

- **Fresh-clone ordering** — probe makefile that `include`s this Makefile with `INTENTD_DIR` pointed at an empty dir and a prerequisite that drops `rust-toolchain.toml` in just before the consuming recipe. Before: PATH head `…/stable-…/bin/`. After: PATH head `…/1.96.0-aarch64-apple-darwin/bin/` on both make versions — pinned by recipe time.
- **Pin absent entirely** — no rustup dir is prepended, PATH head is `$(CARGO_BIN_DIR)`; unchanged fallback behavior.
- **Populated tree** — recipe PATH is byte-identical in shape to the old `:=` version (pinned dir, `$(CARGO_BIN_DIR)`, caller PATH) on both make versions; no recursion error under 4.4.1.
- **Overhead** — `$(addsuffix)` keeps the export to a single `$(RUSTUP_CARGO)` expansion, so it is one `rustup which cargo` per recipe environment, measured at 0.011s. `make ensure-intentd-submodule` timed 3x on old vs new code is 0.02s either way, i.e. within noise. An `$(eval)` memo was tried and deliberately dropped: an assignment made while make builds a recipe's environment does not outlive that expansion, so it added indirection for no saving.
- **Gates** — `make fmt` passes against the pinned toolchain; `bash scripts/bootstrap-dev-host.test.sh` passes; `make doctor` output is byte-identical to pre-change (4 gaps, exit 2, shadowed-cargo warn still printed).

One aside found while verifying, pre-existing and unchanged by this commit: under make 3.81 a recipe line that is a bare single command with no shell metacharacters (`cargo --version`) is exec'd on a fast path that does not honor the recipe's exported PATH, so it resolves the caller's cargo. I confirmed it behaves identically with the old `:=` code, and every Rust recipe here is `cd $(INTENTD_DIR) && cargo …`, which goes through a shell and resolves the pinned toolchain on both make versions.

Not covered by an automated test: the script suite in `scripts/*.test.sh` has no Makefile coverage, and CI enumerates those files explicitly, so a guard for this would mean a new test file plus a `ci.yml` edit. Raising that as a follow-up rather than bundling it here.

PR #5148, opened 2026-09-16
ci: verify Makefile-referenced intentd targets exist at the pinned gitlink

panghy:
## Summary

Adds a compile-free check that fails a monorepo PR whose root Makefile references an intentd crate (-p) or cargo test target (--test) that does not exist at the intentd gitlink recorded in the PR head.

- scripts/check-makefile-targets.mjs: parses tab-indented recipe lines rooted in INTENTD_DIR (cd or --manifest-path), extracts -p/--package and --test references (handling line continuations, Make recipe flags @/-/+, and unquoted shell comments), resolves the pin with git rev-parse HEAD:packages/intentd, and verifies crates/<dir>/tests/<name>.rs, tests/<name>/main.rs, or a [[test]] name entry exists at the gitlink using git object reads only (never the submodule working tree). Exit 1 names Makefile:line, the missing path, the pin, and an auto-bump hint; exit 2 when the gitlink object is not present locally.
- scripts/check-makefile-targets.test.mjs: 19 hermetic node --test cases (inline Makefile text plus throwaway git repos).
- Makefile: new check-makefile-targets target (depends on ensure-intentd-submodule), first prerequisite of make check.
- CI: the docs-check job runs the test file and make check-makefile-targets (it already initializes packages/intentd at the pin).
- AGENTS.md: one clause on the existing submodule-pins sentence.

## Motivation

intent-hq/intent#5133 added lint-fixed-sleeps referencing fixed_sleep_lint while its pinned intentd (8d5f28dd) lacked crates/intent-core/tests/fixed_sleep_lint.rs; all monorepo checks were green and the ordering against intent-hq/intentd#1926 was enforced only by draft status and coordinator instruction. Running this script with --gitlink 8d5f28dd reproduces that failure at the lint-fixed-sleeps recipe line.

## Verification

- node --test scripts/check-makefile-targets.test.mjs: 19/19 pass
- make check-makefile-targets: exit 0 in ~0.07 s (9 references checked against intentd@098b585, no cargo compile)
- make docs-check, node scripts/check-root-hygiene.mjs, node scripts/check-doc-links.mjs: pass
- Negative: a scratch recipe with --test does_not_exist under -p intent-core exits 1 naming the line, crates/intent-core/tests/does_not_exist.rs, the pin, and the hint
- Historical: --gitlink 8d5f28dd exits 1 on fixed_sleep_lint (local-only; CI's shallow init does not carry the old objects)

Monorepo-only change; no gitlink moves.


scripts/check-makefile-targets.mjs
```diff
@@ -0,0 +1,310 @@
...
+    }
+  }
+  return { subcommand, packages, tests };
+}
+
+export function parseMakefile(text) {
+  const invocations = [];
+  for (const { line, text: logical } of joinContinuations(text)) {
+    if (!isIntentdRecipe(logical)) continue;
+    for (const segment of recipeCommand(logical).split(SHELL_SEPARATOR)) {
```
panghy: **should-fix — preserve shell quoting when splitting commands and arguments**

Splitting the recipe on every `&&` / `;` / `|` loses quote context before the echo guard runs. A tab-indented `@echo "cd $(INTENTD_DIR) && cargo test -p nonexistent --test missing"` is only a help message, but the checker reports an unknown package and exits 1 because the second fragment no longer contains `echo`. Separately, the whitespace split in `extractCargoReferences` retains shell quotes: `cd $(INTENTD_DIR) && cargo test -p "intent-core" --test "fixed_sleep_lint"` exits 1 claiming that the literal package `"intent-core"` is unknown, although the unquoted equivalent passes against the same pin.

I reproduced both through the CLI at f64dc10 and confirmed shell behavior with a no-op cargo function. This makes valid recipe quoting or help-only changes fail `make check` / docs-check. Please use quote-aware command/word tokenization, removing syntactic quotes from argument values while preserving quoted separators as data, and add parser + CLI regressions for these cases (including quoted `--package=` / `--test=` forms).

scripts/check-makefile-targets.mjs
```diff
@@ -0,0 +1,310 @@
...
+// Drops an unquoted `#` comment (one that starts a word) from a shell line.
+export function stripShellComment(text) {
+  let quote = null;
+  for (let i = 0; i < text.length; i += 1) {
+    const char = text[i];
+    if (quote) {
+      if (char === quote) quote = null;
+    } else if (char === '"' || char === "'") {
+      quote = char;
+    } else if (char === '#' && (i === 0 || /\s/.test(text[i - 1]))) {
```
panghy: **should-fix — recognize comments immediately after shell operators**

A shell comment begins at a new token, not only after whitespace. For example, a tab-indented `cd $(INTENTD_DIR) && cargo test -p intent-core --test fixed_sleep_lint;# cargo test -p nonexistent --test missing` executes only the existing test: `;` ends the command and `#` starts a comment. This condition leaves the comment intact; the later separator split then extracts its fake cargo invocation. The CLI consequently exits 1 on package `nonexistent` instead of passing.

Reproduced against pin 098b585 and independently checked with a shell cargo stub (only the real invocation runs). Please recognize unquoted/unescaped comment starts at shell token boundaries, including after control operators, and add a regression that this recipe checks only the two real references. Keep the existing quoted/mid-word hash cases passing.

panghy (commented):
## Review summary
Comment-only review of f64dc104e0850df9acb985157e5d53fa5726bc4e. **Two should-fix findings**, posted inline: [shell quoting is lost](https://github.com/intent-hq/intent/pull/5148#discussion_r4023226043) and [comments after operators are treated as commands](https://github.com/intent-hq/intent/pull/5148#discussion_r4023226054). Both are reproducible false positives in otherwise valid recipes; no blockers or nits beyond these findings. Please add regression tests and address them before landing.

## Verification
- `node --test scripts/check-makefile-targets.test.mjs`: exit 0, 19/19 passed (local Node v24.19.0).
- `node scripts/check-makefile-targets.mjs` and `make check-makefile-targets`: exit 0, 9 references at 098b585; existing Makefile recipes do not currently hit the reported edge cases.
- `make docs-check`, `node scripts/check-root-hygiene.mjs`, and diff whitespace check: pass.
- `make -n check`: checker precedes fmt in the ordinary serial make traversal.
- Missing-target CLI repro: expected exit 1, line/path/pin/auto-bump hint present. Historical `--gitlink 8d5f28dd`: expected exit 1 at Makefile:419 for fixed_sleep_lint. Missing object: expected exit 2 with init/fetch guidance.
- Both branch-base and origin/main Makefile contents pass against the branch pin.
- [Live docs-check job](https://github.com/intent-hq/intent/actions/runs/35065141187/job/104693655924) is successful, including both new steps. Workflow uses Node 22 and initializes the pinned submodule with depth 1 before checks; no older history is needed for normal operation.

The git reader consistently uses the resolved HEAD gitlink via cat-file/ls-tree, never submodule filesystem content; the dirty-file fixture and historical-pin run support that invariant. Failure output is actionable. ESM exports, main guard, dependency-free implementation, and temporary git fixtures match the check-submodule-pins style. No files changed, no compilation, no approval or merge performed.

scripts/check-makefile-targets.mjs
panghy: Fixed in f77d9c3679bee4e8d692b2aee375c52d36329f2d. The regex separator split and whitespace tokenizer are replaced by `splitShellCommands`, a quote-aware tokenizer (single quotes literal; double quotes with backslash escapes of `"` `\` `$` and backquote; unquoted backslash escapes) that splits into simple commands only on unquoted `&&` / `||` / `;` / `|`. Root-marker detection (`isIntentdRootedCommand`) and `extractCargoReferences` now operate on word lists, so `@echo "cd $(INTENTD_DIR) && cargo test -p nonexistent --test missing"` yields a single `echo` word plus one string argument (no reference), and `-p "intent-core" --test "fixed_sleep_lint"` resolves with the quotes removed. Regressions added for quoted `-p` / `--test` / `--package=` / `--test=` values, echo/printf-wrapped text, and a CLI success case covering both; both repros pass against pin 098b585.

scripts/check-makefile-targets.mjs
panghy: Fixed in f77d9c3679bee4e8d692b2aee375c52d36329f2d. Comment recognition is now part of the tokenizer: an unquoted, unescaped `#` that starts a new word — at the beginning of the line, after whitespace, or immediately after a control operator (`;` `&&` `||` `|`) — begins a comment running to end of line, while `#` inside a word (`foo#bar`) or in quotes is data. `cd $(INTENTD_DIR) && cargo test -p intent-core --test fixed_sleep_lint;# cargo test -p nonexistent --test missing` now checks only the real reference and passes against pin 098b585. Regressions added for `;#`, `&&#`, `|#`, a `true;# …` tail after a real intentd recipe, mid-word `#`, escaped `\#`, and the existing quoted-hash cases.

scripts/check-makefile-targets.mjs
```diff
@@ -0,0 +1,310 @@
...
+        );
+      }
+    }
+    for (const test of tests) {
+      checked += 1;
+      if (packages.length === 0) {
+        failures.push(
+          `${makefile}:${line}: error: cargo test target '${test}' cannot be verified: the cargo invocation names no -p/--package crate`,
+        );
+      } else if (resolved.length > 0 && !resolved.some((crate) => hasTestTarget(reader, crate, test))) {
```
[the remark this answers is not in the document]
panghy: Fixed in e1ce8dbcdc998128ca0812c520f09cb6ae7e1baf. `verifyInvocations` now iterates every resolved crate of the invocation and emits one failure line per crate that lacks the `--test` target, naming that crate and its expected `crates/<dir>/tests/<name>.rs` path, instead of passing on the first match. Regressions added: a unit test with two `-p` packages where only one has the target (each direction), neither has it (two lines), and both have it (no failures); the hermetic CLI exit-1 case now includes `cargo test -p foo -p bar-crate --test flat` and reports only `bar-crate`, and the CLI exit-0 case includes `-p foo -p bar-crate --test common` with the target present in both crates. Against pin 098b585, `cargo test -p intent-core -p intentd --test fixed_sleep_lint` now exits 1 naming `intentd`.

scripts/check-makefile-targets.mjs
```diff
@@ -0,0 +1,366 @@
...
+// `null` for non-recipe lines.
+export function recipeCommands(logicalText) {
+  if (!logicalText.startsWith('\t')) return null;
+  return splitShellCommands(recipeBody(logicalText));
+}
+
+// A command is rooted in intentd when it is `cd $(INTENTD_DIR)` or passes
+// `--manifest-path $(INTENTD_DIR)/Cargo.toml`.
+export function isIntentdRootedCommand(words) {
+  if (words[0] === 'cd' && words[1] === '$(INTENTD_DIR)') return true;
```
panghy: **should-fix — retain intentd-root detection after shell control keywords**

The quote-aware refactor introduces a false negative for a recipe that the previous checker handled: a tab-indented `if true; then cd $(INTENTD_DIR) && cargo test -p intent-core --test does_not_exist; fi` now exits 0 with `checked 0 cargo references`. After splitting at `;`, the words are `["then", "cd", "$(INTENTD_DIR)"]`, so requiring `cd` at index 0 misses the root and skips the entire invocation. A grouped `{ cd $(INTENTD_DIR) && cargo test -p intent-core --test does_not_exist; }` has the same issue. These are actual cargo invocations, not quoted help text.

I reproduced this through the CLI and compared exported parsing before/after: f64dc10 detects the conditional invocation; f77d9c3 does not. A shell cargo stub confirms the conditional executes the command. This defeats the missing-target gate when a recipe gains an ordinary condition or command group. Please recognize command position after the supported shell control/group prefixes without reintroducing the quoted-help false positives. Add parser and CLI regressions, including a backslash-continued conditional; the missing target should exit 1 with its recipe line/path/pin/hint, while an existing target should still be checked successfully.

The three original quote/echo/operator-comment CLI repros are fixed. Leaving the earlier threads unresolved pending this regression fix. Local tests passed (24 tests, including another agent's uncommitted multi-package test); CI docs-check is green on f77d9c3.

panghy: Fixed in 85243746e913fdf11b6ff27d20c1755da1e5cb43. A new `commandWords` step classifies each simple command from its command name on: leading shell reserved words and group openers (`if` `then` `else` `elif` `do` `while` `until` `!` `{` `(`) and `NAME=value` assignments are skipped before checking for `cd $(INTENTD_DIR)` or `cargo`, and `--manifest-path $(INTENTD_DIR)/Cargo.toml` on the cargo command is a root marker regardless of prefix. Quoted text stays a single word, so `if true; then echo "cd $(INTENTD_DIR) && cargo test …"; fi` remains inert. Regressions added: parser cases for the `if/then` form, `{ …; }`, `( … )`, env-prefixed `cargo`, an env-prefixed `--manifest-path` form after `then`, `while/do/!`, and two quoted-echo negatives; the hermetic CLI exit-1 case includes a backslash-continued `if true; then \ … cargo test -p foo --test conditional; \ fi` reported at its first physical line with path/pin/hint, and the exit-0 case includes `if true; then cd $(INTENTD_DIR) && RUST_LOG=debug cargo test -p foo --test flat; fi` and a grouped `{ … --test declared; }`. Against pin 098b585 both of your repros now exit 1 naming `does_not_exist`.

scripts/check-makefile-targets.mjs
```diff
@@ -0,0 +1,381 @@
...
+import { fileURLToPath } from 'node:url';
+
+export const DEFAULT_INTENTD_DIR = 'packages/intentd';
+export const CARGO_SUBCOMMANDS = new Set(['test', 'run', 'build', 'clippy', 'nextest', 'check', 'bench']);
+export const HINT =
+  'hint: the referenced artifact probably lives on an unmerged intentd branch; merge the intentd PR and wait for auto-bump-submodules to advance the pin before landing this monorepo change.';
+
+const USAGE = 'usage: check-makefile-targets.mjs [--makefile <path>] [--gitlink <sha>] [--intentd-dir <path>]';
+const INTENTD_MANIFEST = '$(INTENTD_DIR)/Cargo.toml';
+const SHELL_OPERATORS = ['&&', '||', ';', '|'];
```
panghy, line 16: **should-fix — tokenize subshell parentheses independently of surrounding whitespace**

The group-prefix fix only recognizes `(` when whitespace already made it a separate word. In shell syntax parentheses are operators and do not require those spaces. A tab-indented `(cd $(INTENTD_DIR) && cargo test -p intent-core --test does_not_exist)` still exits 0 with `checked 0 cargo references`: the tokenizer produces `(cd`, so root detection skips the recipe. Conversely, `( cd $(INTENTD_DIR) && cargo test -p intent-core --test fixed_sleep_lint)` exits 1 looking for `fixed_sleep_lint).rs`, because the closing parenthesis becomes part of the test argument. This leaves both false-negative and false-positive variants of the grouped-recipe finding.

Reproduced through the CLI on clean 85243746e913fdf11b6ff27d20c1755da1e5cb43. An independent shell cargo stub receives `fixed_sleep_lint` without a closing parenthesis. Please handle unquoted/unescaped subshell opener/closer boundaries while preserving Make expressions such as `$(INTENTD_DIR)` (do not blindly split every parenthesis), and add parser + CLI regressions for compact opener/closer forms, a missing target, and an existing target.

All six previously posted exact repros now pass, as do 25 tests and the current/historical pin checks; this is the remaining should-fix from cumulative review. No threads resolved yet.

panghy: Fixed in 0d0dcf94ca4ded23cf6ea534542f0d24b8eb4202. The tokenizer now emits an unquoted `(` / `)` as a word of its own regardless of surrounding whitespace, except when the parenthesis belongs to an expansion: `$(...)`, `$$(...)` and `$((...))` are consumed with balanced nesting and stay inside their word (`$(INTENTD_DIR)`, `$(shell echo $(A))`, `$$(pwd)/x`, `$((1+2))` all survive intact; `\(` and quoted `"(y)"` are literal). `commandWords` strips trailing `)` / `}` alongside the leading openers before root/cargo classification. Regressions: tokenizer/parser cases for the compact opener `(cd $(INTENTD_DIR) && cargo test … --test does_not_exist)`, the compact closer `( cd … --test fixed_sleep_lint)`, `(cd $(INTENTD_DIR)&&cargo test …)`, an inert `(echo "(cd $(INTENTD_DIR) && cargo test …)")`, and an expansion-survival case where `--test $$(basename $(shell pwd))` remains one word; the hermetic CLI exit-1 case adds `(cd $(INTENTD_DIR) && cargo test -p foo --test compact_missing)` reported at Makefile:11 with path/pin/hint, and the exit-0 case adds `( cd … -p foo --test flat)` and `(cd … -p bar-crate --test declared)`. Against pin 098b585 your compact opener repro now exits 1 naming `does_not_exist` and the compact closer repro exits 0 resolving `fixed_sleep_lint`. 27/27 tests; current pin exit 0 (9 refs); `--gitlink 8d5f28dd` exit 1 at Makefile:419.

scripts/check-makefile-targets.mjs
```diff
@@ -0,0 +1,310 @@
...
+import process from 'node:process';
+import { fileURLToPath } from 'node:url';
+
+export const DEFAULT_INTENTD_DIR = 'packages/intentd';
+export const CARGO_SUBCOMMANDS = new Set(['test', 'run', 'build', 'clippy', 'nextest', 'check', 'bench']);
+export const HINT =
+  'hint: the referenced artifact probably lives on an unmerged intentd branch; merge the intentd PR and wait for auto-bump-submodules to advance the pin before landing this monorepo change.';
+
+const USAGE = 'usage: check-makefile-targets.mjs [--makefile <path>] [--gitlink <sha>] [--intentd-dir <path>]';
+const INTENTD_ROOT_MARKERS = [/\bcd\s+"?\$\(INTENTD_DIR\)"?\s+&&/, /--manifest-path[= ]"?\$\(INTENTD_DIR\)"?\/Cargo\.toml/];
```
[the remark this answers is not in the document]
panghy: Fixed by the quote-aware tokenizer in f77d9c3679bee4e8d692b2aee375c52d36329f2d. Independently reverified on head 0d0dcf94ca4ded23cf6ea534542f0d24b8eb4202: both `cd $(INTENTD_DIR); cargo test -p intent-core --test does_not_exist` and the single-quoted directory variant exit 1 with the expected line/path/pin/hint; the existing fixed_sleep_lint variant exits 0 and checks 2 references. Resolving.

scripts/check-makefile-targets.mjs
```diff
@@ -0,0 +1,310 @@
...
+export function isIntentdRecipe(logicalText) {
+  const command = recipeCommand(logicalText);
+  if (command === null) return false;
+  return INTENTD_ROOT_MARKERS.some((marker) => marker.test(command));
+}
+
+// Arguments after a bare `--` belong to the invoked binary and are skipped; an
+// echoed `cargo ...` hint is text, not an invocation.
+export function extractCargoReferences(segment) {
+  const tokens = segment.trim().split(/\s+/);
```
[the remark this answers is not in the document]
panghy: Fixed by quote-aware argument tokenization in f77d9c3679bee4e8d692b2aee375c52d36329f2d. Independently reverified on head 0d0dcf94ca4ded23cf6ea534542f0d24b8eb4202: `cd $(INTENTD_DIR) && cargo test -p 'intent-core' --test 'event_type_lint'` exits 0 and checks 2 references; double-quoted and quoted --package=/--test= forms also pass. Resolving.

Issue #4858, opened 2026-09-12
ws.script.start resolves before script.status leaves idle; same-eval hook fires a false finished wake


panghy:
## Symptom

`ws.script.start(scriptId)` resolves before `ws.script.status(scriptId)` reflects the launch. A self-checking hook scheduled in the same eval immediately after `start` (the pattern AGENTS.md recommends for long gates) sees `status: "idle"` with no `exitCode` on its validation run and dispatches a false "finished" wake.

## Evidence

Workspace `reliable-test` (intent monorepo), agent "Gate and PR", 2026-09-12 ~14:06:48Z, running `make test RESUME=1` as a saved command-mode script:

1. One eval: `await ws.script.start("3e537b11-…")` then `await ws.hook.schedule({ name: "Wait for make test RESUME=1", delayMs: 60_000, code: <poll ws.script.status; dispatch unless status is "running" or "starting"> })`.
2. The hook's immediate validation run dispatched: `[Background hook "Wait for make test RESUME=1"] make test RESUME=1 finished: status=idle exit=undefined`.
3. A `ws.script.status` read a few seconds later returned `status: running, pid: 232454, startedAt: 2026-09-12T14:06:48.265687194Z, restartCount: 0`.
4. The agent rescheduled the hook with a stricter condition (`exitCode !== undefined`); the real completion arrived at 14:08:02Z with `status=exited exit=0`.

## Cost

One wasted agent turn plus a rescheduled hook per long gate started this way. The recommended "start script, then hook on status" pattern is racy by default; any hook that treats "not running" as "finished" (the natural reading) fires immediately.

## Proposed fix (rung 1: make the mistake impossible)

`ws.script.start` should not resolve until the supervisor has moved the script out of `idle` (to `starting`/`running`, or to `exited`/`error` if spawn fails synchronously), so a status read after `start` returns can never observe the pre-launch `idle`. Alternatively, `start` returns a run id / `startedAt` and `status` exposes a distinct pre-launch state (e.g. `launching`) so `idle` unambiguously means "no run pending".

Weaker interim (rung 3): the canonical gate-wait hook snippet in the monorepo AGENTS.md (Resuming local Rust gates) should test `s.status === "exited" && s.exitCode !== undefined` rather than "not running".

## Verification

- Regression test in intentd script_ops: call `start`, then `status` with no intervening yield; assert `status != "idle"`.
- Schedule the recommended hook in the same eval as `start`; the validation run must return `dispatch: false`.

## Dedup

[...]

PR #5207, opened 2026-09-16
docs(scripts): document the total saved-script exit contract and the -1 lost-process sentinel

panghy:
## Summary

Documents the total saved-script exit contract landed by intent-hq/intentd#1942 (https://github.com/intent-hq/intentd/pull/1942), and updates the agent guidance for exit-only gate hooks.

- `docs/protocol/methods/scripts.md`: `exited` always carries `exitCode`; an unobservable or lost exit reports `exitCode: -1` plus an `error` string (`"exit status unobservable"`, or `"lost: the daemon stopped while the script was running"` for a command-mode row that was running at daemon stop). The `script.run` result `exitCode` follows the same rule. Command-row dismiss/persistence semantics; `previouslyRunning` stays service-only. Marked "new in intentd, unreleased" — to be swapped for the carrying tag once the intentd fix ships.
- `AGENTS.md` → Resuming local Rust gates: the dispatch condition is unchanged; one added sentence explains that `exitCode === -1` means the exit could not be observed (read `s.error`, treat as failure). The bullet is compressed back to its pre-change length (121 words / 11 lines).

Docs only; no `packages/*` gitlink moves. `make docs-check` passes.

Merge after intent-hq/intentd#1942 lands (the doc describes daemon behaviour that exists only from that change).


panghy (commented):
Reviewed the full diff at 80bc5d3 against intentd 6ddbb60c. No independent documentation defects found in the requested scope.

- Both error strings match the supervisor constants verbatim; the unobservable sentinel is -1, and the `script.run` result uses the same mapping as runtime state.
- `previouslyRunning` remains service-only; command hydration and dismiss behavior match the code.
- The AGENTS.md gate bullet is exactly 121 words / 11 lines before and after, retains the exit-only predicate and issue #4858 link, and correctly qualifies the command restart case.
- `make docs-check` passed (exit 0, three docs; cross-submodule checks skipped in this uninitialized docs worktree). `git diff --check` passed. The diff contains only AGENTS.md and docs/protocol/methods/scripts.md, both mode 100644: no packages/* gitlink moved. PR title and branch commit headlines are conventional; no release-footer hazard found.

Hold for the source correction in [intentd #1942](https://github.com/intent-hq/intentd/pull/1942): a reproduced noisy-descendant case can starve exit polling indefinitely, so the new liveness statement is not yet true for that path. This is a counterpart implementation blocker, not a request to weaken the documented contract. Actual hook exactly-once delivery remains the explicitly deferred post-ship proof; the WSS tests establish the terminal predicate, not hook delivery.

PR #5389, opened 2026-09-18
test: isolate Node checker suites from host tracer instrumentation

panghy:
## Summary
Node test suites under `scripts/*.test.mjs` that spawn the checker/bridge CLIs with `process.execPath` inherited the parent environment. On dev hosts where Datadog dd-trace is injected (via `NODE_OPTIONS` / `DD_INJECTION_ENABLED`, or `/etc/ld.so.preload`), the tracer's startup banner lands in the child's stderr and `check-makefile-targets.test.mjs`'s exact-stderr assertion fails although the checker is correct.

## Change
- Add `scripts/test-env.mjs` exporting `cleanNodeEnv()`: copies `process.env`, deletes `NODE_OPTIONS` and every `DD_*` key, and sets `DD_TRACE_ENABLED=false` / `DD_TRACE_STARTUP_LOGS=false` so ld.so.preload-injected tracers stay silent too.
- Use it at every `process.execPath` spawn site: `check-makefile-targets.test.mjs`, `uds-rpc.test.mjs`, `uds-ws-bridge.test.mjs`.
- Add a regression test that runs the checker CLI under a noisy `NODE_OPTIONS` preload in the parent and asserts exact stderr (observed failing before the fix).

No checker scripts, CI workflows, assertion strings, dependencies, or submodule pins change.

## Verification (instrumented host, env intact)
- `node --test scripts/check-makefile-targets.test.mjs scripts/uds-rpc.test.mjs scripts/uds-ws-bridge.test.mjs` — 75/75 pass
- same under `NODE_OPTIONS="--require <stderr-noise preload>"` — 75/75 pass
- `node --test scripts/*.test.mjs` — 126/126 pass
- `make docs-check` — exit 0

Monorepo-only test change; nothing ships to the alpha channel.


panghy (commented):
No actionable findings in 847c98bcc438ac545c30c66bc3d5c5d29f4991cd.

Reviewed cleanNodeEnv(), all scripts/*.test.mjs subprocess sites, the noisy-preload regression, and the unchanged CI invocations. All four process.execPath spawn sites use the helper. Existing assertions are unchanged; checker scripts, dependencies, and CI workflows are untouched. The helper copies rather than mutates the parent environment, strips NODE_OPTIONS/DD_* values, and retains the two explicit tracer-disable flags documented in the spec.

Independent verification:
- Full node --test scripts/*.test.mjs with inherited host NODE_OPTIONS and DD_INJECTION_ENABLED: exit 0.
- The three changed suites with a unique noisy NODE_OPTIONS preload: exit 0; parent noise confirmed.
- Each changed suite with its plain CI-style node --test command and parent NODE_OPTIONS unset: exit 0.
- make docs-check: exit 0, all invariants passed.
- Regression negative control, using an in-memory source copy with only the CLI env isolation bypassed: expected exit 1 with NOISE and a strict assertion failure. Passing and failing runs both removed their fixtures.
- Direct cleanup harness verified restoration when NODE_OPTIONS and DD_TRACE_DEBUG were initially absent and initially present, and removal of the temporary preload directory.

Local runtime was Node 24.19.0 on Linux; the PR bridge-test and docs-check jobs also passed on CI Node 22. macOS compatibility was reviewed statically, not executed. No repository files changed. Comment-only review, not a formal approval.

PR #5394, opened 2026-09-18
feat: report cargo-llvm-cov in make doctor and make status

panghy:
## Summary

Agents had no authoritative answer to "can I run the instrumented coverage gate here?": make doctor / STATUS_JSON=1 make status only reported cargo-nextest. On #5366 an implementor stated cargo-llvm-cov was missing and skipped the local coverage run; the verifier found 0.9.0 installed and required a correction round.

This PR makes the local instrumented gate a discoverable path (retrospective rung 3):

- **make doctor**: new non-blocking row, right after cargo-nextest, using the existing [optional] convention (never counted as a gap, so doctorOk is unchanged):
  - present: cargo-llvm-cov: VERSION with llvm-tools-preview (make coverage-e2e / coverage-all)
  - tool present, component missing: names rustup component add llvm-tools-preview
  - absent: not installed, naming BOOTSTRAP_COVERAGE=1 make bootstrap-dev-host and the manual cargo install / rustup commands
- **bootstrap-dev-host**: opt-in install_coverage_tooling step (--coverage flag or BOOTSTRAP_COVERAGE=1, passed through by the Makefile). Off by default because cargo install cargo-llvm-cov compiles from source and make test does not need it; the default bootstrap is unchanged.
- **STATUS_JSON=1 make status**: host block gains coverageTooling {ready, detail} parsed from the doctor row (ready only when both tool and component are present; {false, "unknown"} when the row is absent); human make status prints a Coverage line after Host.
- Tests in scripts/bootstrap-dev-host.test.sh and scripts/dev-status.test.sh; one-line docs mentions in README.md and docs/fe/DEVELOPER_GUIDE.md.

Monorepo-only (scripts + docs); no submodule changes.

## Verification

- make doctor on this host prints the version row; with cargo-llvm-cov hidden from PATH prints the not-installed row with an unchanged gap count.
- STATUS_JSON=1 make status | jq .host shows coverageTooling.ready true here; fixture runs cover the missing-component, not-installed, missing-row and unavailable-doctor branches (ready false in every non-ready case).
- BOOTSTRAP_COVERAGE=1 scripts/bootstrap-dev-host.sh --yes prints a [skip] for coverage tooling; the default run does not mention it.
- bash scripts/bootstrap-dev-host.test.sh and make docs-check pass. scripts/dev-status.test.sh and scripts/dev-sandbox.test.sh fail identically on the unmodified baseline (host-load port-probe timeout / UI sandbox readiness); the new assertions and the doctor-wording guards pass in isolation.

Origin: retrospective finding from #5366.


scripts/dev-status.test.sh
```diff
@@ -87,19 +91,55 @@ SH
...
+PY
+grep -q '^Coverage   cargo-llvm-cov not installed$' \
+  <(CARGO_HOME="$temp_dir/no-cargo" PATH="$bin_dir" SANDBOX_STATE_DIR="$state_dir" bash "$script") \
+  || fail "human status did not report cargo-llvm-cov as not installed"
+
+# With the host PATH restored, coverageTooling.ready must match whether this
+# host can actually run cargo-llvm-cov with llvm-tools-preview installed.
+expected_coverage_ready=false
+if PATH="${CARGO_HOME:-$HOME/.cargo}/bin:$PATH" cargo llvm-cov --version >/dev/null 2>&1 \
+  && PATH="${CARGO_HOME:-$HOME/.cargo}/bin:$PATH" rustup component list --installed 2>/dev/null | grep -q '^llvm-tools'; then
```
panghy: [P2 / should-fix] Compute the expected llvm-tools state for the same pinned toolchain as doctor. llvm_tools_ready uses component list --toolchain "$TOOLCHAIN" --installed, but this oracle checks the active/default toolchain. This repo has no root rust-toolchain file, so a developer with e.g. a newer default and the pinned 1.96.0 installed can have different component sets. Reproduced with the real helpers and a rustup shim: pin_has_tools=1/default_has_tools=0 gives test_expected=false/doctor_actual=true; the inverse also disagrees. The assertions at lines 132/134 then reject correct production output. Read the same pin (respecting INTENTD_DIR) for the expectation, or use deterministic per-toolchain fixtures, and cover both disagreement directions.

scripts/bootstrap-dev-host.sh
```diff
@@ -447,9 +487,26 @@ installable_gap_exists() {
...
   return 1
 }
 
+# Optional row: never counts toward FAILURES, so doctorOk is unaffected.
+report_coverage_tooling() {
+  local version
+  if version=$(llvm_cov_version); then
+    if llvm_tools_ready; then
+      optional "cargo-llvm-cov: $version with llvm-tools-preview (make coverage-e2e / coverage-all)"
+    else
```
panghy, line 517: [P2 / should-fix] Print a component-install command for the toolchain that was checked. llvm_tools_ready explicitly probes the pinned TOOLCHAIN, but this suggested command (and LLVM_COV_INSTALL above) installs into the current/default toolchain. With an initialized checkout and a different root default, copying this command succeeds without repairing the missing component for the pin, so the same optional row remains. Reproduced with the actual report/probe functions and a per-toolchain rustup shim; the before/after row is identical. Include --toolchain "$TOOLCHAIN" when the pin is known (retain the unqualified fallback when it is not), in both the component-only and full-install suggestions. The task spec prescribed the current literal wording, so coordinate the wording/test update with the spec owner.

panghy (commented):
Critical code review of 01cf775 + 41a4719 against 1ed479b. Two should-fix (P2) findings are attached inline: the status test and the printed remediation select a different Rust toolchain from the probe. No P0/P1 issue found. Validation: bootstrap-dev-host.test.sh PASS (existing built-pty assertions skipped); make docs-check PASS; bash syntax and git diff --check PASS; real make doctor/status report optional coverage ready; dev-status.test.sh fails at the existing ports-shape assertion documented in the spec. Independent in-memory tests pass for parser ready/component-missing/cargo-missing/unknown/failed-doctor states, and reproduce both toolchain mismatches. Bash 3.2/macOS compatibility reviewed statically only. This is a COMMENT review; no formal approval or request-changes decision.

scripts/bootstrap-dev-host.sh
panghy: Fixed in 67e01d5: the doctor row and the not-installed row now print `rustup component add llvm-tools-preview --toolchain <pin>` via `llvm_tools_install_command` / `llvm_cov_install_command` (unqualified fallback when no pin is readable). The bootstrap test covers both pin/default disagreement directions and the no-pin fallback.

scripts/dev-status.test.sh
panghy: Fixed in 67e01d5: the oracle now reads the channel from `${INTENTD_DIR:-packages/intentd}/rust-toolchain.toml` and queries `rustup component list --installed --toolchain <pin>` (unqualified when no pin), matching `llvm_tools_ready`. Verified with an `INTENTD_DIR` fixture pinned to an uninstalled toolchain: the doctor reports not-ready and the test agrees.

panghy (commented):
Re-verified commit 67e01d5: both of my P2 findings are addressed; no remaining in-scope defects. Printed component-only and full-install remedies now target the known pin, with the required no-pin fallback. The status-test oracle honors INTENTD_DIR and queries that same pin. Fresh independent stubbed tests pass for both pin/default disagreement directions, both agreement states, both no-pin states, and exact missing-tool command strings. Bootstrap tests, docs-check, syntax/diff checks, and isolated sandbox doctor/preflight guards pass. The full status suite still encounters the accepted baseline ports failure; all current status tests pass when only the existing ports/timing guards are omitted in memory. Both review threads are resolved and their replies accurately describe the fix. No toolchain installs/downloads were performed during this re-review. This is a COMMENT review, not a formal approval or merge action.

PR #5410, opened 2026-09-19
fix(doctor): never auto-install the Rust toolchain from read-only probes

panghy:
## Summary
`make doctor` / `STATUS_JSON=1 make status` (`scripts/bootstrap-dev-host.sh --check`) probed the pinned toolchain with `rustup run <pin> rustc --version` and `rustup component list --toolchain <pin> --installed`. Recent rustup versions auto-install a named-but-absent toolchain on those calls, so a read-only diagnostic could download a multi-hundred-MB toolchain and mutate the host (observed while verifying #5394, where `rustup component list --toolchain <pin>` installed Rust 1.95.0 on the dev host).

## Changes
- `export RUSTUP_AUTO_INSTALL=0` script-wide. This is load-bearing, not belt-and-braces: under rustup 1.29.0 even `rustup toolchain list` auto-installs an absent override pin without it. Install mode is unaffected because its installs are explicit (`rustup toolchain install` / `rustup component add` / `rustup default`).
- New `toolchain_installed()` helper that only reads `rustup toolchain list`; `rust_toolchain_ready()` and the pinned branch of `llvm_tools_ready()` return "not ready" before issuing `rustup run` / `rustup component list --toolchain <pin>` when the pin is absent. Row wording is unchanged.
- Regression test in `scripts/bootstrap-dev-host.test.sh`: a logging `rustup` stub with absent-pin and installed-pin sub-cases, asserting no `run` / `component list --toolchain <pin>` call for an absent pin (and that the probes still run when the pin is present), plus a `RUSTUP_AUTO_INSTALL=0` sentinel. The test fails against the pre-fix script and passes after.

## Verification
- `bash scripts/bootstrap-dev-host.test.sh` — pass (exit 1 against the pre-fix script, exit 0 on this branch)
- `make docs-check` — pass
- `make doctor` on a host with the pin installed — output unchanged, `rustup toolchain list` inventory unchanged
- `scripts/dev-sandbox.test.sh` doctor fixture and both doctor-wording drift guards — pass on head and baseline. The full `dev-sandbox.test.sh` / `dev-status.test.sh` runs fail on the verification host with the same pre-existing timing failures on `origin/main` (2ce4ea1) and on this branch (hardcoded readiness/sleep budgets and the `port_status` timeout in untouched scripts); tracked in #5411.

Monorepo-only script change; no submodule gitlink moved.


scripts/bootstrap-dev-host.test.sh
```diff
@@ -432,6 +434,63 @@ run_doctor
...
+: >"$rustup_log"
+run_doctor
+expect_line "[missing]  Rust toolchain: 1.96.0 with rustfmt and clippy"
+reject_line "[ok]       Rust toolchain:"
+expect_line "$row_no_llvm_tools"
+reject_line "$row_ready"
+grep -qx 'toolchain list' "$rustup_log" || fail "doctor did not consult rustup toolchain list: $(cat "$rustup_log")"
+! grep -q '^run ' "$rustup_log" || fail "doctor ran rustup run against an absent pin: $(cat "$rustup_log")"
+! grep -q 'component list --toolchain 1.96.0' "$rustup_log" || fail "doctor listed components of an absent pin: $(cat "$rustup_log")"
+! grep -q '^RUSTUP_AUTO_INSTALL=' "$rustup_log" || fail "rustup ran without RUSTUP_AUTO_INSTALL=0: $(grep '^RUSTUP_AUTO_INSTALL=' "$rustup_log" | head -n 1)"
```
panghy, line 482: Non-blocking test-hardening nit: run_doctor inherits RUSTUP_AUTO_INSTALL from the caller, so this sentinel can miss removal of the script-wide export when the test runner already has RUSTUP_AUTO_INSTALL=0. I checked the new absent/present fixture in memory: removing only the export still passes with inherited 0, but fails with inherited 1. Please explicitly invoke both new sub-cases with RUSTUP_AUTO_INSTALL=1 (or unset it for the child), so they prove the script disables automatic installs itself regardless of the caller environment. Current production code correctly overrides inherited 1; this is regression-test robustness, not a current behavior bug.

panghy (commented):
Reviewed 93ccd46 against 2ce4ea1. No blocking correctness findings; this is a COMMENT review, not an approval. Exact, host-qualified, named (stable), and partial (1.96) helper matches behave correctly in Bash and bash --posix. Isolated real-rustup checks confirm that 1.96 is a distinct installed name from 1.96.0 and that the export suppresses implicit installs for list/show/which/run/component probes. Stubbed install_rust still explicitly installs/selects the default and skips on the next call. The bootstrap suite, docs-check, syntax checks, and diff whitespace checks passed; native pty.node assertions were skipped because dependencies are absent. I also independently reproduced the new fixture failing on the pre-fix script.

Non-blocking PR-description nit: the Verification paragraph says both full-suite failures happen before the suites reach the bootstrap script. That is true for the cited sandbox assertions, but not for dev-status: scripts/dev-status.sh calls doctor_status() at line 244 before port_status(), and the failing JSON assertion happens after that report is returned. Please say the failures are in unchanged timing-sensitive code and were reproduced on the baseline, and link the follow-up issue https://github.com/intent-hq/intent/issues/5411. This preserves the valid baseline evidence without claiming doctor was never executed.

One additional non-blocking test-environment hardening comment is attached inline. No source files were changed.

scripts/bootstrap-dev-host.test.sh
panghy: Done in f248f92: both sub-cases (absent pin, installed pin) now invoke the script as `RUSTUP_AUTO_INSTALL=1 run_doctor`, so the stub's sentinel only stays silent because `bootstrap-dev-host.sh` exports `RUSTUP_AUTO_INSTALL=0` itself, independent of the test runner's environment. Verified the gap and the fix: with the `export` line deleted from a scratch copy of the script and `RUSTUP_AUTO_INSTALL=0` inherited, the previous test still passed while the hardened test fails with `rustup ran without RUSTUP_AUTO_INSTALL=0: RUSTUP_AUTO_INSTALL=1`; with the export restored, `bash scripts/bootstrap-dev-host.test.sh` passes under inherited 0 and unset, and `make docs-check` passes.

panghy:
Re the Verification-paragraph nit in https://github.com/intent-hq/intent/pull/5410#pullrequestreview-5253677338: the PR body has been corrected to say the two full-suite failures are in unchanged timing-sensitive code, were reproduced on the baseline, and are tracked in https://github.com/intent-hq/intent/issues/5411 (no longer claims doctor never ran in dev-status). The inline test-hardening comment is addressed in f248f92.

Issue #5411, opened 2026-09-19
dev-sandbox.test.sh / dev-status.test.sh fail on loaded hosts due to hardcoded timing budgets


panghy:
## Summary
On a loaded dev host, the full `bash scripts/dev-sandbox.test.sh` and `bash scripts/dev-status.test.sh` runs fail on hardcoded timing budgets in untouched code paths. Observed identically on `origin/main` (2ce4ea1) and on the PR #5410 branch while verifying that PR; the failures occur before either suite reaches the code #5410 changes, so they are pre-existing.

## Failing assertions
- `scripts/dev-status.test.sh:75` (Python heredoc line 12): expected port-key subset of `report[ports]`, got an empty JSON report shape. Root cause: `scripts/dev-ports.sh` takes ~2.3–2.4 s on the host (measured 2.378 s head / 2.300 s baseline with a sanitized PATH), exceeding the hardcoded `timeout=2` in `port_status` at `scripts/dev-status.sh:84`, so the report contains `ports: {}`.
- `scripts/dev-sandbox.test.sh:341`: "signal-window sandbox did not become ready (iteration 2)" — hardcoded ~5 s readiness budget.
- `scripts/dev-sandbox.test.sh:450`: "sandbox-stop signalled a foreign live PID" — the isolated segment took 38.42 s against a hardcoded `sleep 30` budget.

With only those timing knobs relaxed in a scratch copy, the full `dev-sandbox.test.sh` (including the bootstrap fixture and doctor-wording drift guards) passes.

## Suggested fix
Make the readiness / sleep budgets and the `port_status` timeout host-tunable (env override with the current defaults) or derive them from a measured baseline, so the suites are reliable on slower hosts. `make docs-check` / the focused `bootstrap-dev-host.test.sh` are unaffected.

## Context
Found while verifying #5410 (`make doctor` rustup auto-install fix).

PR #5433, opened 2026-09-19
ci: lint unannotated fixed sleeps in the shell test suites

panghy:
## Summary

Adds a fixed-timing-window lint for the monorepo shell test suites (scripts/*.test.sh), mirroring intentd's Rust e2e timing-guard lint (crates/intent-core/tests/fixed_sleep_lint.rs + fixed_sleep_baseline.txt).

Origin: retrospective finding from #5411 / #5422 — three shell suites shared the same fixed-window defect and the dev-ports.test.sh instance was only found by a late loaded rerun. Nothing caught it before merge.

## What changes

- scripts/lint-fixed-sleeps.sh — flags bash sleep <literal>, python time.sleep(<literal>), and sleeping fixed-count loops in scripts/*.test.sh unless the line (or the line above) carries a '# timing-guard: <reason>' marker. Bash 3 compatible.
- scripts/fixed-sleep-baseline.txt — ratcheting baseline of grandfathered unannotated sites (bootstrap-dev-host.test.sh 1, dev-sandbox.test.sh 3). Counts may only go down; a stale, unsorted or zero entry fails with the exact corrective line.
- Existing suites: comment-only changes — timing-guard markers on justified poll intervals and placeholder lifetimes.
- scripts/lint-fixed-sleeps.test.sh — 81 pinned fixtures (positive/negative parser cases, ratchet messages, marker placement, self-skip). Historically: 8 cases fail against the initial lint commit, 3 against the first parser fix, 0 at HEAD.
- Wiring: make lint-shell-sleeps (prerequisite of make check); the lint and its self-test run in the shell-tests and shell-tests-bash3 CI jobs; one paragraph in docs/fe/DEVELOPER_GUIDE.md.

## Not in scope

- Linting non-test scripts/*.sh or any submodule.
- Behavioural changes to the suites. The unbounded restart-writer poll at dev-sandbox.test.sh:483 is baselined rather than marked; adding a deadline there is a follow-up that burns the baseline entry down.

## Verification

- make lint-shell-sleeps exits 0; make -n check lists it.
- bash scripts/lint-fixed-sleeps.test.sh passes under bash 5.2 and real bash 3.2.57 (byte-identical transcript); the exact shell-tests / shell-tests-bash3 step commands pass locally.
- git diff origin/main -- 'scripts/*.test.sh' (excluding the new self-test) is comment-only; all four existing suites pass.
- Monorepo-only change: nothing ships to the alpha channel.


scripts/lint-fixed-sleeps.sh
```diff
@@ -0,0 +1,408 @@
...
+  len = length(masked)
+  cmdpos = 1
+  i = 1
+  while (i <= len) {
+    c = substr(masked, i, 1)
+    if (c == " " || c == "\t") { i++; continue }
+    if (c == ";" || c == "&" || c == "|" || c == "(" || c == ")") {
+      if (i < len && substr(masked, i + 1, 1) == c) i++
+      i++
+      cmdpos = (c != ")")
```
panghy: **[P2] Preserve loop tracking across a function-body opener**

Setting `cmdpos` to false for `)` means the following `{` in `f() {` is ignored by the `if (!cmdpos) continue` branch. A fixed-count loop on that same line is therefore never pushed. In an isolated `scripts/x.test.sh`, these four lines reproduce it:

- `f() { for i in {1..3}; do`
- `  :`
- `  sleep .1 # timing-guard: poll interval`
- `done; }`

With an empty baseline, the lint exits 0 and `--print-counts` prints nothing; the unmarked fixed-count header must still count even though its inner sleep is marked (the existing `loop-inner-marked` fixture asserts exactly that rule). Moving `{` / the loop onto separate lines makes it report one site. A second effect is worse for nesting: putting `f() { while false; do :; done; }` inside an already-open fixed loop skips the nested `while`, so its `done` pops the outer loop and hides a later marked sleep. Both reproductions parse under Bash 5.2 and 3.2 and fail detection identically.

Please recognize the function-body command boundary, without treating an argument named `{` as a boundary, and add regression fixtures for both a same-line function/for header and an inner function containing a while loop. Re-run the self-test with `BASH3_BIN=/tmp/bash32/bash3 REQUIRE_BASH3=1` (or the macOS CI leg). This is a blocking correctness issue for the promised nesting-aware fixed-count-loop predicate.

scripts/lint-fixed-sleeps.sh
```diff
@@ -0,0 +1,408 @@
...
+    emit(st_line[depth], st_text[depth], st_here[depth], st_above[depth])
+  depth--
+}
+
+FNR == 1 { depth = 0; m_above = 0 }
+{
+  m_here = classify_marker($0)
+  masked = mask($0)
+  ns = sleep_positions($0, spos)
+  nk = keyword_tokens(masked, kpos, kind)
```
panghy: **[P2] Keep heredoc data out of the shell loop-keyword stack**

`keyword_tokens(masked, ...)` currently runs on every heredoc body line. A Python heredoc inside a shell fixed-count loop can therefore change the shell loop stack: `for n in range(3):` pushes an unmatched shell loop, while `done = False` pops the enclosing shell loop. These are ordinary Python statements, not Bash reserved words.

Reproduced with valid Bash 5.2/3.2 syntax and Python bodies parsed by `ast.parse`:

1. A shell `for i in {1..3}; do` containing a `python3 - <<'CODE'` block with `for n in range(3):` / `print(n)`, followed by `CODE`, shell `done`, and an outside `sleep 1`, reports **two** sites. Only the outside sleep should count; the shell loop never slept.
2. A shell fixed loop containing a Python block with `import time`, `done = False`, and `time.sleep(.1) # timing-guard: poll interval` exits **0** with an empty baseline. The unmarked shell fixed-count header should still count.

Please use heredoc delimiter state to prevent heredoc lines (including the terminator) from being interpreted as shell loop keywords and pin both cases. Keep the existing raw sleep and marker predicates unchanged: the test suites can also generate shell stubs in heredocs, so suppressing all rule-1 sleep detection there would introduce a separate missed-sleep regression. The coordinator has included the loop-state correction in the PR follow-up scope.

panghy (commented):
## Review result: changes needed

Reviewed `08af7c05840de7a1e1737291a19e563352ac3508`. **2 blocking correctness findings, 1 non-blocking scope note.** I am posting a COMMENT review because the authenticated account is also this PR's author; this is not an approval.

### Blocking
1. [Function-body boundaries lose fixed-loop headers or pop the wrong loop](https://github.com/intent-hq/intent/pull/5433#discussion_r4052230416) — `scripts/lint-fixed-sleeps.sh:180-203`. The fixtures pass Bash syntax checks on 5.2 and 3.2 but the lint silently returns success.
2. [Python heredoc lines mutate the enclosing shell-loop stack](https://github.com/intent-hq/intent/pull/5433#discussion_r4052231845) — `scripts/lint-fixed-sleeps.sh:223-226`. Reproduced both a false positive on an empty shell loop and a missed unmarked fixed-count header. Keep raw sleep detection unchanged while fixing loop-keyword handling.

### Non-blocking scope note
At `keyword_tokens` lines 175/190, the line-local heuristic also treats a case pattern `done)` and an `echo` argument `done` on a backslash-continuation line as loop terminators. Both can hide a later marked interval's unannotated outer fixed-count header. These fall beyond the documented first-word-per-line token model; record the limitation or cover these as targeted follow-up cases rather than presenting this as a complete Bash parser.

### Checks and other requested review areas
- The current lint and self-test pass locally. Real Bash 3.2.57 and Bash 5.2.21 produce byte-identical fixture transcripts. `REQUIRE_BASH3=1` without an available Bash3 exits 1 as required. Both hosted `shell-tests` and `shell-tests-bash3` are green at this head.
- In scratch copies of the actual suites: adding `sleep 1` fails with `scripts/dev-ports.test.sh:176`; adding the previous-line reasoned marker restores success; removing one baselined dev-sandbox sleep produces the exact replacement `scripts/dev-sandbox.test.sh 2`. Diagnostic and ratchet behavior match the agreed contract, including the explicit last-resort baseline hint.
- `make lint-shell-sleeps`, `make docs-check`, and `git diff --check origin/main...HEAD` pass. `make -n check` places the new lint before the Rust checks in the normal serial plan; the old `lint-fixed-sleeps` alias is untouched. YAML parsed with duplicate-key rejection, and each new command occurs exactly once in its intended job; the macOS command sets `REQUIRE_BASH3=1`. Full Rust-backed `make check` was not run in this review (deferred per the spec).
- Existing suite changes are marker comments only. The committed baseline retains genuine timing assumptions, including the explicitly deferred unbounded restart-writer loop. The developer-guide paragraph is short and accurately names the target, marker placement, and baseline policy.
- Fixture helpers are documented and cases are isolated. Add the function/heredoc regressions to `harness_case`; remember its explicit case-count assertion at line 464 when extending the table. No fixture-layout refactor is needed to fix these findings.
- All eight commit subjects are conventional. The intermediate predicate fixes and the two test commits can be absorbed by the repository's normal squash merge; no force-push/history rewrite is needed. This is multi-commit, so the existing conventional PR title is the appropriate squash subject.

No repository edits, pushes, or merge actions were performed during review.

scripts/lint-fixed-sleeps.sh
panghy: Exact fixtures to pin, each in its own `scripts/x.test.sh` with an empty `scripts/fixed-sleep-baseline.txt`:

**Function-body fixed loop:** expected one finding on line 1, actual zero at 08af7c0.

<augment_code_snippet path="scripts/x.test.sh" mode="EXCERPT">
````bash
f() { for i in {1..3}; do
  :
  sleep .1 # timing-guard: poll interval
done; }
````
</augment_code_snippet>

**Nested function loop:** expected one finding on line 1, actual zero at 08af7c0.

<augment_code_snippet path="scripts/x.test.sh" mode="EXCERPT">
````bash
for i in {1..3}; do
  f() { while false; do :; done; }
  sleep .1 # timing-guard: poll interval
done
````
</augment_code_snippet>

File: scripts/lint-fixed-sleeps.sh (hunk 2 above)
Comment (panghy) [reply scripts/lint-fixed-sleeps.sh]: Exact fixtures to pin, each in its own `scripts/x.test.sh` with an empty `scripts/fixed-sleep-baseline.txt`:

**Python for must not open a shell loop:** expected one finding on line 7 only, actual two (lines 1 and 7) at 08af7c0.

<augment_code_snippet path="scripts/x.test.sh" mode="EXCERPT">
````bash
for i in {1..3}; do
python3 - <<'CODE'
for n in range(3):
    print(n)
CODE
done
sleep 1
````
</augment_code_snippet>

**Python done variable must not close a shell loop:** expected one finding on line 1, actual zero at 08af7c0.

<augment_code_snippet path="scripts/x.test.sh" mode="EXCERPT">
````bash
for i in {1..3}; do
python3 - <<'CODE'
import time
done = False
time.sleep(.1) # timing-guard: poll interval
CODE
done
````
</augment_code_snippet>

All four review fixtures pass `bash -n` under Bash 5.2 and 3.2; the Python bodies also pass `ast.parse`.

File: scripts/lint-fixed-sleeps.sh (hunk 1 above)
Comment (panghy) [reply scripts/lint-fixed-sleeps.sh]: Fixed in 5fba415 (`fix:`) with fixtures in c3f470b (`test:`). A closing `)` now restores command position, so `f() {` opens a body and a fixed-count loop on that line — or a nested `f() { while false; do :; done; }` inside an open loop — is tracked; `$(…)`/`<(…)`/`>(…)` are distinguished by the character before `(` and still end mid-word, and an argument `{` (`echo { done`) is not a boundary. Pinned as `loop-in-function-inline`, `loop-in-function-newline`, `loop-nested-function-while` plus subshell/command-substitution/process-substitution guards; both repros fail on 08af7c0 (summed 0 vs expected 1) and pass now. Repo output is byte-identical to 08af7c0 (counts 1/3); self-test passes under bash 5.2.21 and 3.2.57 (`BASH3_BIN=/tmp/bash32/bash3 REQUIRE_BASH3=1`) with identical transcripts. The `case` pattern `done)` limitation is left as noted.

File: scripts/lint-fixed-sleeps.sh (hunk 2 above)
Comment (panghy) [reply scripts/lint-fixed-sleeps.sh]: Fixed in 5fba415 (`fix:`) with fixtures in c3f470b (`test:`). `<<`/`<<-` heredoc bodies (quoted, unquoted or backslash-escaped delimiter; exact terminator line, tabs stripped for `<<-`; `<<<` here-strings and quoted/commented `<<` excluded) are now skipped by `keyword_tokens` only — `sleep_positions` and marker lookup still run on every line, so a generated shell stub with `sleep 5` or a Python `time.sleep(2)` inside a heredoc remains a finding (pinned as `heredoc-shell-sleep-still-counts`, `heredoc-python-sleep-still-counts`, `heredoc-marker-still-exempts`). Both repros pinned as `heredoc-python-for-is-data` (one finding on line 7) and `heredoc-python-done-is-data` (one finding on line 1); both fail on 08af7c0 and pass now. Repo counts unchanged (1/3); self-test passes under bash 5.2 and 3.2.

File: scripts/lint-fixed-sleeps.sh [hunk 3]
```diff
@@ -0,0 +1,408 @@
...
+
+# Fill pos[] with the offset of every fixed sleep on `line` (rules 1 and 2),
+# in offset order; return how many. Comment lines never count.
+function sleep_positions(line, pos,   n, t, rest, off, at, a, b, v) {
+  n = 0
+  split("", pos)
+  t = line
+  sub(/^[ \t]+/, "", t)
+  if (substr(t, 1, 1) == "#") return 0
+  rest = line
```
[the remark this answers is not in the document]
panghy: Fixed in 014c1a1. `sleep_positions` now receives the line cut at its `#` comment (the same word-start unquoted-`#` boundary `mask()` uses for markers, from f93ad1a), so `echo ok # sleep 1` and a heredoc line `x = 0  # time.sleep(1)` no longer count, and a loop whose only "sleep" is in a comment is not a site; a real `sleep 1 # explanatory text` still counts. Pinned as `comment-sleep-after-command`, `comment-python-sleep-in-heredoc`, `comment-after-real-sleep-still-counts`, `comment-sleep-inside-word-counts`, `loop-comment-sleep-only`; repo output byte-identical to 08af7c0.

scripts/lint-fixed-sleeps.sh
```diff
@@ -0,0 +1,408 @@
...
+  if (st_fixed[depth] && st_flag[depth])
+    emit(st_line[depth], st_text[depth], st_here[depth], st_above[depth])
+  depth--
+}
+
+FNR == 1 { depth = 0; m_above = 0 }
+{
+  m_here = classify_marker($0)
+  masked = mask($0)
+  ns = sleep_positions($0, spos)
```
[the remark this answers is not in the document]
panghy: Fixed in 014c1a1. `sleep_positions` now receives the line cut at its `#` comment (the same word-start unquoted-`#` boundary `mask()` uses for markers, from f93ad1a), so `echo ok # sleep 1` and a heredoc line `x = 0  # time.sleep(1)` no longer count, and a loop whose only "sleep" is in a comment is not a site; a real `sleep 1 # explanatory text` still counts. Pinned as `comment-sleep-after-command`, `comment-python-sleep-in-heredoc`, `comment-after-real-sleep-still-counts`, `comment-sleep-inside-word-counts`, `loop-comment-sleep-only`; repo output byte-identical to 08af7c0.

scripts/lint-fixed-sleeps.sh
```diff
@@ -0,0 +1,408 @@
...
+    }
+    if (in_dq) {
+      if (c == "\\") { out = out "  "; i++; continue }
+      if (c == "\"") { in_dq = 0; out = out c } else out = out " "
+      continue
+    }
+    if (c == "\\") { out = out "  "; i++; continue }
+    if (c == "'") { in_sq = 1; out = out c; continue }
+    if (c == "\"") { in_dq = 1; out = out c; continue }
+    if (c == "#") { comment_at = i; break }
```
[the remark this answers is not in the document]
panghy: Fixed in f93ad1a. An unquoted `#` now starts a comment only at the start of a word — first on the line or after a blank or one of `; & | ( ) { }` — in `mask()`, which every `#` consumer (marker lookup, loop-keyword tokeniser, heredoc openers, and from 014c1a1 the sleep scan) goes through. `sleep 1; printf '%s\n' foo#timing-guard: reason` is now an unmarked finding; `sleep 1 #timing-guard: reason` and `sleep 1;# timing-guard: reason` still mark. Pinned as `marker-hash-inside-word`, `marker-hash-no-space`, `marker-hash-after-semicolon`, `marker-hash-after-brace`, `malformed-hash-inside-word-is-not-a-marker`; repo output byte-identical to 08af7c0.

scripts/lint-fixed-sleeps.sh
```diff
@@ -0,0 +1,456 @@
...
+}
+
+# Append the delimiter of every heredoc operator on the line (`<<` / `<<-`,
+# never the `<<<` here-string) to the hd_delim[] / hd_strip[] queue. The
+# operator is located on the masked line so quoted or commented `<<` never
+# opens one; the delimiter word is read from the raw line so `<<'EOF'`,
+# `<<"EOF"` and `<<\EOF` keep their text once the quoting is dropped.
+function heredoc_openers(masked, raw,   len, i, j, strip, word) {
+  len = length(masked)
+  for (i = 1; i < len; i++) {
```
panghy, line 237: **[P2] Do not enqueue arithmetic shifts as heredoc openers**

The new `heredoc_openers` scanner treats every unquoted `<<` as a redirection, including the left-shift operator inside `$((...))` and `((...))`. It queues a delimiter that never appears and suppresses all subsequent loop-keyword tracking through EOF. This is a regression from 08af7c0: with the following `scripts/x.test.sh` and an empty baseline, 014c1a1 exits 0 with no output instead of reporting the unmarked fixed-count header on line 2.

<augment_code_snippet path="scripts/x.test.sh" mode="EXCERPT">
````bash
flags=$((1 << 2))
for i in {1..3}; do
  :
  sleep .1 # timing-guard: poll interval
done
````
</augment_code_snippet>

Replacing the first line with `((flags = 1 << 2))` reproduces the same failure. Both fixtures parse and silently pass the lint under Bash 5.2.21 and Bash 3.2.57; the previous lint reports one site. Please distinguish arithmetic context from an actual heredoc redirection before enqueueing, and pin both expansion and arithmetic-command forms. Re-verify that actual heredocs, here-strings, and the existing four review fixtures still work.

File: scripts/lint-fixed-sleeps.sh [hunk 7]
```diff
@@ -0,0 +1,456 @@
...
+    if (substr(masked, i, 2) != "<<") continue
+    if (substr(masked, i + 2, 1) == "<") { i += 2; continue }
+    i += 2
+    strip = 0
+    if (substr(masked, i, 1) == "-") { strip = 1; i++ }
+    while (i <= len && substr(masked, i, 1) ~ /[ \t]/) i++
+    j = i
+    while (j <= length(raw) && substr(raw, j, 1) !~ /[ \t;&|<>()]/) j++
+    word = substr(raw, i, j - i)
```
panghy: **[P2] Read the complete quoted heredoc delimiter before quote removal**

The delimiter scan stops at whitespace/metacharacters before considering quoting. A valid `<<'END CODE'` delimiter is therefore stored as `END`, so its real `END CODE` terminator never clears heredoc state. Every later fixed-count loop is silently ignored. This fixture parses under Bash 5.2 and 3.2 and should report exactly line 4 with an empty baseline; 014c1a1 returns 0 with no output, whereas 08af7c0 reports the header:

<augment_code_snippet path="scripts/x.test.sh" mode="EXCERPT">
````bash
cat <<'END CODE'
done
END CODE
for i in {1..3}; do
  :
  sleep .1 # timing-guard: poll interval
done
````
</augment_code_snippet>

Please consume the delimiter as a shell word with quote/backslash awareness before applying quote removal, rather than splitting the raw word on spaces and deleting every quote/backslash. Add the single-quoted and double-quoted space-containing delimiter cases (and an escaped-space guard) so the parser reliably resumes normal loop tracking after the terminator. This is within the follow-up's explicitly supported quoted/unquoted delimiter contract.

**Review by panghy — COMMENTED** [review]
## Follow-up review at 014c1a1: changes still needed

The two original blocking findings are fixed: independently replaying all four exact function/Python fixtures now reports the expected sites. Generated shell heredoc sleeps remain detected, and the comment-handling fixes match the updated task contract. The 110-case self-test passes under Bash 5.2.21 and real Bash 3.2.57 with identical transcripts; the clean lint, `make lint-shell-sleeps`, `make docs-check`, and whitespace check pass.

However, the new heredoc state machine introduces **two new blocking regressions**:

1. [Arithmetic left shifts are mistaken for heredoc redirections](https://github.com/intent-hq/intent/pull/5433#discussion_r4052309594), suppressing all later loop headers. Reproduced for both `$((1 << 2))` and `((flags = 1 << 2))`.
2. [Quoted delimiter words are truncated at whitespace](https://github.com/intent-hq/intent/pull/5433#discussion_r4052309595), so the actual terminator never clears heredoc state. Reproduced with `<<'END CODE'`.

Exact source fixtures and expected finding lines are in the inline comments. Both new failures reproduce under Bash 5.2 and Bash 3.2; the prior 08af7c0 lint reports the header in each case, while 014c1a1 exits 0 with no output. Please pin and fix these narrow lexical boundaries before approval. No repository files were edited during this review. The previously noted case-pattern/line-continuation limitations remain non-blocking scope notes; no full Bash parser is requested.

File: scripts/lint-fixed-sleeps.sh (hunk 6 above)
Comment (panghy) [reply scripts/lint-fixed-sleeps.sh:237]: Fixed in f4b13c4. `heredoc_openers` now tracks parens on the masked line: a `(` immediately after another `(` opens an arithmetic level (parens nest inside it), and `<<` is skipped as a shift while any such level is open, so `$((1 << 2))`, `((flags = 1 << 2))`, `$((y<<3))` and `$(( (y<<3) + (1 << 2) ))` no longer queue a delimiter. A real heredoc after the arithmetic on the same line (`x=$((1 << 2)); cat <<EOF`) still opens.

Pinned as `arith-shift-expansion-before-loop`, `arith-shift-statement-before-loop`, `arith-shift-no-space`, `arith-shift-nested-parens` and `arith-then-heredoc-same-line` in `lint-fixed-sleeps.test.sh`; all five sum to 0 against the 014c1a1 lint and report line 2 (line 4 for the last) now. Here-string, quoted-operator and the existing heredoc fixtures are unchanged; repo lint output is byte-identical; the self-test transcript is identical under bash 5.2.21 and 3.2.57.

File: scripts/lint-fixed-sleeps.sh (hunk 7 above)
Comment (panghy) [reply scripts/lint-fixed-sleeps.sh]: Fixed in f4b13c4. The delimiter is now consumed from the raw line as one shell word before quote removal: a `'…'` span is copied verbatim, a `"…"` span honours `\"` `\\` `\$` `` \` ``, a bare `\` escapes the next character, and the word ends only at an unquoted blank or metacharacter. `<<'END CODE'`, `<<"END CODE"`, `<<-'END CODE'` and `<<END\ CODE` all store `END CODE`, so the real terminator line clears heredoc state and the later fixed-count header is reported.

While pinning this I found `<<\EOF` was also broken at 014c1a1 (the blank-skip after `<<` ran on the masked line, where `mask()` blanks the `\E` pair, so the delimiter was read as `OF`); the skip now runs on the raw line and `heredoc-backslash-delimiter` pins it.

Fixtures: `heredoc-quoted-delimiter-space` (yours, verbatim), `heredoc-double-quoted-delimiter-space`, `heredoc-dash-quoted-delimiter-space`, `heredoc-escaped-space-delimiter`, `heredoc-backslash-delimiter`; each sums to 0 against the 014c1a1 lint and reports `scripts/x.test.sh:4:` now. Self-test transcript identical under bash 5.2.21 and 3.2.57; repo lint output byte-identical to 014c1a1.

File: scripts/lint-fixed-sleeps.sh [hunk 8]
```diff
@@ -0,0 +1,493 @@
...
+# opens one, and neither does the left shift inside an arithmetic `$((…))` /
+# `((…))`: a `(` right after another `(` opens an arithmetic level (parens
+# nest inside it), and `<<` is an operator while any such level is open. The
+# delimiter is read from the raw line as one shell word — `'…'` / `"…"`
+# spans and `\`-escapes included, so `<<'END CODE'` and `<<END\ CODE` name
+# `END CODE` — and stored with the quoting removed, as the shell compares it.
+function heredoc_openers(masked, raw,   len, rlen, i, j, c, pd, na, arith, strip, word) {
+  len = length(masked)
+  rlen = length(raw)
+  pd = 0; na = 0
```
panghy, line 235: **[P2] Arithmetic-context recognition still resets before a multiline shift**

The reported single-line forms are fixed by f4b13c4, but `pd = 0; na = 0` resets arithmetic context on every physical line. With an arithmetic expression formatted across lines, the following `<<` is still enqueued as a heredoc opener and loop tracking stays disabled through EOF. Exact fixture, with an empty baseline:

<augment_code_snippet path="scripts/x.test.sh" mode="EXCERPT">
````bash
flags=$((
  1 << 2
))
for i in {1..3}; do
  :
  sleep .1 # timing-guard: poll interval
done
````
</augment_code_snippet>

Expected: one finding on the unmarked fixed-count header at line 4. Actual at f4b13c4: the lint exits 0 with empty stdout/stderr under both Bash 5.2.21 and Bash 3.2.57; both accept the fixture with `-n`. The arithmetic-command form (`((` / `flags = 1 << 2` / `))` on its first three lines) has the same result. This remains part of the arithmetic-context failure, not a new feature request.

The ten newly pinned single-line/quoted-delimiter cases all fail against 014c1a1 and pass now, and the 120-case cross-shell suite is green. Per the coordinator's agreed stop point, I am **not requesting another parser extension automatically**: this needs the planned decision to keep/drop rule 3 or explicitly accept and document a narrower guarantee. Until then I cannot approve the stated arithmetic-context contract.

panghy (commented):
## Re-review at f4b13c4: scope decision required

The reported single-line arithmetic-shift and quoted-delimiter fixtures are fixed. I independently replayed all ten newly added cases: each fails on 014c1a1 and passes at this head. The complete 120-case self-test passes under Bash 5.2.21 and Bash 3.2.57 with identical transcripts; plain lint, `make lint-shell-sleeps`, `make docs-check`, and whitespace checks also pass.

**One blocking form of the arithmetic-context issue remains:** [multiline arithmetic resets the scanner state](https://github.com/intent-hq/intent/pull/5433#discussion_r4052346698). A valid multiline `$((...))` or `((...))` expression containing `<<` still suppresses all later fixed-count-loop headers. Both interpreters reproduce a silent lint pass where one header finding is required.

As agreed with the coordinator, this is the stop point for further automatic parser work. Please decide whether to keep rule 3, remove it, or explicitly narrow/document its guarantee; I have not asked the implementor for another extension. The exact reproduction is in the inline comment. Previous fixes are confirmed, but the current stated contract is **not approved**. No repository edits, pushes, or merge actions were performed during re-review.

scripts/lint-fixed-sleeps.sh
panghy: Accepted as a documented limitation (user decision: keep rule 3, narrow the stated guarantee rather than extend the parser again). Landed in 88ee83b:

- The lint header now carries a "Known limitations of rule 3" list: tokenisation is line-local, so a multi-line `$(( … ))` / `(( … ))` whose `<<` sits on a continuation line is taken as a heredoc opener, and the tracker may then miss or misattribute loop findings for the rest of the file (later headers missed; a `done` inside the mistaken region leaves its loop open, so a later marked sleep is reported against the earlier header). The `case` pattern `done)` early-close is listed alongside. Workaround stated: restructure the expression onto one line or add a `# timing-guard:` marker. `docs/fe/DEVELOPER_GUIDE.md` points at that list.
- Both fixtures from this thread (`shift-multiline-expansion`, `shift-multiline-command`), the `case` `done)` shape, and the in-loop misattribution repro are pinned verbatim in `scripts/lint-fixed-sleeps.test.sh` as a labelled known-limitations group at their current output (0 / 0 / 0 findings; 1 finding on line 1 for the misattribution case), with a comment that a parser change flipping any of them must update both the expectation and the documented list. Harness count 120 → 124.
- Repo lint output is byte-identical to f4b13c4 (exit 0, counts `bootstrap-dev-host 1` / `dev-sandbox 3`); the self-test passes under Bash 5.2.21 and 3.2.57 with identical transcripts; `make lint-shell-sleeps` and `make docs-check` pass.

Resolving with the narrowed contract in place.

panghy (commented):
## Final review: APPROVED against the accepted narrowed contract

Reviewed pushed head `88ee83b2462b4a56ed7aa27a4601d536003df4a5`. **No unaddressed blocking findings; no new non-blocking findings.** This COMMENT conveys the review verdict because the authenticated account is also the PR author; it is not merge authorization.

The user explicitly chose to keep rule 3 with documented line-local limitations. The header now accurately says these shapes may **miss or misattribute loop findings for the rest of the file**; it does not claim that false positives are impossible. The developer guide points to the limitation list. This accepted narrower contract resolves my remaining arithmetic-context objection; it does not mean the parser now handles multiline arithmetic or case patterns correctly.

### Independently verified
- `88ee83b` changes only the lint header, the guide paragraph, and known-limitation fixture rows/count. The executable lint (everything from `set -euo pipefail` onward) is byte-identical to f4b13c4.
- All four newly pinned fixtures parse under both Bash 5.2.21 and Bash 3.2.57 and reproduce the documented current outputs: 0 / 0 / 0 findings for the two multiline-arithmetic and case-pattern fixtures, and 1 finding at line 1 for the misattribution fixture. The group clearly distinguishes current from desired behavior and requires updating both docs and expectations when the parser changes.
- Repository stdout, stderr, exit status, and counts are byte-identical to f4b13c4 under both shells; baseline remains bootstrap-dev-host 1 / dev-sandbox 3.
- Full 124-case self-test passes under both shells with byte-identical transcripts; `make lint-shell-sleeps`, `make docs-check`, and `git diff --check f4b13c4..HEAD` pass. Both hosted `shell-tests` and `shell-tests-bash3` are green at this head.
- The decision and commit are recorded in the remaining review thread, and all review threads are resolved. The original function/heredoc and single-line arithmetic/quoted-delimiter regressions remain covered by passing fixtures.

Previously reviewed Makefile/CI wiring, baseline diagnostics, marker-only suite edits, and conventional commit hygiene remain unchanged. Full Rust-backed `make check` was not executed in this review, consistent with the recorded local review-gate deferral; this verdict does not claim that command ran. No repository edits, pushes, or merge actions were performed by the reviewer.

PR #5488, opened 2026-09-20
ci: add protocol-to-FE field parity check (check-protocol-field-parity)

panghy:
## Summary

Adds a monorepo gate that fails when a wire field emitted by an intentd serde row struct is absent from the cloudlands-fe TypeScript type that consumes it, so a silently dropped protocol field (like parentAgentId on AgentLite, cloudlands-fe#2687) is caught before merge instead of by a review bot.

- New scripts/check-protocol-field-parity.mjs (+ node:test suite, 50 tests). Manifest pairs seeded: AgentLite -> AgentSession (src/shared/types/agent-session.ts) and Workspace -> Workspace (src/shared/types.ts). Per-pair ignore entries carry a one-line reason; a stale entry (field no longer emitted, or now present in the FE type) fails the check.
- Rust side honors rename_all = camelCase, per-field rename (including rename(serialize = ...)), skip / skip_serializing, and expands #[serde(flatten)] for same-file structs (Workspace flattens membership). Multi-line attributes are joined before parsing; comments and string-literal contents (including escaped braces) are stripped before brace-depth counting.
- Wired into make check (target check-protocol-field-parity, next to check-makefile-targets) and the docs-check CI job; documented in docs/README.md (Wire Contract section). Uninitialized submodule -> skip notice, exit 0.

Enforcement rung: 2 (mechanical pre-merge check). Rung 1 (shared/generated types) is out of scope.

## Known FE gaps recorded as ignore entries (not fixed here)

AgentLite -> AgentSession: contextReferences and fileBlocks (read untyped via the agent-session slice; accepted typing debt) and contextUsage (not consumed by the FE today). Workspace -> Workspace: tokenUsage plus the flattened membership fields. See the manifest reasons in the script.

## Verification

- node --test scripts/check-protocol-field-parity.test.mjs (24/24), node --test scripts/check-makefile-targets.test.mjs (39/39)
- node scripts/check-protocol-field-parity.mjs and make check-protocol-field-parity exit 0 at current pins; make docs-check and node scripts/check-doc-links.mjs pass; make -n check lists the target
- Live negative checks: deleting parentAgentId from the FE AgentSession interface and adding a fake field to AgentLite each exit 1 naming the field and both files; a stale ignore entry exits 1 naming the field and the affected file (submodules restored afterwards)
- Verifier passes found and fixed (with failing-first regression tests): string braces, multi-line serde attrs, prototype-named ignore lookup, escaped-brace leak. PR review (7 comments) then drove a hardening pass: Rust line/block comments stripped before parsing; TS declaration lookup and key collection carry comment/template/paren/bracket state (method parameters, tuple labels, commented-out declarations and generic-bound bodies no longer count as row keys); directional rename(serialize = ...) supported; cfg_attr, container into/try_into/remote and any unrecognized serde form fail with an actionable unsupported-attribute error; a stale manifest path with an initialized submodule fails closed. The reviewer's nine verbatim repro fixtures are a table-driven regression test.
- Known unsupported TS syntax (documented, not resolved): interface extends and computed keys
- No packages/* gitlink is moved relative to the merge base


scripts/check-protocol-field-parity.mjs
```diff
@@ -0,0 +1,389 @@
...
+  const re = TS_BLOCK_RE(typeName);
+  const idx = lines.findIndex((l) => re.test(l));
+  if (idx === -1) return null;
+  const keys = [];
+  const state = { block: false, template: false };
+  let depth = 0;
+  let endLine = lines.length;
+  for (let i = idx; i < lines.length; i += 1) {
+    const clean = stripNoise(lines[i], state);
+    const body = i === idx ? clean.slice(clean.indexOf('{')) : clean;
```
panghy: [P2] Locate the interface body after generic constraints

The first `{` on the declaration line can belong to a generic bound, not the interface body. With `export interface Row<T extends {` on the first line, `parentAgentId: string;` on the next, then `}> {` and `id: string;`, extractTsKeys returns both parentAgentId and id. A Rust Row that emits parentAgentId consequently passes even though Row<T> itself has only id. This is a separate header/body boundary from nested member values: either locate the actual body past balanced generic parameters or reject generic headers explicitly. Add a regression that a bound's properties cannot satisfy row fields.

scripts/check-protocol-field-parity.mjs
```diff
@@ -0,0 +1,389 @@
...
+/**
+ * Collect the top-level property names of `export interface <typeName> {` or
+ * `export const <typeName> = z.object({` in `source`. Returns
+ * `{ keys: [{ name, line }], startLine, endLine }`, or `null` when the block
+ * is not found.
+ */
+export function extractTsKeys(source, typeName) {
+  const lines = source.split('\n');
+  const re = TS_BLOCK_RE(typeName);
+  const idx = lines.findIndex((l) => re.test(l));
```
panghy: [P2] Find declarations outside comments and template literals

The declaration search runs on raw source, before stripNoise has any lexical state. A block comment containing an old multiline `export interface Row { ... }` above the live interface is selected as the real contract. I also reproduced this with the actual AgentSession sources: prepend a single valid block-commented copy of its property list, remove parentAgentId only from the live interface, and comparePair still returns `[]`. Starting stripNoise at the matched line loses the opening comment. Please carry comment/template state from the start of the file while locating declarations, on both the TS and Rust sides, and add a fixture where a documented/disabled declaration must not mask the live missing field.

scripts/check-protocol-field-parity.mjs
```diff
@@ -0,0 +1,389 @@
...
+    errors.push({ file, line: structLine, message: `struct ${structName} has ${have}; only ${Object.keys(SUPPORTED_RENAME_ALL).join(', ')} is supported` });
+    return { fields: [], structLine, errors };
+  }
+
+  const fields = [];
+  let attrs = [];
+  for (let i = idx + 1; i < lines.length; i += 1) {
+    const raw = lines[i].text;
+    const l = raw.trim();
+    if (l === '}') break;
```
panghy, line 319: [P2] Strip Rust comments before interpreting fields and attributes

The Rust walk consumes raw lines, so valid block comments can change the emitted-field set. I reproduced this against the real AgentLite/AgentSession sources in memory: insert a block comment whose middle line is `#[serde(skip)]`, then add `pub reviewer_new_field: String,` immediately after it. `comparePair` returns `[]` even though that new field is serialized and absent from AgentSession. A standalone `}` inside a block comment also terminates the struct walk early. Separately, a trailing `// wire name` after `#[serde(rename = "parentAgentId")]` causes the rename to be discarded by the anchored attribute regex. Please lex/strip Rust comments (including nested block comments) while preserving line numbers before declaration/attribute/field parsing, and add comparePair regressions proving these comments cannot suppress a missing-field diagnostic.

scripts/check-protocol-field-parity.mjs
```diff
@@ -0,0 +1,389 @@
...
+  const idx = lines.findIndex((l) => re.test(l));
+  if (idx === -1) return null;
+  const keys = [];
+  const state = { block: false, template: false };
+  let depth = 0;
+  let endLine = lines.length;
+  for (let i = idx; i < lines.length; i += 1) {
+    const clean = stripNoise(lines[i], state);
+    const body = i === idx ? clean.slice(clean.indexOf('{')) : clean;
+    if (i > idx && depth === 1) {
```
panghy: [P2] Track parameter and tuple scopes before collecting TS keys

Brace depth 1 does not mean the start of an interface member. Removing the real AgentSession.parentAgentId property and adding a multiline `describe(...)` method whose parameter is `parentAgentId: string` makes `comparePair` return `[]`: the parameter is promoted into `tsKeys`. The same false pass occurs for a multiline labelled tuple member, with `parentAgentId: string` inside the tuple. These are valid TypeScript but neither declares the row property. Please track parentheses/brackets and member boundaries (or fail closed on unsupported forms), and add comparison-level regressions for both forms. A nested name must not satisfy AgentLite.parentAgentId.

scripts/check-protocol-field-parity.mjs
```diff
@@ -0,0 +1,389 @@
...
+  }
+  if (cur.trim()) args.push(cur.trim());
+  return args;
+}
+
+function parseSerdeAttrs(attrs) {
+  const out = { renameAll: null, rename: null, skip: false, flatten: false };
+  for (const attr of attrs) {
+    const args = serdeArgs(attr);
+    if (!args) continue;
```
panghy: [P2] Reject unsupported serialization renames instead of silently guessing

Some serde syntax that changes emitted names is silently treated as no attribute. For a camelCase Row, `#[cfg_attr(all(), serde(rename = "parentAgentId"))]` on `owner_id` is unconditionally active, but the scanner reports ownerId; an FE type containing only ownerId therefore passes. `#[serde(rename(serialize = "parentAgentId"))]` has the same result. Both probes return no comparison errors even though parentAgentId is the emitted wire name. Supporting every cfg predicate is unnecessary here: at minimum fail with an actionable unsupported-attribute error for wire-affecting syntax that the scanner cannot interpret, rather than asserting parity. Please cover these forms with negative tests.

panghy (commented):
## Independent review at 2bfacd2

Comment-only review, as requested; no approval/request-changes action and no source edits. Five anchored correctness findings are posted.

### Blocking correctness gaps
- Rust block comments can suppress real emitted fields or terminate the struct walk.
- Multiline TypeScript method parameters and tuple labels can satisfy missing top-level row properties.
- A commented-out TypeScript declaration can be selected instead of the live contract.

These are reproduced false passes, including probes against the actual AgentLite/AgentSession sources in memory. The existing suite does not catch them.

### Should fix
- Reject or interpret serialization-affecting cfg_attr / directional serde rename syntax instead of silently assuming camelCase.
- Find the real interface body past generic bounds; bound properties currently count as row properties.

### Validation
- node --test scripts/check-protocol-field-parity.test.mjs scripts/check-makefile-targets.test.mjs: 63/63 pass.
- make check-protocol-field-parity docs-check, node scripts/check-doc-links.mjs, make -n check, and git diff --check: pass locally.
- Basic parentAgentId-removal/new-field negative probes and both stale-ignore forms: expected errors.
- Three independent missing-field regression assertions: FAIL, demonstrating the blocking false passes.
- No packages/* gitlinks in the PR's merge-base diff; frontend/intentd worktrees untouched.
- Hosted docs-check currently fails at the existing make docs-check step, so the new parity steps were skipped; event-catalog-check also fails. Those failures need separate investigation before calling hosted CI verified.

### Non-blocking limitations / wording
- Interface inheritance and literal computed keys are not resolved; they produce false missing-field reports. Please document the supported direct-member syntax, or reject unsupported syntax with a specific diagnostic. The index signature correctly does not invent explicit named fields; ordinary tuple/enum-valued Rust fields parse.
- PR-body nit: contextUsage has no FE reader at these pins; only contextReferences/fileBlocks are read untyped through the session slice. The current grouped parenthesis implies all three are read that way.
- PR-body nit: stale-ignore errors name the affected source file, not both files, unlike the missing-field diagnostic. Narrow the verification claim accordingly.

The Makefile/CI/README placement is consistent with neighboring checks. The manifest's individual ignore reasons match the inspected frontend source. Please add comparison-level regression tests for the findings before re-review.

scripts/check-protocol-field-parity.mjs
```diff
@@ -0,0 +1,389 @@
...
+    for (const arg of args) {
```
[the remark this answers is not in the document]
panghy: Fixed in abc802e: `parseSerdeAttrs` accepts `rename(serialize = "x")`, `rename(serialize = "x", deserialize = "y")` in either order (top-level comma split is now paren-aware), and `rename_all(serialize = ...)` at struct level; a `rename(deserialize = ...)`-only attribute leaves the wire name as the camelCased Rust name. The regression test covers all four forms.

scripts/check-protocol-field-parity.mjs
```diff
@@ -0,0 +1,389 @@
...
+
+function submoduleOf(file) {
+  const m = file.match(/^(packages\/[^/]+)\//);
+  return m ? m[1] : null;
+}
+
+async function readIfAvailable(root, file) {
+  const sub = submoduleOf(file);
+  if (sub && !(await exists(path.join(root, sub, '.git')))) return null;
+  if (!(await exists(path.join(root, file)))) return null;
```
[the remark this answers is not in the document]
panghy: Fixed in abc802e: `readManifestFile` returns skipped only when the submodule has no `.git` entry; a missing path inside an initialized submodule (or outside any submodule) is an error `manifest path <file> does not exist (submodule <sub> is initialized); update PAIRS ...` at line 1, and the pair is neither checked nor skipped. Tests cover both branches; verified live by pointing a pair at a non-existent FE path.

scripts/check-protocol-field-parity.mjs
panghy: Fixed in abc802e: the declaration regex no longer requires `{` on the declaration line; `findTsBodyOpen` scans from the interface name past balanced `<...>` (skipping `=>`) and the `extends` type arguments to the first brace at angle depth 0, possibly lines later. Regression: `export interface Row<T extends {` / `parentAgentId: string;` / `}> {` / `id: string;` yields keys `[id]` with range 1-5 and comparePair reports `parentAgentId`; a one-line `Row<T extends () => void> extends Base<{ parentAgentId: string }> {` header is covered too.

scripts/check-protocol-field-parity.mjs
panghy: Fixed in abc802e: `extractTsKeys` now cleans every line from the start of the file with the comment/template state and matches the declaration on the cleaned text; the Rust side is covered by the same-commit comment stripper. Regressions: a block-commented copy of the TS interface (plus a template-literal copy) above the live one, with `parentAgentId` removed only from the live interface, fails naming the live block range; a block-commented copy of the Rust struct does not mask the live struct.

scripts/check-protocol-field-parity.mjs
panghy: Fixed in abc802e: Rust source is passed through `stripRustComments` (nested `/* */`, `//`, string/char literals respected, line numbers preserved) before `logicalLines`/declaration/attribute/field parsing. Regressions (comparePair-level): a block-commented `#[serde(skip)]` followed by `reviewer_new_field` is reported; a standalone `}` in a (nested) block comment does not end the walk; `#[serde(rename = ...)] // wire name` keeps the rename.

scripts/check-protocol-field-parity.mjs
panghy: Fixed in abc802e: key detection now requires brace depth 1 with paren and bracket depth 0 at the start of the line; `stripNoise` also drops `()[]` inside string literals. Regressions (comparePair-level): a multiline `describe(parentAgentId: string, ...)` method and a multiline `(parentAgentId: string) => void` function member, and a labelled tuple `[parentAgentId: string, ...]`, all still report `Row.parentAgentId` missing.

scripts/check-protocol-field-parity.mjs
panghy: Fixed in abc802e: attributes are now interpreted against an allowlist. `#[cfg_attr(...)]` and `#[cfg(...)]` on a field or struct, and any `#[serde(...)]` argument outside the known-harmless set (`default`, `skip_serializing_if`, `deserialize_with`, `alias`, ...) or the supported wire-affecting set, produce `file:line: error: unsupported attribute <attr> on Struct.field: ...` (exit 1) instead of asserting parity. Negative tests: `cfg_attr(all(), serde(rename = ...))`, `cfg(feature = ...)`, `rename(bogus = ...)`, `rename_all(serialize = ...)` on a field, an unknown `getter_of_wire_name = ...`, and struct-level `tag = "kind"`; a positive test pins the harmless arguments used in model.rs.

panghy: Correction to my reply above — it overstated the breadth of rejection. The precise behavior as of c94e240 (which also closes two gaps the verifier found at 875c8c5):

**Rejected with `unsupported attribute <attr> on <target>: ...` (exit 1):**
- `#[cfg(...)]` and `#[cfg_attr(...)]` on a struct or field, in any whitespace form (`# [cfg_attr (...)]` included).
- Struct-level `into`, `try_into`, `remote` (the struct's own fields are not the wire shape). `into` and `remote` were wrongly allowlisted at abc802e; fixed in c94e240.
- Any `#[serde(...)]` argument not in the lists below, e.g. `tag`, `content`, `untagged`, `transparent`, `rename(bogus = ...)`, or an unknown key.
- Any `serde` attribute whose body does not parse as `serde(<args>)`.

**Rejected with a different message:** a struct whose only `rename_all` is `rename_all(deserialize = ...)` has no outgoing rename and fails with `struct Row has no #[serde(rename_all)]; only camelCase is supported` — the same error as a struct with no `rename_all` or a non-camelCase one.

**Interpreted (wire-affecting, understood):** struct `rename_all = "camelCase"` / `rename_all(serialize = "camelCase")`; field `rename = ".."`, `rename(serialize = ".." [, deserialize = ".."])` in either order, `skip`, `skip_serializing`, `flatten` (same-file target). A `rename(deserialize = ..)`-only field keeps the camelCased Rust name.

**Accepted as harmless (no effect on emitted names):**
- struct: `default`, `bound`, `deny_unknown_fields`, `crate`, `from`, `try_from`, `expecting`, `rename` (container rename does not affect field names).
- field: `default`, `bound`, `skip_serializing_if`, `skip_deserializing`, `deserialize_with`, `serialize_with`, `with`, `alias`, `borrow`, `getter`.
Each is accepted as a bare word or `key = "string"` only.

Whitespace between `#`, `[`, the path and `(` is normalized (c94e240), so `# [serde (rename = "x")]` is interpreted like `#[serde(rename = "x")]` rather than ignored.

Test coverage: comparison-level regressions exercise `cfg` / `cfg_attr` (plain and spaced), unknown and `tag` arguments, `rename(bogus = ..)`, field-level `rename_all(serialize = ..)`, the shape-replacing containers `into` / `try_into` / `remote`, and the spaced attribute forms, plus a positive test pinning the harmless arguments used in model.rs. `content` / `untagged` / `transparent` and unparsable `serde` bodies fall through the same default branch but are not individually exercised by a test.

PR #5476, opened 2026-09-20
ci: run CI on merge_group and gate the queue on a CI Gate job

panghy:
## Why

The monorepo merge queue currently lands entries without running any check: ci.yml only triggers on pull_request and the main ruleset has no required status checks. Two PRs that are each green on their own branch can therefore merge into a combination that fails CI, and the failure only surfaces on the next unrelated PR.

Incident: intent-hq/intent#5433 and intent-hq/intent#5446 were each green alone; together they broke the shell-tests job on every subsequent PR (intent-hq/intent#5455) for ~14 h until intent-hq/intent#5468.

## What

- .github/workflows/ci.yml: add the merge_group (checks_requested) trigger so the queue runs CI on the actual merged tree, and add an aggregate, event-aware gate job named CI Gate (same pattern as intentd). It needs every other job, runs with always(), and fails unless the always-on jobs succeed and the PR-only jobs (pr-title, breaking-token, submodule-pins) succeed on pull_request or are skipped on merge_group. No push trigger: the merge-group run already tests the identical tree.
- AGENTS.md: the Merging bullet now describes the required CI Gate on all three repos instead of saying the monorepo queue lands entries without a check run (compressed, 432 -> 398 words).

## Follow-up after merge

The required_status_checks rule (context CI Gate, GitHub Actions integration) is added to ruleset Default (18861801) immediately after this PR lands, so no open PR is stranded without the check. Then a throwaway PR is enqueued to confirm the merge_group run happens and the auto/submodule-bump PR still lands unattended.

## Verification

- PyYAML parse of the workflow; guard grep on github.event.pull_request / head_ref / base_ref; 159 simulated result combinations against the extracted gate script (failure/cancelled/skipped/unknown per event) all behave as designed.
- make docs-check and node scripts/check-doc-links.mjs pass.
- No packages/* gitlink moved.

Issue #5455, opened 2026-09-19
ci: shell-tests fails on main — lint-fixed-sleeps flags 5 unannotated sleeps in scripts/dev-sandbox.test.sh


panghy:
Since #5446 (merged 2026-09-19T13:12Z) landed on top of the fixed-sleep lint from #5433, every monorepo CI run fails the shell-tests job: bash scripts/lint-fixed-sleeps.sh exits 1 on scripts/dev-sandbox.test.sh lines 332, 420, 456, 561, 722 (baseline scripts/fixed-sleep-baseline.txt allows 3). Both PRs passed CI on their own branches; the combination on main is what fails.

Evidence:
- auto/submodule-bump runs at 13:23, 13:25, 13:30 UTC: shell-tests=failure (the 10:40 run was success).
- intent-hq/intent#5453 runs 35446221431 and 35447150290: shell-tests=failure with the lint output above; the PR does not touch any of the lint's inputs, and they are byte-identical to origin/main.

Fix options: annotate the five sleeps per the lint's convention (preferred) or extend the baseline. Rung-2 follow-up worth considering: the merge queue runs merge_group CI in intentd/cloudlands-fe but the monorepo ruleset has no required status checks, so two independently-green PRs can combine into a red main.

PR #5507, opened 2026-09-20
ci: detect main branch ruleset drift against committed rules

panghy:
## Summary

The expected main-branch rules for intent, intentd and cloudlands-fe are now committed under .github/rulesets/<repo>.main.json, and monorepo CI fails when the live effective rules diverge (required check missing, thread resolution off, merge-queue settings changed). A silent ruleset edit can no longer let the merge queue land untested trees.

Motivation: the monorepo queue recently ran without a required status check while the expected rules existed only as prose in AGENTS.md (intent-hq/intent#5476, intent-hq/intent#5455; see also intent-hq/intent#4959).

## What changed

- scripts/check-rulesets.mjs fetches GET /repos/intent-hq/<repo>/rules/branches/main for all three repos (public; the default GITHUB_TOKEN suffices, no new secret), normalizes both sides (drops ruleset_id, sorts keys/arrays) and diffs against the committed snapshot. Exit 1 with a diff naming the divergent field/check; 404/401 exit 2; transient API failures (network, 5xx, rate limit) print a warning and exit 0. UPDATE=1 regenerates the snapshots (make check-rulesets UPDATE=1). For the monorepo it also cross-checks that every required check context matches a job name in ci.yml, so renaming the CI Gate job cannot leave the queue waiting on a check nothing produces.
- scripts/check-rulesets.test.mjs: 20 node:test cases, including regressions for the two comparison false negatives found in review (same-source rules of the same type, and repeated check contexts with different integration ids) plus an empty-diff safety net.
- .github/workflows/ci.yml: always-on ruleset-check job on pull_request and merge_group, wired into the CI Gate (ci-gate.sh / ci-gate.test.sh).
- .github/workflows/ruleset-drift.yml: daily schedule + workflow_dispatch running the same check.
- AGENTS.md § Merging now points at the committed rulesets and the UPDATE=1 path instead of restating the parameters (same line count, fewer words; #4959 guidance kept).

## Verification

- node --test scripts/check-rulesets.test.mjs: 20/20 pass.
- node scripts/check-rulesets.mjs: exit 0, all three repos match live.
- Removing CI Gate / flipping required_review_thread_resolution / changing a merge_queue parameter / renaming the gate job each exit 1 naming the field.
- make check-rulesets UPDATE=1 then make check-rulesets: exit 0, no snapshot diff.
- bash scripts/ci-gate.test.sh, make docs-check, check-doc-links, check-root-hygiene: pass.
- Independent verifier pass approved after two fix rounds.


scripts/check-rulesets.mjs
```diff
@@ -0,0 +1,456 @@
...
+    super(message);
+    this.transient = transient;
+  }
+}
+
+function rateLimited(response) {
+  return (
+    (response.status === 403 || response.status === 429) &&
+    (response.headers.get('x-ratelimit-remaining') === '0' || response.headers.has('retry-after'))
+  );
```
panghy: [P2] Treat secondary rate limits without retry headers as transient

This predicate misses a documented GitHub response: secondary rate limits can return 403 or 429 while `x-ratelimit-remaining` is nonzero and `retry-after` is absent ([GitHub documentation](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api#exceeding-the-rate-limit)). I injected both HTTP 429 with no rate headers and HTTP 403 with remaining=4000 plus the documented secondary-rate-limit message; `run()` returned 2 without a warning in both cases. Because this job is always required by CI Gate, a temporary secondary limit would fail a PR or eject a merge group, contrary to the fail-soft requirement. Treat 429 itself as transient, and recognize the secondary-limit 403 response without treating ordinary permission-denied 403s as transient. Please add regression cases without `retry-after`/remaining=0; the existing tests cover only those headers.

scripts/check-rulesets.mjs
```diff
@@ -0,0 +1,456 @@
...
+  if (!response.ok) {
+    throw new RulesFetchError(`${OWNER}/${repo}: HTTP ${response.status} reading ${rulesUrl(repo, apiBase)}`, {
+      transient: false,
+    });
+  }
+  let body;
+  try {
+    body = await response.json();
+  } catch {
+    throw new RulesFetchError(`${OWNER}/${repo}: response is not JSON`, { transient: false });
```
panghy: [P2] Preserve fail-soft behavior when the response body connection fails

`fetch()` resolves once the headers arrive; network failure can still happen while `response.json()` consumes the body. This catch labels every such failure as a non-transient configuration problem. I reproduced it with a local HTTP server returning 200, writing the start of an array, then destroying the socket before the advertised Content-Length: `run()` returned 2 with `response is not JSON` and no warning. A GitHub connection reset after headers would therefore turn the required CI Gate red, despite the explicit transient-failure acceptance criterion. Distinguish body transport failures from a complete but invalid JSON document (for example, separate body reading from JSON parsing), retain exit 2 for actual malformed JSON, and add a regression that interrupts the response after successful headers.

panghy (commented):
## Review summary

Reviewed head 5e2645b, focusing on failure paths and workflow behavior not covered by the independent acceptance-criteria verification. Two reproducible P2 findings are posted inline: secondary rate limits without retry headers and connection failures during response-body consumption both incorrectly become exit 2 rather than warning + exit 0. These violate the transient-failure criterion and can fail the required CI Gate.

## Additional evidence

- Hosted pull_request run [35497416964](https://github.com/intent-hq/intent/actions/runs/35497416964) is green, including [ruleset-check](https://github.com/intent-hq/intent/actions/runs/35497416964/job/106042863746) and CI Gate. The ruleset-check log explicitly reports a live match for all three repositories, not merely a transient skip.
- An independent 625-pair multiset comparison probe passed: reordered and repeated same-source rules preserve symmetric equality/drift detection, including different check integrations.
- Static review confirms ruleset-check runs on pull_request and merge_group, participates in gate.needs and the always-on result checks, and fails the gate if unsuccessful. No hosted merge_group run was initiated.
- The daily workflow uses read-only contents permission, and the token is passed through the environment rather than argv. No new workflow, Makefile, or AGENTS.md blocker was identified.

No source files changed, and no approval or merge was performed. Please add regression tests for the two inline findings and rerun the focused ruleset suite after fixing them.

scripts/check-rulesets.mjs
panghy: Verified in 719d99e: independent headerless 429 and secondary-limit 403 probes both return 0 with a warning; ordinary permission 403 stays 2. The focused suite passes 22/22. Resolving this finding.

scripts/check-rulesets.mjs
panghy: Reverification of 719d99e: the original interrupted HTTP 200 now correctly warns and exits 0, and complete malformed JSON still exits 2. One control regression remains: the new body read also runs before classifying definitive HTTP 401/404 errors. With a real loopback server returning either status, Content-Length 1000, a partial body and then a socket reset, run now returns 0 plus warning instead of the required configuration-error exit 2. Please classify definitive non-success statuses before consuming their irrelevant bodies (retain the special 403 body inspection for secondary limits), and add interrupted 401/404 regression cases. Keeping this thread open for that narrow follow-up; implementor notified.

panghy: Verified in 21293b4: the independent real-loopback interrupted-body probes now produce 200 -> exit 0 with warning, 401/404 -> exit 2 without warning. Headerless 429 and secondary-limit 403 still warn and pass; permission 403 and complete invalid JSON still fail with exit 2. The focused suite passes 22/22, and the current hosted ruleset-check is green in run 35497793705. Resolving this finding and its follow-up.

scripts/check-rulesets.mjs
```diff
@@ -0,0 +1,456 @@
...
+    throw new RulesFetchError(`${OWNER}/${repo}: response is not a rules array`, { transient: false });
+  }
+  return body;
+}
+
+// A fixture keeps the tests off the network: a JSON object keyed by repository
+// whose values are either a rules array (HTTP 200) or a canned response
+// `{ status, headers?, body? }`; `{ error: "..." }` simulates a network error.
+export function fetchFromFixture(fixture) {
+  return async (url) => {
```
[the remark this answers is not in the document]
panghy: Valid point, and deliberately out of scope for this PR (see the Non-goals in the plan). The check snapshots GET /repos/{owner}/{repo}/rules/branches/main, which returns the effective rule objects and omits per-ruleset bypass_actors. GitHub only returns bypass_actors to callers with write access to the ruleset, so asserting it from CI needs an admin-scoped token secret in all three repos, which this PR avoids on purpose so it can run with the default GITHUB_TOKEN. Tracked as a follow-up issue (linked below); resolving this thread.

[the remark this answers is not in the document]
panghy: Follow-up tracked in intent-hq/intent#5510.

PR #5526, opened 2026-09-21
docs(protocol): catch the method catalog up to the intentd multiplayer tip

panghy:
## Why

The rolling submodule bump #5524 is BLOCKED on `docs-check`: intentd `main` dispatches methods the protocol catalog does not name, and `check-protocol-catalog` errors on a pinned method the docs lack. Docs lead the pin, so this PR documents the intentd tip first; the bump passes once it lands.

Validated against intentd `origin/main` **7304c635** (`test(transport): bound tunnel backpressure test waits by wall clock`, #2033).

## What

**Catalog (`05-method-catalog.md`)**

- Router: `github.identityProof.create` / `.delete` (#1965), `principal.list` (#2023), `workspace.members.add` (#2025) — rows and counts.
- Fast-path: `invite.inspect` / `invite.accept` (#1963), `invite.challenge` / `invite.prove` (#1967), `workspace.invite.create`; the multiplayer fast-path paragraph rewritten for the gist identity proof.
- `invite.redeem` (removed by intentd#1969) **stays listed as an explicitly marked pin-transitional entry**: the pinned intentd (66270a0) still exports it, and the checker treats a pinned method missing from the docs as an *error* while a docs entry missing from the pin is only a *warning*. Follow-up after the bump: drop it and the counts fall from 57 / 388 to 56 / 387 — the prose says so inline.

**`methods/multiplayer.md`**

- Method table: the four `/invite` methods, `principal.list`, `workspace.members.add`, `github.identityProof.*`; shapes block updated.
- "Joining by invite" rewritten from the host-run device flow to the **gist identity proof** (challenge → publish on the guest's own daemon → prove → clean up → connect) and the **returning-guest credential join** (`invite.inspect` + `invite.accept`, credential rotated in the join transaction); bounds updated (nonce caps 8 / 256, 10 min TTL) — every claim checked against `invite_ops.rs` / `invite.rs` / `ws.rs`.
- Rules: reusable unpinned links (#1988), guest cap (`guest-limit` vs `workspace-full`), projected `system.status` for collaborators (field list from `control.rs`).
- Collaborator allowlist brought in line with the frozen goldens (`COLLABORATOR_METHODS` 188; refused remainder 209): `agent.create` / `delegate` / `wakeOrCreate` moved to refused (#1998), `system.status` added to allowed, `agent.memoryUsage`, `principal.list`, `workspace.members.add`, `github.identityProof.*` and the four `invite.*` named in the refused paragraph.
[...]

PR #5528, opened 2026-09-21
docs(protocol): Member rows use the nullable-key convention, not omitted keys

panghy:
Follow-up to #5526, landing the one review finding that could not be pushed after the PR entered the merge queue (thread https://github.com/intent-hq/intent/pull/5526#discussion_r4059040947).

`workspace.members.list` rows are built in `crates/intent-services/src/capability.rs` (intentd 7304c635) with `json!` and `Option` values, so `login` / `displayName` / `avatarUrl` are **present and `null`** when the cached profile lacks them. The `principal.list` row in `docs/protocol/methods/multiplayer.md` compared its always-present nullable keys against "the omitted-key convention of `Member` / `WorkspaceInvite`"; `Member` now sits with `principal.me` on the nullable side, and only `WorkspaceInvite` omits absent keys.

One-line docs change; no gitlink move.

Verification: `make check-protocol-catalog` — consistent (Layer 1 + Layer 2; 10 warnings: docs lead the intentd pin).


panghy (commented):
Verdict: Approved (review scope), at 9f55b5a. The one-line correction now accurately groups Member with the always-present nullable profile-field convention, matching intentd 7304c635 capability.rs:394-405; the remaining finding from #5526 is resolved. No additional findings. Read-only catalog checks at the pinned revision and target tip, plus git diff --check, all exit 0 with only expected transitional warnings. CI docs-check has passed; the shell-test jobs were still running at verification time. No merge, auto-merge or enqueue action taken.

PR #5533, opened 2026-09-21
fix(bootstrap): read the launcher version from the semver-shaped output line

### LINKED ISSUE #5509: Frontend RSS tests parse Datadog startup output as a number
Observed during final verification of cloudlands-fe#2699 at 7c34b882f on the remote Linux host. Full Vitest fails the two run-svelte-check.test.ts cases named reports a child that touched 256 MiB and exited 0/1, even within one interval. At scripts/run-svelte-check.test.ts:381, Number(output.trim()) becomes NaN because inherited host Datadog instrumentation writes startup configuration to the child stdout. These test/runner files are unchanged by the feature branch.

Evidence: full run had 27949 passing / 3 failing tests; the third is a separate new GitLab icon fixture failure. Running DD_TRACE_STARTUP_LOGS=false corepack pnpm exec vitest run scripts/run-svelte-check.test.ts src/lib/components/shared/icons/FaWrapper.test.ts gives 47 passing / only the unrelated icon failure. No feature-code change is needed for the RSS failure.

Suggested fix: make the numeric measurement channel independent of injected child stdout, or sanitize child tracing setup in this harness. Add a regression where child stdout has an unrelated preamble and the RSS measurement is still reported correctly. Keep this follow-up separate from the forge feature. Dedup searched Datadog, run-svelte-check NaN, and RSS test; no matching issue found.

panghy:
## Summary

`launcher_probe` in `scripts/bootstrap-dev-host.sh` kept the **first** line of `corepack --version` / `pnpm --version`. On a host that injects Datadog tracing into Node, the tracer's startup configuration lands on stdout first, so `pnpm_ready` compared that log line against the pinned `packageManager` version and reported pnpm as not ready until `DD_TRACE_STARTUP_LOGS=false` was exported (second site named in the issue's comment).

The probe now keeps the **last semver-shaped line** (`^v?[0-9]+\.[0-9]+\.[0-9]+`) and falls back to the first line when there is none, so both the doctor's `Corepack:` line and the pnpm pin comparison ignore a leading banner. Error reporting (non-zero exit → first output line) is unchanged. Only `grep`/`sed` are used, matching the hermetic test's PATH allowlist.

Companion PR on cloudlands-fe fixes the RSS test oracle (first site).

Fixes intent-hq/intent#5509

## Tests added (`scripts/bootstrap-dev-host.test.sh`)

- Doctor: a `corepack` launcher that prints a Datadog banner then `0.35.0` → `[ok] Corepack: 0.35.0` (RED on `origin/main`: the doctor printed the banner as the version).
- `pnpm_ready` (install mode, via the extracted-functions fixture): banner + `10.30.3` → ready with `PROBE_OUTPUT=10.30.3`; banner + `10.29.0` → not ready; banner only → not ready.

## Shim evidence

Fake `pnpm` on PATH printing `DATADOG TRACER CONFIGURATION - {"lang":"nodejs"}` then `10.30.3`, `packageManager: pnpm@10.30.3`:

```text
--- origin/main:
pnpm_ready=1 PROBE_OUTPUT=[DATADOG TRACER CONFIGURATION - {"lang":"nodejs"}]
--- fixed:
pnpm_ready=0 PROBE_OUTPUT=[10.30.3]
```

## Verification

- `bash -n` and `bash --posix -n scripts/bootstrap-dev-host.sh`: OK.
- `bash scripts/bootstrap-dev-host.test.sh`: passed (RED on `origin/main` script + new tests: `expected doctor output to contain: [ok] Corepack: 0.35.0`).
- `bash scripts/dev-sandbox.test.sh` (guards the frontend-toolchain wording): passed.


panghy (commented):
Reviewed 55f53f64cf8a0589962b175559821248237f5bf5: no blocking findings or nits.

The successful-probe change preserves nonzero-exit and timeout handling, uses only the existing PATH allowlist, and handles grep no-match safely under the script’s set -u (without set -e). The new shim assertions use fixed fixture versions rather than host state.

Verified: bash -n and bash --posix -n; bootstrap-dev-host.test.sh; dev-sandbox.test.sh; and an independent hermetic 18-case matrix in normal Bash and --posix mode covering last-match selection, 10.30.3, v24.19.0, wrong pins, no-match/empty output, and a failed launcher that still prints a valid version. Parent e1f5ead fails the banner negative control in both modes. All head checks passed.

Limits: Bash 3.2 is unavailable here; the added shell syntax and POSIX grep/sed expressions were checked statically for compatibility. Existing loadable node-pty assertions were skipped because this checkout has no built binary; the new regression cases ran.

PR #5619, opened 2026-09-22
docs(protocol): document the image dimension sidecar (v10.7)

panghy:
## Summary

Protocol docs for the **image dimension sidecar** (stable inline chat image sizing), landing ahead of the intentd / cloudlands-fe PRs per "docs lead the pin". Docs only — no `packages/*` gitlink moves.

Additive minor bump **10.6 → 10.7**:

- `docs/protocol/methods/agents.md` — `agent.getConversation`: text block `media?: { [src]: { width, height } }` (key = the Markdown `src` exactly as written; probed sources `workspace-asset://`, `intent://local/[ws/]file/…`, bare workspace-relative paths; `http(s)`/`data:` never probed; omitted when empty), image block `width?`/`height?` (intrinsic dims of the original), slim-projection behavior (both kept, alongside `dataIsThumbnail`), write-path-only cost contract. `agent.getMessageBlock`: fields served as persisted.
- `docs/protocol/07-agent-streaming.md` — §7.1 delta envelope: live text chunk deltas carry only the `media` entries resolved by that chunk; the client unions them; persisted = union, so the snapshot+delta invariant holds.
- `docs/protocol/versioning.md` + `docs/protocol/README.md` — 10.7 entry and version bump.

## Verification

- `make docs-check` — pass
- `make consumer-checks` — pass (the 9 catalog warnings and 4 event-catalog warnings are pre-existing "docs lead the pin" notices for `sourceControl.*` / `invite.redeem` / `principal:identity-changed`, unrelated to this change)

Component PRs follow on branch `images-chat` in intentd and cloudlands-fe.


docs/protocol/methods/agents.md
```diff
@@ -9,8 +9,8 @@ The largest namespace. Every `agent.*` method is served daemon-primary by `inten
...
...get, keeping the page's anchor end (newest for legacy backward pages, oldest for `prevToken` forward continuations, the target message — grown outward within budget — for seek pages) and always serving at least one message (a single over-budget message serves alone; never an empty page or a token loop). The continuation cursor(s) — `nextToken`, and `prevToken` on the seek path — are re-minted at the first excluded row, so an existing token walk resumes seamlessly and reconstructs the identical sequence as unbudgeted paging, just in more round-trips. Frame ceiling = budget + the slim size of the last admitted message — NOT a hard bound: the ≥1-message floor means a single message with enough capped blocks can alone exceed the budget (and even the transport's 1 MiB large-frame warn), so the budget bounds the typical page at ~512 KiB but the worst case degrades to one whole message per page rather than a guaranteed frame size. `totalMessages` / `truncated` semantics are unchanged (transcript-wide, not page-length), and the seq-0 `chat.subscribe` snapshot inherits the bound automatically (it reuses this read) — since v8.0 every wire read is a slim read, so every wire read is budgeted |
```
panghy: [P2] Distinguish live text probing from image-block persistence timing

The new write-path paragraph says "probing happens when the block is persisted", but the same row and section 7.1 require each live text delta to carry the dimensions resolved by that delta. The Spec Wire contract likewise requires those dimensions during streaming, before final message persistence; persistence-time-only probing would miss the reserved-box behavior while the image first loads. Please describe text-reference probing as occurring on the write/stream path before the corresponding delta is emitted, with the accumulated sidecar persisted, while retaining the separate persistence-time statement for image-block dimensions. The no-read-time-probing/no-backfill rule should stay unchanged.

Related wording nit in section 7.1: the preceding delta-envelope paragraph still calls incremental textDelta "the one exception" to full-current-block replacement; media now adds another exception. Referring to the media merge rule there would keep the surrounding reducer contract consistent.

panghy (commented):
Reviewed against the workspace Spec Wire contract and the documentation task definition of done. make docs-check and make consumer-checks both pass (exit 0; expected docs-lead-the-pin warnings only). The PR contains only four protocol Markdown files, with no packages/* paths or gitlink moves. Version 10.7, optional field shapes, exact-src keys, allowed sources, omission rules, original image dimensions and slim projection all match the spec. No release-triggering footer tokens were found in the changed documents, PR title/body or commit messages. One write-path timing inconsistency needs clarification below. This is a comment-only review, not a GitHub approval.

docs/protocol/methods/agents.md
panghy: Re-verified at 7a06200f8fce323d16fcda2797912cccfd2d0ea7. The method row now distinguishes pre-delta stream-path text probing from image-block persistence-time stamping, and section 7.1 names both delta exceptions. Both make docs-check and make consumer-checks pass. This finding is resolved.

panghy (commented):
Final documentation re-review at a180a5816286e0ac28661997a18c4b237acdb5fc: no remaining findings in the requested scope. The stream-path/persistence distinction and both delta exceptions are now consistent with the Spec Wire contract. The current client.hello/system.status examples advertise 10.7; historical release prose is unchanged. Field shapes, exact-source keys, allowed sources, omission and union rules, original image dimensions and slim projection match the contract. All three review threads are resolved. Independently reran make docs-check and make consumer-checks: both exit 0 with accepted docs-lead-the-pin warnings. Whitespace, release-footer-token and scope checks pass; the PR changes only four protocol Markdown files and no gitlinks. This is a comment-only review, not a GitHub approval. Runtime daemon/frontend behavior remains separate component verification.

PR #5638, opened 2026-09-22
ci: regenerate the MCP bindings index in the submodule auto-bump

panghy:
## Summary

An intentd change to a `ws.*` binding's help line had no automated green landing path: `check-mcp-bindings` requires `docs/protocol/methods/mcp-bindings.md` to match the *pinned* intentd help text line for line, so the docs PR cannot regenerate the index ahead of the pin, and the auto-bump moved the gitlink only. The rolling `auto/submodule-bump` PR then failed `docs-check` (#5621) and a hand-filed `submodule-pin-intended` PR (#5620) was the only green path — which in turn stranded #5621 open and red.

This PR makes the auto-bump produce a tree that passes the existing check:

- **Regenerate the index in the bump commit** — `scripts/auto-bump-submodules.sh` fetches `tools.rs` at the new intentd tip (single-blob partial fetch, git auth only), runs `check-mcp-bindings.mjs --write` against a temporary root, and adds the regenerated index to the temporary index only when it differs from `HEAD`. The worktree is never touched, the "identical tree → skip push" idempotency still holds, and every regeneration failure warns and falls back to today's gitlink-only bump. The workflow gains a `setup-node` step.
- **Close the stranded rolling PR** — when no pin is behind and an `auto/submodule-bump` PR is open, close it with a comment and delete the branch instead of exiting silently. `--dry-run` stays `gh`-free.
- **Tests** — new `scripts/auto-bump-submodules.test.sh` (hermetic `file://` bare-remote fixtures, stateful `gh` stub, fault injection for the fail-soft paths), registered in the `shell-tests` job.
- **Prose** — `AGENTS.md` Phase 2 and the `consumer-checks.yml` header now name the auto-bump as the index regenerator (replaced, not added).

Evidence: intent-hq/intentd#2060 → #5607 (docs PR failed `docs-check` twice), #5621 (rolling PR red), #5620 (manual pin PR).

## Verification

- `bash scripts/auto-bump-submodules.test.sh` — pass (incl. changed/unchanged help text, rerun skip, stale-PR close, dry-run gh-free, hash-object / cacheinfo fault injection)
- `npx shellcheck scripts/auto-bump-submodules.sh scripts/auto-bump-submodules.test.sh` — clean
- `bash scripts/auto-bump-submodules.sh --dry-run` from the repo root with a logging `gh` stub — exit 0, no `gh` calls, worktree/index/refs unchanged
- `bash scripts/docs-check.test.sh`, `make lint-shell-sleeps`, `make docs-check`, `node scripts/check-doc-links.mjs` — pass


scripts/auto-bump-submodules.sh
```diff
@@ -72,10 +88,38 @@ while read -r key path; do
...
+  fi
+}
+
 if [ ${#paths[@]} -eq 0 ]; then
   echo "All submodule pins match their remote tips; nothing to do."
+  if [ "$DRY_RUN" = 1 ]; then
+    echo "dry-run: skipping the stale $BRANCH PR check (requires gh)."
+    exit 0
+  fi
+  close_stale_pr
```
panghy, line 132: [P2] Do not close the rolling PR when a submodule tip could not be checked

An empty paths array does not prove all pins are current: the existing iOS failure path at lines 68-71 warns and continues without recording the unknown result. If a prior run opened an iOS bump, and the next run cannot read the private iOS remote while intentd and cloudlands-fe are current, this new call closes the still-needed PR and deletes its branch, incorrectly claiming main already carries its pins.

I reproduced this using the new local bare-Git fixtures plus an iOS gitlink: main pointed to commit A, the pending rolling PR pointed iOS to commit B, and the iOS remote was unreadable. The script exited 0 and the gh stub recorded pr close 7 --delete-branch even though A != B.

Please track incomplete/skipped pin reads and skip stale-PR cleanup with a warning in that case, or compare the pending PR gitlinks with main before closing it. Add a regression with current public pins, an unreadable iOS remote, and a pending iOS-only bump; it must leave the PR/branch untouched while still exiting 0. Keep the existing proven-no-drift close test.

panghy (commented):
Critical review completed against 4de6f3f..90d2732. One P2 correctness issue needs fixing before landing; see the inline comment. No cosmetic nits. The auto-bump fixture suite, docs-check fixture suite, shellcheck on both shell files, lint-shell-sleeps, docs-check, and documentation-link check all passed locally. I also inspected checkout@v7 token persistence: the explicit --includes lookup matches its includeIf-backed credential config. The changed-help/unchanged-help and exact-tree rerun paths pass the fixtures. No code was edited and no formal approval or request-changes review is being submitted.

scripts/auto-bump-submodules.sh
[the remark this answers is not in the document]
panghy: Fixed in 9a5e8287a1939219ce06adc441b7928695b19d55: the detection loop now records every skipped pin read (`packages/ios` with an unreadable tip, or a path with no gitlink in HEAD) in a `skipped` array. When no pin is behind, `close_stale_pr` only runs if nothing was skipped; otherwise the script prints `Skipped <paths>: an open auto/submodule-bump PR may still carry its pin, so it cannot be proven stale; leaving it untouched.` and exits 0. The header comment was updated to match.

Regression case added to `scripts/auto-bump-submodules.test.sh` (intentd current, `packages/ios` with a failing `ls-remote`, gh stub reporting open PR 7): asserts no `gh pr close`, PR still open, exit 0. Confirmed it fails against the pre-fix script (it closed PR #7) and passes after the fix; shellcheck on both files and `make lint-shell-sleeps` are clean.

panghy: Fixed in 9a5e8287a1939219ce06adc441b7928695b19d55: the detection loop now records every skipped pin read (`packages/ios` with an unreadable tip, or a path with no gitlink in HEAD) in a `skipped` array. When no pin is behind, `close_stale_pr` only runs if nothing was skipped; otherwise the script prints `Skipped <paths>: an open auto/submodule-bump PR may still carry its pin, so it cannot be proven stale; leaving it untouched.` and exits 0. The header comment was updated to match.

Regression case added to `scripts/auto-bump-submodules.test.sh` (intentd current, `packages/ios` with a failing `ls-remote`, gh stub reporting open PR 7): asserts no `gh pr close`, PR still open, exit 0. Confirmed it fails against the pre-fix script (it closed PR #7) and passes after the fix; shellcheck on both files and `make lint-shell-sleeps` are clean.

PR #5655, opened 2026-09-22
fix(status): make the dev-status probe budget tunable

panghy:
Every non-port probe in scripts/dev-status.sh (doctor, dev-sandbox.sh
status, sandbox health, git, gh) ran under fixed literal budgets
(3 s / 1 s / 0.4 s). On a loaded host the doctor and sandbox-status
probes each cost ~0.9 s at load average 16 and exceeded the 3 s budget
at load ~50, emptying "sandboxes" and flaking the populated-sandbox
assertion in scripts/dev-status.test.sh.

Add DEV_STATUS_PROBE_TIMEOUT (default 10 s, fractional ok) as the single
knob for every non-port probe, sharing the port knob's parsing and
warning behaviour; no numeric timeout= literal remains in the script.
The test suite exports a generous default for both knobs, derives its
no-gh wall-clock bound from both, guards against a reintroduced literal,
and checks the degraded (tiny budget) and invalid-value paths.

Follow-up to intent-hq/intent#5411 / #5422, which covered the port
probe only.

Agent-Id: agent-50f35ead-aaeb-4453-b50c-8bf52ee1feb2


scripts/dev-status.test.sh
```diff
@@ -160,6 +167,31 @@ grep -q "$coverage_line" <(PATH="$bin_dir:$PATH" SANDBOX_STATE_DIR="$state_dir"
...
+  || fail "status exited non-zero under an invalid DEV_STATUS_PROBE_TIMEOUT"
+grep -q "ignoring DEV_STATUS_PROBE_TIMEOUT='abc'" "$temp_dir/invalid-probe.err" \
+  || fail "invalid DEV_STATUS_PROBE_TIMEOUT was not reported on stderr: $(cat "$temp_dir/invalid-probe.err")"
+[[ "$(grep -c 'ignoring DEV_STATUS_PROBE_TIMEOUT' "$temp_dir/invalid-probe.err")" == 1 ]] \
+  || fail "invalid DEV_STATUS_PROBE_TIMEOUT warning was not printed exactly once"
+python3 - "$temp_dir/invalid-probe.json" <<'PY' || fail "invalid probe budget did not fall back to the default"
+import json
+import sys
+report = json.load(open(sys.argv[1], encoding="utf-8"))
+assert len(report["sandboxes"]) == 1, report["sandboxes"]
```
panghy: [P2] Avoid requiring the invalid-value case to finish within the fixed default

Setting DEV_STATUS_PROBE_TIMEOUT=abc intentionally replaces the suite override with the production 10-second default. Requiring one sandbox here therefore restores a non-tunable 10-second ceiling on a real host probe: a loaded host whose sandbox status needs 11 seconds passes the normal 30-second case but fails this assertion, and raising the suite knob cannot help. I reproduced that outcome using the actual sandbox_status definition with a mocked 11-second subprocess (30 -> one sandbox; abc -> empty list). This conflicts with the goal of making the suite tunable under host load. Please test the exact fallback value with a controlled/mocked probe and leave the integration case checking the warning, exit status, and valid report without requiring the real probe to beat the fixed default.

scripts/dev-status.test.sh
```diff
@@ -54,12 +54,19 @@ for command in bash cksum date dirname git grep head python3 sed awk; do
...
-# i.e. the historical 5000 ms at the former 2 s default) so it still catches a
-# missing `gh` hanging the report.
+# `ports`, and the probe budget (doctor, sandbox, git, gh) likewise never
+# empties `sandboxes`; the no-gh wall-clock bound scales with both (3 s of
+# slack on top, i.e. the historical 5000 ms at the former 2 s port default) so
+# it still catches a missing `gh` hanging the report.
 export DEV_STATUS_PORT_TIMEOUT="${DEV_STATUS_PORT_TIMEOUT:-30}"
-no_gh_budget_ms=$(python3 -c 'import os; print(int(float(os.environ["DEV_STATUS_PORT_TIMEOUT"]) * 1000) + 3000)')
+export DEV_STATUS_PROBE_TIMEOUT="${DEV_STATUS_PROBE_TIMEOUT:-30}"
+no_gh_budget_ms=$(python3 -c 'import os; print(int((float(os.environ["DEV_STATUS_PORT_TIMEOUT"]) + float(os.environ["DEV_STATUS_PROBE_TIMEOUT"])) * 1000) + 3000)')
```
panghy: [P2] Account for sequential non-port probes in the total bound

DEV_STATUS_PROBE_TIMEOUT is a per-call limit, not a shared limit for all non-port work. Even with gh absent and no sandboxes, this invocation runs doctor_status and sandbox_status sequentially, followed by multiple git probes. The new formula budgets for only one of those calls. With both knobs at 30, doctor, ports, and sandbox status taking 25 seconds each are all within their allowed limits but require at least 75 seconds, so the 63-second assertion fails on a valid report. I verified that trace against the actual functions with mocked durations. Please derive the bound from the sequential probe count, or isolate the no-gh timing check with controlled fast probes. The task currently prescribes this formula, so its requirement should be corrected alongside the test.

scripts/dev-status.test.sh
```diff
@@ -54,12 +54,19 @@ for command in bash cksum date dirname git grep head python3 sed awk; do
   ln -s "$(command -v "$command")" "$bin_dir/$command"
 done
 
+# Every probe budget in dev-status.sh must come from a knob so the suite can
+# raise it under load; a numeric literal reintroduces a fixed wall-clock budget.
+! grep -nE 'timeout=[0-9]' "$script" \
```
panghy: [P2] Recognize numeric Python timeouts rather than one text spelling

This guard allows timeout = 3, timeout= 3, timeout=.4, and timeout=(3), all valid ways to reintroduce the fixed budgets it is intended to reject. It also rejects a harmless comment or string containing timeout=3. I ran those strings through the exact grep: the four numeric variants passed and both comment/string examples failed. Please inspect the embedded Python syntax (including function defaults and keyword arguments), or otherwise broaden the check and cover whitespace, leading-dot fractions, parenthesized literals, and comment/string-only cases with mutation tests. Otherwise ordinary formatting can silently defeat this acceptance criterion.

scripts/dev-status.sh
```diff
@@ -36,7 +39,52 @@ import urllib.request
...
-def run(command, *, cwd=root, env=None, timeout=3):
+def timeout_from_env(name, default):
+    raw = os.environ.get(name)
+    if raw is None or not raw.strip():
+        return default
+    try:
+        value = float(raw)
+    except ValueError:
+        value = None
+    if value is None or not math.isfinite(value) or value <= 0:
```
panghy: [P3] Reject finite values that the timeout APIs cannot represent

The finite/positive check accepts 1e20, but passing it to subprocess.run raises OverflowError, which run does not catch. Reproduced with DEV_STATUS_PROBE_TIMEOUT=1e20 bash scripts/dev-status.sh --json: exit 1, no JSON, and OverflowError: timestamp too large to convert to C _PyTime_t. This is an extreme configuration edge, but it bypasses the intended warning/default or graceful-degradation behavior. Consider rejecting unrepresentable values for the new knob (while preserving the stated port-knob compatibility constraint), and add a regression test. Health socket timeouts also need to be considered if validation happens only in run.

panghy (commented):
Reviewed 8cfda9e against 0376ba2. The focused dev-status suite passes (no-gh 4661ms), as do bash syntax, docs-check, fixed-sleep lint, and diff checks. Shared parsing/default routing looks correct for ordinary inputs. The inline comments cover three reliability gaps in the new tests and one lower-priority input-validation edge. The no-gh formula is prescribed by the task itself, so that finding also needs a spec correction. No source changes made.

scripts/dev-status.test.sh
panghy: Addressed in 4117445. The bound is now `(DEV_STATUS_PORT_TIMEOUT + 2 × DEV_STATUS_PROBE_TIMEOUT) × 1000 + 3000` ms, counting the doctor probe, the port probe and the sandbox probe that the no-gh run executes serially (with `gh` absent from `PATH`, `github_ready()` returns before any `gh` call). The git calls also share the probe budget but run against the real checkout and are fast, so they are covered by the 3 s slack rather than the formula; the comment above the export explains the derivation. The task/spec formula was corrected alongside (the coordinator carried the spec correction into the follow-up task).

scripts/dev-status.sh
```diff
@@ -146,15 +168,15 @@ def sandbox_health(url):
         return None
     health_url = urllib.parse.urljoin(url.rstrip("/") + "/", "__sandbox/health")
     try:
-        with urllib.request.urlopen(health_url, timeout=0.4) as response:
+        with urllib.request.urlopen(health_url, timeout=probe_timeout()) as response:
```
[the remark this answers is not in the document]
panghy: Resolving without a code change; this semantics predates the diff. The former `timeout=0.4` was equally a per-socket-operation timeout on the same `urlopen(...)` + `json.load(response)` pair, so a server that trickles bytes below the interval could already extend the probe indefinitely; this PR only replaces the literal with the knob. In practice the health endpoint is the sandbox's own `/__sandbox/health`, a tiny JSON object served by a local Vite dev server on loopback, not an untrusted peer, so a slow-streaming body is not a scenario we budget for here. Bounding the total (a deadline around connect + read via a background thread or `socket` deadline arithmetic) would add complexity to a read-only status helper for no observed failure mode. If a hung sandbox ever stalls `make status`, the right fix is a per-request total deadline as a separate change; a note to that effect is in the PR Context log for the coordinator.

scripts/dev-status.test.sh
panghy: Addressed in 4117445. The rejected-value check (now a loop over `abc` and `1e20`) asserts only exit 0, exactly one stderr warning that names the fallback (`... ; using 10`), and a complete JSON report shape (`{"host","ports","sandboxes","repos","docs"}` with `sandboxes` a list). It no longer requires the live fixture sandbox to be reported, so nothing in the suite depends on a real probe finishing within the fixed production default; the comment above the loop states that explicitly. The exact fallback value is asserted through the warning text rather than through a live probe.

scripts/dev-status.test.sh
panghy: Addressed in 4117445. The grep is replaced by an `ast` walk: the suite extracts the heredoc between `<<'PY'` and the closing `PY`, parses it, and fails on any `timeout` keyword argument, parameter default (positional or keyword-only, including lambdas) or `timeout` assignment whose value is a numeric `Constant` (int/float/complex, `bool` excluded, unary `-`/`+` unwrapped). Comments and strings are not AST candidates, so they never count; a parse failure also fails the guard.

Mutation checks against scratch copies of `dev-status.sh`, running the suite's guard block verbatim: `timeout=3`, `timeout = 3`, `timeout=.4`, `timeout=(3)`, `timeout=-1` on the `run()` default; `timeout=3` on the `urlopen` keyword; `timeout= 3` on the `port_status` keyword; a `timeout = 3` assignment inside `run()` — all fail and name the heredoc line and literal. The unchanged script, a comment containing `timeout=3` and a string `"timeout=3"` all pass.

scripts/dev-status.sh
panghy: Addressed in 4117445. The shared `timeout_from_env` now rejects values above `TIMEOUT_MAX = 86400.0` on the same warning-and-default path as non-numeric, non-finite and non-positive values, so both knobs are covered (including the `urlopen` health socket timeout, which reads `probe_timeout()`). The warning names the bound: `dev-status: ignoring DEV_STATUS_PROBE_TIMEOUT='1e20' (expected a positive number of seconds, at most 86400); using 10`. Values in range behave exactly as before, so the port knob's compatibility constraint is preserved (its only behavioral change is that a value that previously crashed now warns and falls back).

Verified: `DEV_STATUS_PROBE_TIMEOUT=1e20 STATUS_JSON=1 make status` and `DEV_STATUS_PORT_TIMEOUT=1e20 STATUS_JSON=1 make status` both exit 0 with the warning and a full JSON report. The suite's rejected-value loop now covers `1e20` alongside `abc` as the regression test; the header comment and `docs/fe/DEVELOPER_GUIDE.md` document the accepted range.

PR #5681, opened 2026-09-22
fix(scripts): treat a masked REST rate limit in shipped-in.sh as transient (exit 4)

panghy:
## Summary

`scripts/shipped-in.sh` classified a failed `gh` call as transient (exit 4, retry later) only when gh's stderr matched `transient_pattern`. Under REST throttling, `gh release download <tag> --pattern release-manifest.json` does not say "rate limit" — gh masks the 403 as `release not found` — so the script exited 1, and the canonical "Wait for shipped alpha" hook from AGENTS.md threw and was evicted exactly when GitHub was throttling (2026-09-22 09:47Z, workspace frosty-beaver).

## Change

- `gh_fail`: a failure whose stderr does not look transient now gets one REST probe on the releases repo (`gh api repos/intent-hq/cloudlands-releases` — same core quota; `gh api rate_limit` kept reporting 5000 remaining during the incident). A rate-limited probe reclassifies the failure as exit 4 and appends the rate-limit text to the original message; a healthy or non-transiently failing probe keeps the original message and exit 1, so a genuinely missing release/manifest is not masked. The fast path (stderr already transient) is unchanged and makes no probe call.
- Header comment updated for exit 4.
- `scripts/shipped-in.test.sh`: `notfound` stub mode, a probe stub with a rate-limit knob, and four cases — masked rate limit → 4; probe healthy → 1; no probe on the fast path; generic probe failure → 1.

The existing exit-4 contract in the Makefile help line, AGENTS.md and docs/RELEASING.md ("transient GitHub failure (rate limit, 5xx, network)") is unchanged.

## Verification

- `bash scripts/shipped-in.test.sh` — passed (bash 5.2.21; the Bash 3 pass ran the static gate only, no Bash 3 on the host — CI's `REQUIRE_BASH3=1` job covers it).
- `make check-makefile-targets` — passed.
- Independent verifier pass on the diff and tests — approved.


panghy (commented):
## Review summary

Reviewed c63eb5e4c8ab4a4932e4b8c941fa49450af0b01d. No blocking findings against the spec's acceptance criteria. This is a comment-only review, not a formal approval.

### Non-blocking: label non-rate-limit probe failures accurately (P3)

At `scripts/shipped-in.sh:121-125`, the probe uses the full `transient_pattern`, which also matches HTTP 5xx and network errors, but its diagnostic unconditionally says `GitHub REST API is rate limited`. I reproduced this without network access by substituting a shell-function `gh`: the release list returns v2.3.0, the manifest download fails with `release not found`, and the repository probe fails with either `gh: Server Error (HTTP 502)` or `error connecting to api.github.com`. Both runs exit 4, but both incorrectly claim a rate limit. The nearby comment at lines 111-112 also says anything other than a rate-limited probe keeps exit 1, which does not describe these branches.

Please consider keeping the useful exit-4 behavior while making the annotation generic (for example, `GitHub REST probe failed transiently: ...`), or selecting rate-limit wording only for rate-limit evidence. Update that adjacent comment and add probe-specific 502/network fixtures. The current fixtures exercise 5xx/network failures on the original call, not on the follow-up probe, so they cannot catch this diagnostic mismatch. This does not block the masked-rate-limit fix: the original error and the actual probe detail are preserved, and retry behavior remains safe.

## Verification

- `bash scripts/shipped-in.test.sh`: PASS, exit 0, Bash 5.2.21; local real Bash 3 run skipped because no interpreter is installed.
- `make check-makefile-targets`: PASS, exit 0; checked 4 cargo references against intentd@f05b6a0.
- `bash -n scripts/shipped-in.sh`, `bash -n scripts/shipped-in.test.sh`, and `git diff --check`: PASS.
- Independent probe matrix: healthy and non-transient 403 probes preserve the original message and exit 1; rate-limited, 502, and network probes exit 4 with empty stdout. The latter two demonstrate the diagnostic issue above.
- [Required macOS Bash 3 CI job](https://github.com/intent-hq/intent/actions/runs/35723051539/job/106730135160): SUCCESS; the workflow runs `REQUIRE_BASH3=1 bash scripts/shipped-in.test.sh`.

The regression suite covers the masked download failure, healthy-probe hard failure, exactly one probe on the slow path, and zero probes for already-transient failures. The new `if ! gh ... && grep ...` condition is safe under `set -euo pipefail`; variable uses and redirections are quoted, and separate stderr files retain the original failure. The added API cost is one repository GET only on the otherwise-hard-failure path, with no probe on successful runs or transient fast-path failures. The documented exit-4 contract in AGENTS.md, Makefile, and docs/RELEASING.md remains accurate. No tracked files were edited.

panghy:
Addressed the P3 finding in a4a056d (`fix(scripts): name the actual REST probe failure in shipped-in.sh`):

- `gh_fail` now selects the rate-limit wording only on rate-limit evidence in the probe's own stderr (`rate limit|HTTP 429`); any other transient probe failure (5xx, network) still exits 4 but is annotated `(GitHub REST API is unavailable: <probe stderr>)`. The original gh text stays first in the message.
- The comment above `gh_fail` now says a probe failing transiently for any reason yields exit 4, and only a healthy or non-transiently failing probe keeps exit 1.
- `shipped-in.test.sh`: the `probe.fail` knob now takes a failure mode as its content (empty = ratelimit), and two probe-specific fixtures were added — `release not found` + 502 probe and + network probe — asserting exit 4, no "rate limit" in stderr, the `is unavailable` annotation, and exactly one probe call.

Gates: `bash scripts/shipped-in.test.sh` PASS (bash 5.2.21; no Bash 3 on host, CI covers `REQUIRE_BASH3=1`), `make check-makefile-targets` PASS, `git diff --check` clean.

